PR #7, opened 2018-10-30
impl Send+Sync

### LINKED ISSUE #2: Unable to use hashbrown sctuctures between threads
It seems like hashbrown structures doesn't implement Send making it very hard to use them in multithreaded applications

For example, attempt to return them from thread via join handle:
```Rust
extern crate hashbrown;
fn main () {
    std::thread::spawn(move || { hashbrown::HashSet::new() } );
    std::thread::spawn(move || { hashbrown::HashMap::new() } );
}
```
Wouldn't compile
```
➜  cargo build
error[E0277]: `std::ptr::NonNull<u8>` cannot be sent between threads safely
 --> src/main.rs:4:5
  |
4 |     std::thread::spawn(move || { hashbrown::HashSet::new() } );
  |     ^^^^^^^^^^^^^^^^^^ `std::ptr::NonNull<u8>` cannot be sent between threads safely
  |
  = help: within `hashbrown::HashSet<_>`, the trait `std::marker::Send` is not implemented for `std::ptr::NonNull<u8>`
  = note: required because it appears within the type `hashbrown::raw::RawTable<(_, ())>`
  = note: required because it appears within the type `hashbrown::HashMap<_, ()>`
  = note: required because it appears within the type `hashbrown::HashSet<_>`
  = note: required by `std::thread::spawn`

error[E0277]: `std::ptr::NonNull<(_, ())>` cannot be sent between threads safely
 --> src/main.rs:4:5
  |
4 |     std::thread::spawn(move || { hashbrown::HashSet::new() } );
  |     ^^^^^^^^^^^^^^^^^^ `std::ptr::NonNull<(_, ())>` cannot be sent between threads safely
  |
  = help: within `hashbrown::HashSet<_>`, the trait `std::marker::Send` is not implemented for `std::ptr::NonNull<(_, ())>`
  = note: required because it appears within the type `hashbrown::raw::RawTable<(_, ())>`

.....
```

Xaeroxe:
I believe this PR is safe but I'd appreciate some more eyes on this crate to confirm that.

Fixes #2

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1143,6 +1143,9 @@ pub struct OccupiedEntry<'a, K: 'a, V: 'a, S: 'a> {
     table: &'a mut HashMap<K, V, S>,
 }
 
+unsafe impl<'a, K, V, S> Send for OccupiedEntry<'a, K, V, S> where K: Send, V: Send, S: Send {}
+unsafe impl<'a, K, V, S> Sync for OccupiedEntry<'a, K, V, S> where K: Sync, V: Sync, S: Sync {}
+
 impl<'a, K: 'a + Debug, V: 'a + Debug, S> Debug for OccupiedEntry<'a, K, V, S> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.debug_struct("OccupiedEntry")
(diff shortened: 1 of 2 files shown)


Amanieu:
Thanks!

PR #14, opened 2018-10-31
Add Serde support

hcpl:
The code is taken as is from upstream Serde impls for `std::collections::{HashMap, HashSet}` barring few syntactic sugar changes.

Closes https://github.com/Amanieu/hashbrown/issues/6.


Amanieu:
Thanks!

PR #15, opened 2018-10-31
Allow more branches to be built on CI

hcpl:
Adapted from https://github.com/rust-lang-nursery/rust-clippy/pull/3359.


Amanieu:
Thanks!

Issue #10, opened 2018-10-31
Possible unsafety in the pub(crate) API of RawTable


cptroot:
After reviewing the PR `unsafe impl`'ing the `Send` and `Sync` traits I was looking around trying to see if I could trigger unsafe behavior. I didn't find anything with `Send` and `Sync`, but I think I've found a use-after-free in the `pub(crate)` API of `RawTable`.

Can someone take a look at this and make sure I'm not crazy? Again, this is in the non-exposed API, so it's not urgent, but it's always good to make unsafety impossible even internally to the codebase.

```
fn main() {
    let table = RawTable::new();
    let hash_builder = DefaultHashBuilder::new();
    let k = 50;
    let v = 500;
    let hash = make_hash(&hash_builder, &k);
    table.insert(hash, (k, v), |x| make_hash(&hash_builder, &x.0));
    let raw_iter = table.iter();
    drop(table); /* drops the pointers to data? */
    for bad in raw_iter {
        /* badness ensues */
    }
}
```

Amanieu:
The `RawTable` API was never designed to be safe. As you said, it is a completely internal API, so there is no risk of any user of this crate triggering unsafety.

cptroot:
Is there any harm in making the `RawTable::iter` method unsafe?

Amanieu:
I will accept a PR for that if you want to write one.

cptroot:
Awesome! I've made a PR.

PR #19, opened 2018-10-31
Change some internal APIs to be unsafe

cptroot:
This is to reflect the fact that the `iter` method doesn't tie it's lifetime to the originating `RawTable`, allowing for use-after-free. I've also update the comments around these changes to explain why calling the `iter` method is safe etc.


cptroot:
Meant to add that it addresses my concerns from #10.

Amanieu:
Thanks!

PR #21, opened 2018-10-31
Copy+paste fix

shepmaster:


Amanieu:
In fact, could you add links to commit ranges for the other releases as well (except 0.1.0, which doesn't have one).

PR #25, opened 2018-11-03
Insertion and removal tests

sambley:
Noticing failures with the removal of items at random places. Added a test to reproduce the issue, since the issue is random it may require the "cargo +nightly bench" to be run a few times to be able to see the failure.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ categories = ["data-structures", "no-std"]
 scopeguard = { version = "0.3", default-features = false }
 byteorder = { version = "1.0", default-features = false }
 serde = { version = "1.0", default-features = false, optional = true }
+rand = "0.5.1"
 
 [dev-dependencies]
 rustc-hash = "1.0"
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -13,11 +13,14 @@
 extern crate hashbrown;
 extern crate rustc_hash;
 extern crate test;
+extern crate rand;
 
 use std::hash::Hash;
 use test::Bencher;
 
 use hashbrown::HashMap;
+use hashbrown::HashSet;
+use rand::Rng;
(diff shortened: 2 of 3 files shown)


Amanieu:
Can you change this into a `#[test]` and change the random-number generator to a deterministic one (such as `XorShiftRng`)? This should allow the test to reliably reproduce the issue.

Amanieu:
This is fixed in #26

Cargo.toml
```diff
@@ -13,6 +13,7 @@ categories = ["data-structures", "no-std"]
 scopeguard = { version = "0.3", default-features = false }
 byteorder = { version = "1.0", default-features = false }
 serde = { version = "1.0", default-features = false, optional = true }
+rand = "0.5.1"
```
Amanieu: This should be in dev-dependencies since it is only used for tests.

benches/bench.rs
```diff
@@ -58,6 +61,27 @@ fn grow_by_insertion(b: &mut Bencher) {
...
+            assert_eq!(m.insert(tx[i].clone()), true);
+        }
+        for i in 0..4096 {
+            println!("removing {} {:?}", i, tx[i]);
+            assert_eq!(m.remove(&tx[i]), true);
+        }
+    });
+}
+
+
```
Amanieu: Delete this since we have the version in tests/set.rs now.

Amanieu:
Oops, didn't mean to close this.

sambley:
Thanks, that worked.

hcpl:
@sambley I think you didn't mean to close this? https://github.com/Amanieu/hashbrown/pull/26 only fixes the test failures you mentioned here, the tests themselves haven't been merged yet.

benches/bench.rs
Amanieu: This is not resolved. You should revert your changes to the benchmarks file and only keep the file in `tests`.

PR #28, opened 2018-11-08
Fix subtraction overflow in generic::Group::match_byte

cuviper:
The former `cmp - repeat(0x1)` would overflow if there was a match in the most
significant byte.  A `wrapping_sub` better matches the C code that this was
derived from.

This generic code is only used on non-SSE2 platforms, so I've also added an
i586 entry to the CI config.


Amanieu:
bors r+

Thanks!

PR #30, opened 2018-11-19
Add Rayon support

hcpl:
The code is adapted from `rayon-hash` to match the internal structure of SwissTables.

Compared to `rayon-hash` this implementation adds an `impl Send` for `ParIter` and `Debug` impls for parallel iterators to mirror iterators from `std::collections::{hash_map, hash_set}` and `hashbrown::{hash_map, hash_set}`.

The tests pass right now, but there are a couple of questions/potential improvements I would like to see resolved before merging --- mostly about correctness of everything in `external_crate_impls::rayon::raw`.

I tried to reduce usage of unsafe code from the beginning, so the only unsafe operations left are checking & setting bucket emptiness, getting references & values from buckets, setting `RawTable.items` field and manual impls of `Send`:
- I ended up writing a custom iterator over non-empty buckets in a raw table to emulate the approach from `rayon-hash`. Should I have reused `RawIter` and `RawIntoIter` instead?
- Manual `Send` impls. I have zero experience doing this, so I had to make some guesses based on `Send` and `Sync` docs + `rayon-hash` code.
  From what I could gather, any (only parallel? or sequential too?) iterator should `impl Send` iff:
  + Values the iterator returns immutable/shared references of are `Sync`.
  + Values the iterator returns mutable/unique references of are `Send`.
  + Values the iterator returns directly (by-value) are `Send`.
  E.g. `ParIterMut<'a, K, V>` is an `Iterator<Item = (&'a K, &'a mut V)>`, so the impl is
  ```rust
  unsafe impl<'a, K: Sync, V: Send> Send for ParIterMut<'a, K, V> {}
  ```
- `IntoParIter` relies on a sequence of events to work and I'm not sure if doing so is memory-safe/non-UB:
  + Set `RawTable.items` to 0 without doing any cleanup.
  + Iterate over all non-empty buckets. While doing so, set their control bytes to `EMPTY` and also read the value to pass to the folder.
  + Rely on the default `Drop` impl for `IntoParIter` which will drop the inner table.

The point about manual `Send` impls is also applicable for `HashSet` parallel iterators.

cc @cuviper


Amanieu:
I haven't reviewed this in detail yet, but here are my first impressions:
- I think `SplitBuckets` needs to be rewritten to use group-based splitting: the offset is always a multiple of the group size, and you never split a group (including cases where the table is smaller than the group size). This allows iteration to work the same way as the normal `RawIter` after splitting is done.
- Don't bother trimming empty buckets, it's not worth the trouble when buckets are roughly evenly distributed. Instead the group-based iterator can automatically skip consecutive empty buckets.

I will get back with further feedback when I find some time.

Amanieu:
I just pushed a [branch](https://github.com/Amanieu/hashbrown/tree/iter_range) which has `RawIterRange`. This should allow you to iterate over a subset of the table.

Amanieu:
Your implementation of `IntoParIter` is wrong: `hashbrown` does not use the item count to determine whether the table is empty. Instead you should use `into_alloc` (added in the `iter_range` branch) to convert the table into a raw allocation after you have extracted a `RawIterRange` from it. In the `Drop`, make sure that you finish iterating and dropping all items, and then free the table allocation.

Amanieu:
> So I conflated the item count and the allocated capacity. Will need to re-read the code more thoroughly.

Once you switch to using `into_alloc`, you won't need to make any modifications to the table as you read elements. In particular you do not need to reset the item count or clear the control byte for moved/dropped elements.

> One question though: what are the rules for handling panics? I see that Drop for RawTable<T> and RawIntoIter<T> don't use scopeguard, but RawDrain<'a, T> does. I think that's because RawDrain<'a, T> has borrow semantics which requires it to clear without deallocation, whereas RawTable<T> and RawIntoIter<T> have owned semantics so they deallocate on drop. Still, do their Drop impls just leak memory on panic?

The rule of thumb here is that a panic must not leave the table in an "unsafe" state. In the case of `RawDrain`, this happens if a control byte indicates an item as filled when in reality we have already moved the value out of its slot. Note that this can't happen for `RawIntoIter` since the table no longer exists. Remember that leaking memory is not considered unsafe, however accessing a dropped value is.

Amanieu:
Are you still working on this PR?

hcpl:
Sorry, I didn't know how to proceed, but forgot to ask.

Do I need to rebase on `iter_range` branch now or wait until it gets merged into master and only then rebase?

Amanieu:
Just rebase on top of `iter_range`.

hcpl:
Rebased and applied your suggestions, @Amanieu. Anything else to add or fix?

src/raw/mod.rs
```diff
@@ -874,6 +920,55 @@ impl<T> Iterator for RawIter<T> {
...
+impl<T> FusedIterator for RawIterRange<T> {}
+
+/// Iterator which returns a raw pointer to every full bucket in the table.
+pub struct RawIter<T> {
+    pub iter: RawIterRange<T>,
+    items: usize,
+}
+
+unsafe impl<T> Send for RawIter<T> where T: Send {}
+unsafe impl<T> Sync for RawIter<T> where T: Sync {}
```
Amanieu: I think these can actually be removed since they are implied by the impl on `RawIterRange`.

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,539 @@
...
+    type Item = (&'a K, &'a V);
+    type Iter = ParIter<'a, K, V>;
+
+    #[inline]
+    fn into_par_iter(self) -> Self::Iter {
+        ParIter { table: self }
+    }
+}
+
+unsafe impl<'a, K: Sync, V: Sync> Send for ParIter<'a, K, V> {}
```
Amanieu: This is already implied `&'a RawTable<(K, V)>`.

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+impl<'a, K, V> From<RawIterRange<(K, V)>> for ParIterProducer<'a, K, V> {
+    #[inline]
+    fn from(iter: RawIterRange<(K, V)>) -> Self {
+        Self { iter, marker: PhantomData }
+    }
+}
+
+unsafe impl<'a, K: Sync, V: Sync> Send for ParIterProducer<'a, K, V> {}
```
Amanieu: This will be implied by the `PhantomData` once you add it.

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+impl<'a, K, V> From<RawIterRange<(K, V)>> for ParKeysProducer<'a, K, V> {
+    #[inline]
+    fn from(iter: RawIterRange<(K, V)>) -> Self {
+        Self { iter, marker: PhantomData }
+    }
+}
+
+unsafe impl<'a, K: Sync, V> Send for ParKeysProducer<'a, K, V> {}
```
Amanieu: This will be implied by the `PhantomData` once you add it.

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+impl<'a, K, V> From<RawIterRange<(K, V)>> for ParValuesProducer<'a, K, V> {
+    #[inline]
+    fn from(iter: RawIterRange<(K, V)>) -> Self {
+        Self { iter, marker: PhantomData }
+    }
+}
+
+unsafe impl<'a, K, V: Sync> Send for ParValuesProducer<'a, K, V> {}
```
Amanieu: This will be implied by the `PhantomData` once you add it.

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+impl<K: Sync, V> RawTable<(K, V)> {
+    #[inline]
+    pub fn par_keys(&self) -> ParKeys<K, V> {
+        ParKeys { table: self }
+    }
+}
+
+unsafe impl<'a, K: Sync, V> Send for ParKeys<'a, K, V> {}
```
Amanieu: This is already implied `&'a RawTable<(K, V)>`.

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+impl<K, V: Sync> RawTable<(K, V)> {
+    #[inline]
+    pub fn par_values(&self) -> ParValues<K, V> {
+        ParValues { table: self }
+    }
+}
+
+unsafe impl<'a, K, V: Sync> Send for ParValues<'a, K, V> {}
```
Amanieu: This is already implied `&'a RawTable<(K, V)>`.

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,539 @@
...
+    type Item = (&'a K, &'a mut V);
+    type Iter = ParIterMut<'a, K, V>;
+
+    #[inline]
+    fn into_par_iter(self) -> Self::Iter {
+        ParIterMut { table: self }
+    }
+}
+
+unsafe impl<'a, K: Sync, V: Send> Send for ParIterMut<'a, K, V> {}
```
Amanieu: This is not exactly implied by `&'a mut RawTable<(K, V)>`, but it is good enough. Remove this impl.

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+impl<K, V: Send> RawTable<(K, V)> {
+    #[inline]
+    pub fn par_values_mut(&mut self) -> ParValuesMut<K, V> {
+        ParValuesMut { table: self }
+    }
+}
+
+unsafe impl<'a, K, V: Send> Send for ParValuesMut<'a, K, V> {}
```
Amanieu: This is not exactly implied by `&'a mut RawTable<(K, V)>`, but it is good enough. Remove this impl.

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+impl<'a, K, V> From<RawIterRange<(K, V)>> for ParIterMutProducer<'a, K, V> {
+    #[inline]
+    fn from(iter: RawIterRange<(K, V)>) -> Self {
+        Self { iter, marker: PhantomData }
+    }
+}
+
+unsafe impl<'a, K: Sync, V: Send> Send for ParIterMutProducer<'a, K, V> {}
```
Amanieu: This will be implied by the `PhantomData` once you add it.

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+impl<'a, K, V> From<RawIterRange<(K, V)>> for ParValuesMutProducer<'a, K, V> {
+    #[inline]
+    fn from(iter: RawIterRange<(K, V)>) -> Self {
+        Self { iter, marker: PhantomData }
+    }
+}
+
+unsafe impl<'a, K, V: Send> Send for ParValuesMutProducer<'a, K, V> {}
```
Amanieu: This will be implied by the `PhantomData` once you add it.

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,539 @@
...
+        f.debug_list().entries(entries).finish()
+    }
+}
+
+struct IntoParIterProducer<K, V> {
+    iter: RawIterRange<(K, V)>,
+    alloc: Option<(NonNull<u8>, Layout)>,
+}
+
+unsafe impl<K: Send, V: Send> Send for IntoParIterProducer<K, V> {}
```
Amanieu: You can add a `Sync` impl here as well.

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,539 @@
...
+            if mem::needs_drop::<(K, V)>() {
+                while let Some(item) = self.iter.next() {
+                    item.drop();
+                }
+            }
+
+            // Free the table
+            if let Some((ptr, layout)) = self.alloc {
+                dealloc(ptr.as_ptr(), layout);
+            }
```
cuviper: I don't think this is safe to `dealloc` from one of potentially many producers sharing this data.  You're keeping `alloc` in the left-most split, but nothing guarantees that will outlive the other splits.

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,539 @@
...
+    type Item = (K, V);
+
+    fn drive_unindexed<C>(self, consumer: C) -> C::Result
+    where
+        C: UnindexedConsumer<Self::Item>,
+    {
+        let iter = unsafe { self.table.iter().iter };
+        let alloc = self.table.into_alloc();
+        let producer = IntoParIterProducer { iter, alloc };
+        plumbing::bridge_unindexed(producer, consumer)
```
cuviper: I would arrange for `dealloc` after the bridge returns here, rather than doing it within a producer, as here we know the table is definitely done across all producer splits.

src/raw/mod.rs
```diff
@@ -874,6 +920,55 @@ impl<T> Iterator for RawIter<T> {
...
     }
 
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        // We don't have an item count, so just guess based on the range size.
+        (0, Some(unsafe { offset_from(self.end, self.ctrl) }))
+    }
+}
+
+impl<T> ExactSizeIterator for RawIterRange<T> {}
```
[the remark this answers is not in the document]
Amanieu: Oops, my bad. Yea this impl should be removed.

src/external_trait_impls/rayon/raw.rs
Amanieu: Also make sure you use a scope guard to avoid leaking if a panic happens.

Amanieu:
Status update?

Amanieu:
Since @hcpl is inactive, I am taking over this PR in #37.

PR #37, opened 2018-12-09
Add Rayon support

### LINKED ISSUE #16: Rayon support
It'd be great to have Rayon compatibility for hashbrown's `HashMap` and `HashSet`. I think it's a case of implementing `IntoParallelIterator`, `FromParallelIterator` and `ParallelExtend`.

Amanieu:
Taking over from #30 since the author seems to be inactive.

Fixes #16 

cc @hcpl @cuviper

### DIFF
--- a/src/external_trait_impls/rayon/raw.rs
+++ b/src/external_trait_impls/rayon/raw.rs
@@ -12,7 +12,6 @@ use rayon::iter::{
 
 use raw::{RawIterRange, RawTable};
 
-
 #[inline]
 unsafe fn raw_iter_range_split_and_convert<T, P>(iter_range: RawIterRange<T>) -> (P, Option<P>)
 where
@@ -21,7 +20,6 @@ where
     let (left, right) = iter_range.split();
     (P::from(left), right.map(P::from))
 }
-
 
 /// Parallel iterator over shared references to entries in a map.
 ///
@@ -52,7 +50,8 @@ impl<'a, K: Sync, V: Sync> ParallelIterator for ParIter<'a, K, V> {
     type Item = (&'a K, &'a V);
 
     fn drive_unindexed<C>(self, consumer: C) -> C::Result
-        where C: UnindexedConsumer<Self::Item>
+    where
+        C: UnindexedConsumer<Self::Item>,
     {
         let buckets = unsafe { self.table.iter().iter };
         let producer = ParIterProducer::from(buckets);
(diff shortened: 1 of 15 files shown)


hcpl:
I apologise for not responding. After reading review comments in my PR, I thought that making proposed changes will require several hours of debugging segfaults (like the pointer range calculation bug), which I didn't feel like doing at the time. Re-reading comments and skimming your https://github.com/Amanieu/hashbrown/pull/37/commits/0f8d03539184f56e58ca8b711af9a5d0079c7128 made me realise that it's easy enough to deal with in an hour or so, so I overestimated the time needed.

That said, thanks for taking over and making progress!

Amanieu:
It's fine, I just wanted to get this in so that the `RawIter` changes can be ported to the libstd PR as well. I want to keep libstd and hashbrown in sync as much as possible.

Sorry about the `split` bug, this was my fault for being careless.

cuviper:
> I want to keep libstd and hashbrown in sync as much as possible.

Are you going to keep maintaining hashbrown (w/ rayon) once it's in std?  This would make `rayon-hash` obsolete, which is fine by me.  Or if std exposes enough of the `RawIter` interface to stable users, then `rayon` itself could steal the parallel implementation from here.

Amanieu:
I will keep maintaining it for a while, but I eventually plan on making some breaking changes to the `HashMap` API, at which point I will change the version to 0.2.

cuviper:
OK, then maybe I'll keep going with rayon-hash too, keeping std-compatible.  It's only fair to steal the code back when AFAICT the impl here started as a copy of rayon-hash's. :smile:

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,184 @@
...
+    {
+        folder.consume_iter(self.iter)
+    }
+}
+
+/// Parallel iterator which consumes a table and returns elements.
+pub struct RawIntoParIter<T> {
+    iter: RawIterRange<T>,
+    alloc: Option<(NonNull<u8>, Layout)>,
+}
```
cuviper: If this struct gets dropped without ever driving it, then I think you'll leak both the items and the raw allocation.

src/external_trait_impls/rayon/raw.rs
```diff
@@ -0,0 +1,184 @@
...
+        });
+        let producer = ParDrainProducer { iter: self.iter };
+        plumbing::bridge_unindexed(producer, consumer)
+    }
+}
+
+/// Parallel iterator which consumes elements without freeing the table storage.
+pub struct RawParDrain<'a, T> {
+    table: &'a mut RawTable<T>,
+}
```
cuviper: If this struct gets dropped without ever driving it, then it won't have drained any items.

Amanieu:
The leak is fixed by #38.

PR #46, opened 2019-02-25
Prepare hashbrown for inclusion in the standard library

Amanieu:
See https://github.com/rust-lang/rust/pull/58623.


src/rustc_entry.rs
```diff
@@ -0,0 +1,575 @@
...
+        if let Some(elem) = self.table.find(hash, |q| q.0.eq(&key)) {
+            RustcEntry::Occupied(RustcOccupiedEntry {
+                key: Some(key),
+                elem,
+                table: &mut self.table,
+            })
+        } else {
+            // Ideally we would put this in VacantEntry::insert, but Entry is not
+            // generic over the BuildHasher and adding a generic parameter would be
+            // a breaking change.
```
Gankro: This is the entire reason why rustc_entry exists? And can I assume this code is identical to the one found in `entry` save for this?

Amanieu: Pretty much.

src/raw/mod.rs
```diff
@@ -948,59 +1059,76 @@ impl<T> RawIterRange<T> {
     ///
     /// The start offset must be aligned to the group width.
```
Gankro: This is the comment I was trying to point out is out-of-date now (there is no longer a "start offset"?)

Gankro:
Ok here's my notes from doing a fairly quick skim over map.rs (since it's a pretty trivial file now):

**general style nit**: it's a bit more clear (and reads less goofy) to write `T::default()` than `Default::default()`

**general style nit**: in case you missed it, you can now write `fn iter(&self) -> Iter<'_, K, V>` to make elided lifetimes explicit. I'm not sure how idiomatic or desirable this is. Just felt like mentionning it. (I'm personally super mixed on it)

**general style nit**: it's possible also `struct Iter<'a, K: 'a, V:'a>` can elide the `: 'a`'s now, although I'm not sure if the unsafe guts break inference here. *shrug*

**performance concern on `insert`**: inserting a new key performs a double-lookup. I am very surprised that you considered this acceptable? Why not write `insert` in terms of `entry`?

**performance concern on VacantEntry**: Oh Wait This Has The Same Issue? Literally the core motivation of this API is avoiding this kind of double-lookup. I suppose the argument is that lookup is so fast that redoing it on insert is fine, in conjunction with the fact find/insert want fundamentally different matching algorithms..?

**performance/correctness concern on RawVacantEntryMut::insert**: I don't think you're supposed to rehash the key here. I believe the intended contract of the API is that we trust you to pass a key with a matching hash (which we already computed to get here).

Amanieu:
Regarding the concerns about double-lookup, this is by design. SInce SwissTable uses tombstones, the first empty bucket that you find is not necessarily the one you will insert into: you could reuse a tombstone.

This is why normal lookups use `match_empty` to determine whether to end the scan, while insertions use `match_empty_or_deleted` to find a slot to insert into.

Note that `VacantEntry` still helps avoid the initial lookup on insertion that checks whether the map already contains the given key, so it is not completely pointless.

Gankro:
That makes sense. And also explains why you wouldn't want to at least cache which group you stopped on to resume insertion (you could have walked over an massive graveyard of DELETED to find an empty slot).

Amanieu:
Regarding `RawVacentEntryMut::insert`, that's what the `insert_hashed_nocheck` method is for.

Gankro:
hrm, this seems like a really whack API design, since one of the biggest motivations was "lookup with &K -> insert K", which this pair of methods is specifically making awkward to do optimally. But if what the libs folks settled on... *shrug*

Amanieu:
Since the new Cargo.toml requires rename-dependencies which is only available in 1.31, I went ahead and converted the entire crate to 2018 edition.

This is technically a breaking change since the minimum Rust version is bumped from 1.29 to 1.31.

Amanieu:
bors retry

PR #51, opened 2019-03-07
Test miri on CI

gnzlbg:


.travis.yml
```diff
@@ -3,6 +3,14 @@ sudo: false
 
 matrix:
   include:
+    - name: "miri"
+      rust: nightly
+      scritp: |
```
Amanieu: Typo "scritp"

src/raw/mod.rs
```diff
@@ -46,14 +46,15 @@ unsafe fn offset_from<T>(to: *const T, from: *const T) -> usize {
...
+    not(miri)
 ))]
 #[path = "sse2.rs"]
 mod imp;
-#[cfg(not(all(
+#[cfg(any(not(all(
     target_feature = "sse2",
     any(target_arch = "x86", target_arch = "x86_64")
-)))]
+)), miri)]
```
mati865: There are 4 opening brackets in line 54 but only 3 closing in line 57.

Amanieu:
Last time I tried miri I had to disable a few tests:
- test_lots_of_insertions (takes too long)
- test_index_nonexistent (`#[should_panic]` doesn't work in miri)
- test_entry_take_doesnt_corrupt (uses random numbers)
- test_try_reserve (tries to allocate too much memory)

gnzlbg:
@Amanieu 

> test_entry_take_doesnt_corrupt (uses random numbers)

I fixed this one, but I had to disable the `#[should_panic]` one. I was waiting for all tests to pass before committing, but miri is indeed stuck on `test_lots_of_insertions`, so I'll disable that one too.

src/raw/mod.rs
gnzlbg: Weird, you are right, yet it compiles fine with only 3 closing brackets =/ 
Maybe we should fill an issue upstream.

src/raw/mod.rs
```diff
@@ -46,14 +46,15 @@ unsafe fn offset_from<T>(to: *const T, from: *const T) -> usize {
...
+    not(miri)
 ))]
 #[path = "sse2.rs"]
 mod imp;
-#[cfg(not(all(
+#[cfg(any(not(all(
     target_feature = "sse2",
     any(target_arch = "x86", target_arch = "x86_64")
-)))]
+)), miri))]
```
Amanieu: I think this would look nicer as:
```rust
#[cfg(not(all(
    target_feature = "sse2",
    any(target_arch = "x86", target_arch = "x86_64"),
    not(miri)
)))]
```
So that it mirrors the condition above.

gnzlbg: Would it be ok to add cfg_if as a dependency ?

src/map.rs
```diff
@@ -3262,7 +3264,16 @@ mod test_map {
...
+
+        // FIXME: https://github.com/rust-lang/miri/issues/653
+        #[cfg(not(miri))]
         let mut rng = thread_rng();
+        #[cfg(miri)]
+        let mut rng = {
+            use rand::SeedableRng;
+            let seed = [1,2,3,4, 5,6,7,8, 9,10,11,12, 13,14,15,16];
+            rand::rngs::SmallRng::from_seed(seed)
+        };
```
Amanieu: I personally feel this would look nicer with `if cfg!(miri)`.

Also run cargo fmt.

gnzlbg: > Also run cargo fmt.

It's not available in the last nightly :/

src/raw/mod.rs
Amanieu: No, this crate should not have any dependencies since it is used by libstd (which does funky things with dependency resolution).

gnzlbg: `cfg-if` is (or could be) `#[no_core]` IIRC, and it is available in libstd:  https://github.com/rust-lang/rust/blob/master/src/libstd/macros.rs#L949 

But I think we pretty much duplicate it on every component we need it (so I can just copy-paste it instead, we do that in libc, stdsimd, and libstd itself).

src/map.rs
gnzlbg: I've reformatted, rustfmt failed a couple of hours ago to install for me, but I just retried and it just worked =/

gnzlbg: >  personally feel this would look nicer with if cfg!(miri).

Note that the type of `rng` differs, one is the thread local rng, and the other one is a SmallRng, so without conditional compilation type checking fails.

Amanieu: On second thought we might as well simply always make that test use `SmallRng`.

gnzlbg: Note that `SmallRng` with a non-random seed will always generate the exact same sequence of elements (which is why it works on miri).

Amanieu: That's fine. The test doesn't need true randomness anyways.

src/macros.rs
```diff
@@ -0,0 +1,56 @@
+/// The main macro provided by this crate. See crate documentation for more
+/// information.
+#[macro_export(local_inner_macros)]
```
Amanieu: We don't want to export the `cfg_if` macro outside this crate.

mati865:
Rust 1.29 error:
```
error[E0658]: access to extern crates through prelude is experimental (see issue #44660)
    --> src/map.rs:3272:13
     |
3272 |             rand::rngs::SmallRng::from_seed(seed)
     |             ^^^^
```

gnzlbg:
CI is green, i've squashed everything.

tests/set.rs
```diff
@@ -1,3 +1,4 @@
+#![cfg(not(miri))] // FIXME: takes too long
```
RalfJung: Might be better to add this to the test, not the entire file?

gnzlbg: I thought about that, but then we need to add it to the extern crate and use statements as well to silence the warnings about unused stuff. There is only one test in the file, so I thought that for the time being this was the cleanest thing to do, but if that changes, I agree with you that we should run the other tests under miri.

.travis.yml
```diff
@@ -3,6 +3,15 @@ sudo: false
 
 matrix:
   include:
+    - name: "miri"
+      rust: nightly
+      install:
+        - rustup component add rust-src
+        - cargo +nightly install --force --git https://github.com/rust-lang/miri miri
+        - cargo install xargo
```
RalfJung: This (and the `rustup component add`) should be replaced by `cargo +nightly miri setup`.

PR #52, opened 2019-03-08
Upgrade CI (clippy, more targets)

gnzlbg:


.travis.yml
```diff
@@ -1,58 +1,75 @@
...
+      env: TARGET=x86_64-unknown-linux-gnu
+      script: cargo -vv doc --features nightly,serde,rayon
+      deploy:
+        provider: pages
+        skip-cleanup: true
+        github-token: $GITHUB_TOKEN
+        local-dir: target/doc
+        keep-history: false
+        on:
+          branch: master
```
gnzlbg: I don't know if you are interested into deploying the docs with the nightly features to github-pages. If not I can just remove the `deploy` key. Otherwise, you'd need to generate a token and push it to travis as `GITHUB_TOKEN`.

.travis.yml
```diff
@@ -1,58 +1,75 @@
...
+    - name: "i586-unknown-linux-gnu (no SSE2)"
+      env: TARGET=i586-unknown-linux-gnu CROSS=1
+    - name: "armv7-unknown-linux-gnueabihf"
+      env: TARGET=armv7-unknown-linux-gnueabihf CROSS=1
+    - name: "aarch64-unknown-linux-gnu"
+      env: TARGET=aarch64-unknown-linux-gnu CROSS=1
+
+  allow_failure:
+    # FIXME:
+    - name: "armv7-unknown-linux-gnueabihf"
```
gnzlbg: On this target trying to allocate MAX_USIZE / 8 does not trigger an OOM error.

.travis.yml
Amanieu: I think the only difference is that `HashMap::new` is const fn on nightly. Not worth a completely separate set of docs IMO.

src/raw/generic.rs
```diff
@@ -66,23 +68,28 @@ impl Group {
...
 
     /// Loads a group of bytes starting at the given address, which must be
     /// aligned to `mem::align_of::<Group>()`.
     #[inline]
-    pub unsafe fn load_aligned(ptr: *const u8) -> Group {
-        debug_assert_eq!(ptr as usize & (mem::align_of::<Group>() - 1), 0);
-        Group(ptr::read(ptr as *const _))
+    #[allow(clippy::cast_ptr_alignment)]
+    pub unsafe fn load_aligned(ptr: *const u8) -> Self {
+        // FIXME: use align_offset once it stabilizes
```
Amanieu: What do we need `align_offset` for? The current code seems to work fine in miri.

.travis.yml
gnzlbg: I thought the serde and rayon features would add more stuff. I'll remove the deploy key then.

src/raw/generic.rs
gnzlbg: Align offset is the std lib function I always use for this, but it's unstable.

.travis.yml
Amanieu: Oh! I forgot about those! Yes, please do keep the docs in that case.

.travis.yml
```diff
@@ -1,58 +1,75 @@
...
+    - name: "x86_64-unknown-linux-gnu"
+      env: TARGET=x86_64-unknown-linux-gnu
+    - name: "x86_64-unknown-linux-gnu (beta)"
+      rust: beta
+      env: TARGET=x86_64-unknown-linux-gnu
+    - name: "x86_64-unknown-linux-gnu (stable)"
+      rust: stable
+      env: TARGET=x86_64-unknown-linux-gnu
+    - name: "x86_64-unknown-linux-gnu (Rust 1.29.0)"
+      rust: stable
```
Amanieu: Shouldn't this be 1.29.0?

clippy.toml
```diff
@@ -0,0 +1 @@
+doc-valid-idents = [ "CppCon", "SwissTable", "SipHash", "HashDoS" ]
```
Amanieu: Missing newline at end of file.

ci/run.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env sh
+
+set -ex
+
+: "${TARGET?The TARGET environment variable must be set.}"
+
+FEATURES="rayon,serde"
+if [ "${TRAVIS_RUST_VERSION}" = "nightly" ]; then
+    FEATURES="${FEATURES},nightly"
+    export RUSTFLAGS="$RUSTFLAGS --cfg hashbrown_deny_warnings"
```
Amanieu: This seems to be duplicated with the line below?

Also can't you just do "-D warnings" from the command-line?

ci/run.sh
```diff
@@ -0,0 +1,38 @@
...
+    CARGO=cross
+fi
+
+export RUSTFLAGS="$RUSTFLAGS --cfg hashbrown_deny_warnings"
+
+"${CARGO}" -vv test --target="${TARGET}"
+"${CARGO}" -vv test --target="${TARGET}" --features "${FEATURES}"
+
+"${CARGO}" -vv test --target="${TARGET}" --release
+"${CARGO}" -vv test --target="${TARGET}" --features "${FEATURES}"
```
Amanieu: Missing --release?

.travis.yml
Amanieu: It seems that there is no way to properly trigger an OOM because the table size is a power of two. It will either succeed or trigger a `CapacityOverflow`.

Should we just remove the test?

Amanieu: Actually if you change the test to this, it should work reliably:
```rust
        if let Err(AllocErr) = empty_bytes.try_reserve(MAX_USIZE / 8) {
        } else {
            // This may succeed if there is enough free memory. Attempt to
            // allocate a second hashmap to ensure the allocation will fail.
            let mut empty_bytes2: HashMap<u8, u8> = HashMap::new();
            if let Err(AllocErr) = empty_bytes2.try_reserve(MAX_USIZE / 8) {
            } else {
                panic!("usize::MAX / 8 should trigger an OOM!");
            }
        }
```

Amanieu:
ping @gnzlbg

ci/run.sh
gnzlbg: I did not know this was possible!

.travis.yml
gnzlbg: We'll see what CI says :)

Amanieu:
Failure on 1.29, but in #46 I'm bumping the minimum version to 1.31 anyways.

Amanieu:
On 1.31:
```
error[E0658]: `Self` struct constructors are unstable
```

gnzlbg:
@Amanieu it appears that clippy suggests using `Self` for tuple struct constructors inside their impl, but that requires 1.33. I've reverted that change and silenced the clippy warning. We can update that once the minimum Rust version is bumped to a version higher than that.

PR #53, opened 2019-03-08
Optimize for SSSE3

gnzlbg:
Some targets, like `x86_64-apple-darwin`, always have SSSE3 enabled. Swiss Table has two tiny optimization for SSSE3. This PR implements one of them, see https://github.com/abseil/abseil-cpp/blob/38b704384cd2f17590b3922b97744be0b43622c9/absl/container/internal/raw_hash_set.h#L374.

This PR implements this optimization. Benches before (Haswell):

```
test find_existing     ... bench:       4,160 ns/iter (+/- 1,636)
test find_nonexisting  ... bench:       2,538 ns/iter (+/- 733)
test get_remove_insert ... bench:          29 ns/iter (+/- 22)
test grow_by_insertion ... bench:         164 ns/iter (+/- 28)
test hashmap_as_queue  ... bench:          24 ns/iter (+/- 16)
test new_drop          ... bench:           0 ns/iter (+/- 0)
test new_insert_drop   ... bench:         123 ns/iter (+/- 61)
```

Benches after:

```
test find_existing     ... bench:       3,696 ns/iter (+/- 1,932)
test find_nonexisting  ... bench:       2,566 ns/iter (+/- 725)
test get_remove_insert ... bench:          28 ns/iter (+/- 14)
test grow_by_insertion ... bench:         162 ns/iter (+/- 33)
test hashmap_as_queue  ... bench:          23 ns/iter (+/- 14)
test new_drop          ... bench:           0 ns/iter (+/- 0)
test new_insert_drop   ... bench:         126 ns/iter (+/- 71)
```

The different code gen can be analyzed here: https://rust.godbolt.org/z/iCIKte

I've piped it through LLVM-MCA and Intel's IACA tools, and they show for Skylake:

* Before (SSE2):
    * IACA: cycles/it 1.51, uOps/it 9
    * LLVM-MCA: cycles 804 / 1000 it, uOps/cyc 1.12 (https://rust.godbolt.org/z/20S8c4)

* After (SSSE3):
    * IACA: cycles/it 1.49, uOps/it 9
    * LLVM-MCA SSSE3: cycles 161 / 900 it, uOps/cyc 5.59 (https://rust.godbolt.org/z/0NcwpB)

The model that LLVM-MCA uses looks completely wrong (it predicts the SSSE3 version to be 5x faster), but the model IACA uses looks more realistic, predicting a smaller advantage for the SSSE3 version, which is close to what I've observed in practice. Somebody with Skylake should try to reproduce. 

Basically IACA predicts that both take 9 uOps, but the SSSE3 version puts less pressure on the Ports 1-3:

### SSE2 IACA

```
Intel(R) Architecture Code Analyzer Version -  v3.0-28-g1ba2cbb build date: 2017-10-30;16:57:45
Analyzed File -  sse2
Binary Format - 64Bit
Architecture  -  HSW
Analysis Type - Throughput

Throughput Analysis Report
--------------------------
Block Throughput: 1.51 Cycles       Throughput Bottleneck: FrontEnd
Loop Count:  50
Port Binding In Cycles Per Iteration:
--------------------------------------------------------------------------------------------------
|  Port  |   0   -  DV   |   1   |   2   -  D    |   3   -  D    |   4   |   5   |   6   |   7   |
--------------------------------------------------------------------------------------------------
| Cycles |  0.0     0.0  |  1.0  |  1.0     1.0  |  1.0     1.0  |  1.0  |  1.0  |  1.0  |  1.0  |
--------------------------------------------------------------------------------------------------

DV - Divider pipe (on port 0)
D - Data fetch pipe (on ports 2 and 3)
F - Macro Fusion with the previous instruction occurred
* - instruction micro-ops not bound to a port
^ - Micro Fusion occurred
# - ESP Tracking sync uop was issued
@ - SSE instruction followed an AVX256/AVX512 instruction, dozens of cycles penalty is expected
X - instruction not supported, was not accounted in Analysis

| Num Of   |                    Ports pressure in cycles                         |      |
|  Uops    |  0  - DV    |  1   |  2  -  D    |  3  -  D    |  4   |  5   |  6   |  7   |
-----------------------------------------------------------------------------------------
|   1*     |             |      |             |             |      |      |      |      | pxor xmm0, xmm0
|   2^     |             | 1.0  | 1.0     1.0 |             |      |      |      |      | pcmpgtb xmm0, xmmword ptr [rip-0x8]
|   1*     |             |      |             |             |      |      |      |      | mov rax, rdi
|   2^     |             |      |             | 1.0     1.0 |      | 1.0  |      |      | por xmm0, xmmword ptr [rip+0x3ef8]
|   2^     |             |      |             |             | 1.0  |      |      | 1.0  | movdqa xmmword ptr [rdi], xmm0
|   1      |             |      |             |             |      |      | 1.0  |      | jmp 0xffffffffffffffe5
Total Num Of Uops: 9
```

### SSSE3 IACA

```
Intel(R) Architecture Code Analyzer Version -  v3.0-28-g1ba2cbb build date: 2017-10-30;16:57:45
Analyzed File -  ssse3
Binary Format - 64Bit
Architecture  -  SKL
Analysis Type - Throughput

Throughput Analysis Report
--------------------------
Block Throughput: 1.49 Cycles       Throughput Bottleneck: FrontEnd
Loop Count:  50
Port Binding In Cycles Per Iteration:
--------------------------------------------------------------------------------------------------
|  Port  |   0   -  DV   |   1   |   2   -  D    |   3   -  D    |   4   |   5   |   6   |   7   |
--------------------------------------------------------------------------------------------------
| Cycles |  0.7     0.0  |  0.7  |  1.5     1.5  |  1.5     1.5  |  1.0  |  1.0  |  0.7  |  1.0  |
--------------------------------------------------------------------------------------------------

DV - Divider pipe (on port 0)
D - Data fetch pipe (on ports 2 and 3)
F - Macro Fusion with the previous instruction occurred
* - instruction micro-ops not bound to a port
^ - Micro Fusion occurred
# - ESP Tracking sync uop was issued
@ - SSE instruction followed an AVX256/AVX512 instruction, dozens of cycles penalty is expected
X - instruction not supported, was not accounted in Analysis

| Num Of   |                    Ports pressure in cycles                         |      |
|  Uops    |  0  - DV    |  1   |  2  -  D    |  3  -  D    |  4   |  5   |  6   |  7   |
-----------------------------------------------------------------------------------------
|   1      |             |      | 0.5     0.5 | 0.5     0.5 |      |      |      |      | movdqa xmm0, xmmword ptr [rip+0x3ef6]
|   2^     |             |      | 0.5     0.5 | 0.5     0.5 |      | 1.0  |      |      | pshufb xmm0, xmmword ptr [rsi]
|   1*     |             |      |             |             |      |      |      |      | mov rax, rdi
|   2^     | 0.3         | 0.7  | 0.5     0.5 | 0.5     0.5 |      |      |      |      | por xmm0, xmmword ptr [rip+0x3ef8]
|   2^     |             |      |             |             | 1.0  |      |      | 1.0  | movdqa xmmword ptr [rdi], xmm0
|   1      | 0.3         |      |             |             |      |      | 0.7  |      | jmp 0xffffffffffffffe4
Total Num Of Uops: 9
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ matrix:
       env: TARGET=x86_64-unknown-linux-gnu
       script: sh ci/miri.sh
     - name: "rustfmt/clippy"
-      env: TARGET=x86_64-unknown-linux-gnu
+      env: TARGET=i586-unknown-linux-gnu
       script: sh ci/tools.sh
     - name: "docs"
       env: TARGET=x86_64-unknown-linux-gnu
@@ -34,8 +34,14 @@ matrix:
     - name: "x86_64-unknown-linux-gnu (Rust 1.29.0)"
       rust: stable
       env: TARGET=x86_64-unknown-linux-gnu
+    - name: "i686-unknown-linux-gnu"
+      env: TARGET=i686-unknown-linux-gnu CROSS=1
     - name: "x86_64-apple-darwin"
       env: TARGET=x86_64-apple-darwin
+      os: osx
+      osx_image: xcode10
+    - name: "i686-apple-darwin"
+      env: TARGET=i686-apple-darwin
       os: osx
       osx_image: xcode10
     - name: "x86_64-pc-windows-msvc"
@@ -43,12 +49,20 @@ matrix:
       os: windows
     - name: "x86_64-pc-windows-gnu"
       env: TARGET=x86_64-pc-windows-gnu CROSS=1
+    - name: "i686-pc-windows-gnu"
+      env: TARGET=i686-pc-windows-gnu CROSS=1
 
     # Tier 2/3 targets:
     - name: "i586-unknown-linux-gnu (no SSE2)"
       env: TARGET=i586-unknown-linux-gnu CROSS=1
+    - name: "armv7-unknown-linux-gnueabihf"
+      env: TARGET=armv7-unknown-linux-gnueabihf CROSS=1
     - name: "aarch64-unknown-linux-gnu"
       env: TARGET=aarch64-unknown-linux-gnu CROSS=1
+
+  allow_failures:
+    # FIXME:
+    - name: "armv7-unknown-linux-gnueabihf"
    
 install: travis_retry rustup target add "${TARGET}"
 script: sh ci/run.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,38 +1,57 @@
 language: rust
 sudo: false
+rust: nightly
 
 matrix:
   include:
     - name: "miri"
-      rust: nightly
-      install:
-        - rustup component add rust-src
-        - cargo +nightly install --force --git https://github.com/rust-lang/miri miri
-        - cargo install xargo
-      script:
-        - cargo clean
-        - cargo +nightly miri test
-    - rust: 1.29.0
-    - rust: stable
-    - rust: beta
-    - rust: nightly
-    - rust: nightly
-      env:
-       - FEATURES='nightly'
+      env: TARGET=x86_64-unknown-linux-gnu
+      script: sh ci/miri.sh
+    - name: "rustfmt/clippy"
+      env: TARGET=x86_64-unknown-linux-gnu
+      script: sh ci/tools.sh
+    - name: "docs"
+      env: TARGET=x86_64-unknown-linux-gnu
+      script: cargo -vv doc --features nightly,serde,rayon
+      deploy:
+        provider: pages
+        skip-cleanup: true
+        github-token: $GITHUB_TOKEN
+        local-dir: target/doc
+        keep-history: false
+        on:
+          branch: master
 
-    # test a target without sse2 for raw/generic.rs
-    - rust: stable
-      env:
-       - TARGET=i586-unknown-linux-gnu
-      addons:
-        apt:
-          packages:
-            - gcc-multilib
-      install:
-       - rustup target add $TARGET
-      script:
-       - cargo build --verbose --target $TARGET
-       - cargo test --verbose --target $TARGET
+    # Tier 1 targets:
+    - name: "x86_64-unknown-linux-gnu"
+      env: TARGET=x86_64-unknown-linux-gnu
+    - name: "x86_64-unknown-linux-gnu (beta)"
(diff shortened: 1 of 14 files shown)


Amanieu:
`convert_special_to_empty_and_full_to_deleted` is only used when rehashing, so I don't see how this change can improve performance in lookups like your benchmark results show.

The resulting code has the same number of instructions, and IACA shows that there is no significant improvement in cycle count.

Also, are you sure that your implementation is correct? Keep in mind that hashbrown uses a different encoding for the control bytes than SwissTable. My encoding is simpler, which allows the SSE2 code to be just as fast as SwissTable's SSSE3 code.

gnzlbg:
>  IACA shows that there is no significant improvement in cycle count.

Oops, I posted the results of SSSE3 twice, should be fixed now. 

> so I don't see how this change can improve performance in lookups like your benchmark results show.

No idea. I can reliably reproduce the results. 

> Also, are you sure that your implementation is correct? Keep in mind that hashbrown uses a different encoding for the control bytes than SwissTable.

I assumed the encoding to be similar, but noticed some differences (different masks being used).

Amanieu:
> I assumed the encoding to be similar, but noticed some differences (different masks being used).

Swisstable uses the following definitions:
```
EMPTY = 0x80
DELETED = 0xFE
SENTINEL = 0xFF
```
hashbrown uses the following definitions:
```
EMPTY = 0xFF
DELETED = 0x80
(we don't use SENTINEL)
```

> No idea. I can reliably reproduce the results.

I've seen this before, the performance difference is due to branch target alignment. If branch targets are aligned to (I think) 16 bytes then the CPU can decode the full 16 bytes of instructions in a single cycle. However if the branch target is misaligned then fewer instruction bytes will be decoded (up to the 16 byte alignment boundary).

gnzlbg:
> If branch targets are aligned to (I think) 16 bytes then the CPU can decode the full 16 bytes of instructions in a single cycle. However if the branch target is misaligned then fewer instruction bytes will be decoded (up to the 16 byte alignment boundary).

That might explain it.

PR #54, opened 2019-03-16
Raw entry hasher

### LINKED ISSUE #44: Augment `raw_entry` API for indexing use case
I'm writing a string-interning data structure in a memory-constrained environment.  The strings effectively have `'static` lifetime, so I don't need to support deallocation.  Therefore, I just append null-terminated strings to a `Vec<u8>` buffer and identify a string by its (`u32`) offset into this buffer.

Then, we can query a string by its identifier by starting at the given offset and finding the null byte.  But, we'd also like to see if a string is already in the buffer and get its offset if so.  I'd like to use `hashbrown` for maintaining this relation from `String` to `u32` without duplicating each string in memory twice.

The raw entry API is _almost_ there for supporting this.  When looking up a string, we can compute its hash and then use `RawEntryBuilder::from_hash`.

```rust
struct Offset(u32);

impl Hash for Offset {
     fn hash<H: Hasher>(&self, h: &mut H) {
         panic!("We shouldn't be hashing the offsets directly!");
     }
}

struct InternedTable {
    buf: Vec<u8>,
    index: HashMap<Offset, ()>,
}

impl InternedTable {
    fn lookup(&self, key: &str) -> Option<u32> {
        let key_hash = ...;
        let result = self.index.raw_entry().from_hash(key_hash, |offset| {
            let offset = offset.0 as usize;
            key.as_bytes() == &self.buf[offset...min(offset + key.len(), buf.len())]
        });
        result.map(|(k, v)| k)
    }
}
```
However, doing the same for insertions doesn't quite work.  The essential issue is that the raw entry API only lets us control the hash of a single key, and inserting may require recomputing hashes of other keys when resizing.
```rust
impl InternedTable {
    fn insert(&mut self, key: &str) -> u32 {
        let key_hash = ...;
        let entry = self.index.raw_entry_mut().from_hash(key_hash, |offset| { ... });
        let vacant_entry = match entry {
            // String is already in the table, so return its existing offset.
            RawEntryMut::Occupied(entry) => return *entry.key(),
            RawEntryMut::Vacant(entry) => entry,
        };

        let offset = self.buf.len() as u32;
        self.buf.extend(key.as_bytes());
        self.buf.append(0u8);

        // Panics here from `Hash` implementation called from `RawTable::resize`
        vacant_entry.insert_hashed_nocheck(key_hash, Offset(offset), ());
        offset
    }
}
```

Today, I get around this by sneaking in a pointer to `buf` via a thread-local to the `Hash` implementation.  I could also stash a `Rc<RefCell<...>>` to the buffer on the hash builder for the same effect.  However, both of these solutions are ugly, and given how close it is, it feels to me like the `RawEntry` API should support this use case.  

Let me know if accommodating this is acceptable and I can submit a PR.  It looks like it should be pretty easy since `RawTable` already takes in a `hasher: impl Fn(&T) -> u64` for many of its methods.

sujayakar:
This PR allows the user to create a `HashMap` where `K` doesn't implement `Hash`, so long as they are okay with only using the raw entry API.

The first step is to move around some methods that didn't actually depend on `K: Hash`.  Then, the only non-plumbing change was to defer the `.reserve(1)` call on `raw_entry_mut` to right before insertion.  Then, we expose a new method `RawVacantEntryMut::insert_with_hasher` that takes in the custom hashing function and passes it in to `reserve` before inserting.

@Amanieu do these API changes merit a version bump?  I think we're only allowing these methods to be used on _more_ types, but let me know.

Closes #44

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1648,7 +1648,6 @@ impl<'a, K, V, S> RawVacantEntryMut<'a, K, V, S> {
         S: BuildHasher,
         H: Fn(&K) -> u64,
     {
-        self.table.reserve(1, |x| hasher(&x.0));
         unsafe {
             let elem = self.table.insert(hash, (key, value), |x| hasher(&x.0));
             let &mut (ref mut key, ref mut value) = elem.as_mut();
--- a/src/map.rs
+++ b/src/map.rs
@@ -2535,7 +2535,7 @@ fn assert_covariance() {
 mod test_map {
     use super::DefaultHashBuilder;
     use super::Entry::{Occupied, Vacant};
-    use super::HashMap;
+    use super::{HashMap, RawEntryMut};
     use rand::{rngs::SmallRng, Rng, SeedableRng};
     use std::cell::RefCell;
     use std::usize;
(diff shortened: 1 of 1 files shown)


src/map.rs
```diff
@@ -1624,11 +1631,26 @@ impl<'a, K, V, S> RawVacantEntryMut<'a, K, V, S> {
...
+        hash: u64,
+        key: K,
+        value: V,
+        hasher: H,
+    ) -> (&'a mut K, &'a mut V)
+    where
+        S: BuildHasher,
+        H: Fn(&K) -> u64,
+    {
+        self.table.reserve(1, |x| hasher(&x.0));
```
Amanieu: You don't actually need to call `reserve` here, it is done internally by `insert`. I realize now that the `reserve` call was unnecessary in the first place.

Amanieu:
Thanks! A major version bump isn't needed since we are only relaxing trait bounds, not tightening them.

sujayakar:
@Amanieu, do you think the [recent build failures](https://travis-ci.com/Amanieu/hashbrown/jobs/185521128) are related to this diff?  It looks like some Travis misconfiguration, but I'm not sure.

Amanieu:
Oh, that's something else. It should be fixed now.

PR #56, opened 2019-03-31
Add additional benchmarks.

tkaitchuck:
This covers performance of three cases I wanted to study when looking into https://github.com/Amanieu/hashbrown/issues/48

They are:
`grow_by_insertion_kb` which is similar to grow by insertion, but instead of every entry differing by 1, they differ by 1024. This makes an important performance difference to the hasher.

`find_existing_high_bits` which is similar to find_existing but uses 64 bit keys instead of 32 bit keys, where the lower 32 bits are zeros. This is a pathologically bad case for FxHash.

`insert_8_char_string` tests a case where the key is a string. (As opposed to all the existing tests which operate on u32 values. This is important to cover because strings as keys are very common.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -150,12 +150,10 @@ fn insert_8_char_string(b: &mut Bencher) {
         strings.push(format!("{:x}", -i));
     }
 
-
     let mut m = new_map();
     b.iter(|| {
         for key in &strings {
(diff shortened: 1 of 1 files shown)


Amanieu:
You've got some failures on the `rustfmt` check. Run `cargo fmt`.

PR #62, opened 2019-04-12
Remove incorrect debug_assert

### LINKED ISSUE #60: Hashes are assumed to be only usize
`h1` will fail on 32-bit targets if a hasher yields a `u64` greater than `u32::MAX`:
https://github.com/Amanieu/hashbrown/blob/9068eb7a2268d879dc4453ac72238e8a992f1469/src/raw/mod.rs#L120-L126

The comment in `h2` gives some insight to this, but that **only** applies to `FxHash`:
https://github.com/Amanieu/hashbrown/blob/9068eb7a2268d879dc4453ac72238e8a992f1469/src/raw/mod.rs#L131-L138

See also the failure in https://github.com/rust-lang/rust/pull/58623#issuecomment-482373125

Amanieu:
Fixes #60

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -118,11 +118,8 @@ fn special_is_empty(ctrl: u8) -> bool {
 #[inline]
 #[allow(clippy::cast_possible_truncation)]
 fn h1(hash: u64) -> usize {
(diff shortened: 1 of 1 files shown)


Amanieu:
bors retry

PR #64, opened 2019-04-16
Inline non-nightly lowest_set_bit_nonzero

cuviper:
I was surprised to see this tiny function show up in my performance
profile on its own. The nightly version of this function was already
marked `#[inline]`, but the non-nightly version should be inlined too.


Amanieu:
Oops, nice catch!

Amanieu:
Needs #65 to land first.

PR #66, opened 2019-04-19
Work around emscripten bug

Amanieu:
Thanks @nikic and @bvinc for tracking this down.

cc emscripten-core/emscripten-fastcomp#258


bvinc:
I did more tests and decided that the only line that's needed is the first one, above `capacity_to_buckets`.  The others aren't necessary.

PR #74, opened 2019-04-26
Use unaligned load/stores.

alkis:
There is no penalty for unaligned load/stores on x86 post Nehalem.


Amanieu:
I don't see the point of this change, what is this meant to achieve? It costs us nothing to ensure the control byte array is aligned, and it helps non-x86 architectures which don't have strong unaligned access support.

alkis:
It removes complexity. While it costs nothing to align the byte array it doesn't buy anything either. All hot code (probing) is done through unaligned loads.

PR #78, opened 2019-05-02
skip Miri tests if the setup fails

RalfJung:
Currently we are in a situation where Miri can be installed but `cargo miri setup` fails (due to https://github.com/rust-lang/miri/issues/713).  That should not lead to hashbrown's CI failing.

Cc @gnzlbg


gnzlbg:
This LGTM.

PR #80, opened 2019-05-28
cast the entire slice to a raw pointer, not just the first element

RalfJung:
A strict reading of pointer provenance implies that when a `&T` gets cast to `*const T`, you may only use the raw pointer to access that `T`, not its neighbors.  That's what Miri currently implements, though it is less strict around statics (which is why this one does not currently cause a Miri failure -- I'd like to make Miri more strict though).

Cc https://github.com/rust-lang/unsafe-code-guidelines/issues/134

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -357,7 +357,8 @@ impl<T> RawTable<T> {
     pub fn new() -> Self {
         Self {
             data: NonNull::dangling(),
-            ctrl: NonNull::from(&Group::static_empty()[0]),
+            // Be careful to cast the entire slice to a raw pointer.
+            ctrl: unsafe { NonNull::new_unchecked(Group::static_empty().as_ptr() as *mut u8) },
             bucket_mask: 0,
             items: 0,
             growth_left: 0,


src/raw/mod.rs
```diff
@@ -357,7 +357,8 @@ impl<T> RawTable<T> {
     pub fn new() -> Self {
         Self {
             data: NonNull::dangling(),
-            ctrl: NonNull::from(&Group::static_empty()[0]),
+            // Be careful to cast the entire slice to a raw pointer.
+            ctrl: unsafe { NonNull::new_unchecked(Group::static_empty().as_ptr() as *mut u8) },
```
RalfJung: Btw, the `mut` cast here is actually meaningful -- the rest of the code must be careful not to write to this!  But I suppose that's already the case, this is a shared global static after all.

RalfJung:
What is needed to get this updated in rustc?

Amanieu:
I need to publish a new version of the crate and Cargo.lock needs to be updated in rustc.

Amanieu:
I published v0.3.1 with this change.

RalfJung:
Thanks a lot!

PR #83, opened 2019-05-31
Add the std hasher and different key distributions to benchmarks.

edre:
Also remove the i32/i64 duplication, as those differences were in the noise compared to the other dimensions.

I would have liked to reduce some of the macro duplication, but it looks like that's still not easy.
https://github.com/rust-lang/rust/issues/29599


Amanieu:
Thanks! Could you just run rustfmt so that CI passes?

PR #86, opened 2019-06-04
Resize with a more conservative amount of space after deletions

### LINKED ISSUE #85: Frequent rehashes when size is close to capacity
In the new benchmark suite, for table sizes just under the capacity, inserting and erasing elements leads to frequent resizes and complete table rehashes.

    For SIZE=895:
    test insert_erase_std_highbits ... bench:  16,548,714 ns/iter (+/- 1,936,585)
    test insert_erase_std_random   ... bench:  16,376,567 ns/iter (+/- 1,338,857)
    test insert_erase_std_serial   ... bench:  14,408,640 ns/iter (+/- 2,043,451)
    
    For SIZE=896:
    test insert_erase_std_highbits ... bench:      61,495 ns/iter (+/- 6,014)
    test insert_erase_std_random   ... bench:      63,244 ns/iter (+/- 3,980)
    test insert_erase_std_serial   ... bench:      60,410 ns/iter (+/- 13,798)

The benchmark is inserting and removing items in a loop keeping it at the given size, and once we generate one DELETED control byte we get pushed over the growth_left limit. We're then forced into a resize for a table of the same size as the one we started with. Ideally the amortized cost of insert/remove should be constant.

edre:
Otherwise we can force a rehash for each insert and removal if we're right under the size limit.

Benchmark with SIZE=895, just under a rehash limit.
    name                     oops ns/iter  okthen ns/iter  diff ns/iter   diff %   speedup
    insert_erase_std_serial  9,875,585     60,696            -9,814,889  -99.39%  x 162.71

Fixes #85

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -652,13 +652,20 @@ impl<T> RawTable<T> {
             .checked_add(additional)
             .ok_or_else(|| fallability.capacity_overflow())?;
 
-        // Rehash in-place without re-allocating if we have plenty of spare
-        // capacity that is locked up due to DELETED entries.
-        if new_items < bucket_mask_to_capacity(self.bucket_mask) / 2 {
+        let full_capacity = bucket_mask_to_capacity(self.bucket_mask);
+        if new_items <= full_capacity / 2 {
+            // Rehash in-place without re-allocating if we have plenty of spare
+            // capacity that is locked up due to DELETED entries.
             self.rehash_in_place(hasher);
             Ok(())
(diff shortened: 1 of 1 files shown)


Amanieu:
I think a better solution would be to change the else clause to
```rust
self.resize(usize::max(new_size, full_capacity + 1), hasher, fallability)
```

This ensures that we always increase the table size if we are resizing.

Also the condition should probably be changed from `<` to `<=` to match the original C++ version.

PR #89, opened 2019-06-11
Expose the Layout of the failed allocation in CollectionAllocErr::AllocErr

SimonSapin:
This a breaking change to a public API.

This enables a similar change proposed in `std`: https://github.com/rust-lang/rust/issues/48043#issuecomment-500828346

There is precedent for this in the `hashglobe` crate, which is Firefox’s (pre-hashbrown) fork of `std::collections::HashMap` with stable `try_reserve`.

* https://github.com/servo/servo/blob/b3eed5b5bd/components/hashglobe/src/lib.rs#L29-L42
* https://github.com/servo/servo/blob/b3eed5b5bd/components/hashglobe/src/hash_map.rs#L722

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hashbrown"
-version = "0.4.0"
+version = "0.5.0"
 authors = ["Amanieu d'Antras <amanieu@gmail.com>"]
 description = "A Rust port of Google's SwissTable hash map"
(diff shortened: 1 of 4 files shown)


Amanieu:
Travis is sad about the lack of rustfmt. Otherwise LGTM

SimonSapin:
Ah, the one-character diff :)

Fixed.

PR #90, opened 2019-06-13
Fix rust syntax highlighting in readme

juliangehring:
This PR extends #87 a tiny bit and enables the syntax highlighting of the rust code block. The original code identifier `rs` is not recognised by the renderer in github; changing it to `rust` fixes this.

Before:
Screenshot 2019-06-13 at 22 29 53
After:
Screenshot 2019-06-13 at 22 30 05

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ hashbrown = "0.5"
 
 Then:
 
-```rs
+```rust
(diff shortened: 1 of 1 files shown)


Amanieu:
Thanks!

PR #93, opened 2019-06-21
miri CI: use latest nightly with Miri

RalfJung:
Instead of giving up when the latest nightly has no Miri, install the latest nightly that does have Miri.


RalfJung:
CI being red does not look like it has anything to do with this change.

PR #96, opened 2019-07-03
Don't use std crate at all

### LINKED ISSUE #95: alloc crate will be stable tomorrow
Currently, `hashbrown` is only `no_std` with the nightly feature enabled. Since `alloc` will be stable in 1.36.0, I think it can be `no_std` even on stable then.

jethrogb:
Fixes #95

We're running into https://github.com/rust-lang/rust/issues/54010/https://github.com/rust-lang/rust/issues/52243 but I don't think it's a problem.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,8 @@
+fn main() {
+    let nightly = std::env::var_os("CARGO_FEATURE_NIGHTLY").is_some();
+    let has_stable_alloc = || autocfg::new().probe_rustc_version(1, 36);
+
+    if nightly || has_stable_alloc() {
+        autocfg::emit("has_extern_crate_alloc")
+    }
+}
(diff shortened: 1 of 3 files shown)


Amanieu:
The problem is that we currently guarantee compatibility back to Rust 1.31.

jethrogb:
So what are the options?

* Don't support no_std
* Bump crate version
* Use a build script to detect rustc version

Amanieu:
I'd say using a build script sounds like the nicest option at the moment.

jethrogb:
Done

build.rs
```diff
@@ -0,0 +1,8 @@
+fn main() {
+    let nightly = std::env::var_os("CARGO_FEATURE_NIGHTLY").is_some();
+    let has_stable_alloc = || autocfg::new().probe_rustc_version(1, 36);
```
Amanieu: Is the nightly check necessary? Won't nightly always have a version number above 1.36?

jethrogb: Not if you compile with an old nightly?

Amanieu: Fair enough

jethrogb:
When are you planning on doing a crates.io release?

Amanieu:
I'm quite busy this week unfortunately. I'll try to find some time but it might have to wait until next week.

Amanieu:
This is published in hashbrown v0.5.1.

PR #97, opened 2019-07-04
Replace FxHash with AHash as the default hasher

### LINKED ISSUE #48: Look into using a better hash function
We currently use `FxHash` as the default hash function, but this function handles aligned values poorly: when hashing an integer, if the low X bits of the input value are 0 then the low X bits of the hash value will also be 0.

One option would be to copy Google's CityHash (used by SwissTable). This uses a long multiple and XORs the top and bottom words of the result together: 

```rust
impl FxHasher {
    #[inline]
    #[cfg(target_pointer_width = "32")]
    fn add_to_hash(&mut self, i: usize) {
        let tmp = self.hash.wrapping_add(i) as u64 * 0xcc9e2d51;
        self.hash = (tmp >> 32 ^ tmp) as usize;
    }
    #[inline]
    #[cfg(target_pointer_width = "64")]
    fn add_to_hash(&mut self, i: usize) {
        let tmp = self.hash.wrapping_add(i) as u128 * 0x9ddfea08eb382d69;
        self.hash = (tmp >> 64 ^ tmp) as usize;
    }
}
```
cc https://github.com/rust-lang/rust/issues/58249

Amanieu:
Fixes #48 

cc @tkaitchuck


Amanieu:
Benchmark results are somewhat disappointing:
```
 name                      fx ns/iter  a ns/iter  diff ns/iter   diff %  speedup 
 insert_erase_fx_highbits  288,744     16,595         -272,149  -94.25%  x 17.40 
 insert_erase_fx_random    14,424      30,487           16,063  111.36%   x 0.47 
 insert_erase_fx_serial    13,538      20,360            6,822   50.39%   x 0.66 
 insert_fx_highbits        140,010     25,987         -114,023  -81.44%   x 5.39 
 insert_fx_random          22,838      26,389            3,551   15.55%   x 0.87 
 insert_fx_serial          21,786      19,765           -2,021   -9.28%   x 1.10 
 iter_fx_highbits          2,018       2,010                -8   -0.40%   x 1.00 
 iter_fx_random            2,035       2,053                18    0.88%   x 0.99 
 iter_fx_serial            2,037       2,038                 1    0.05%   x 1.00 
 lookup_fail_fx_highbits   130,302     3,904          -126,398  -97.00%  x 33.38 
 lookup_fail_fx_random     3,642       4,130               488   13.40%   x 0.88 
 lookup_fail_fx_serial     2,707       3,845             1,138   42.04%   x 0.70 
 lookup_fx_highbits        69,104      3,861           -65,243  -94.41%  x 17.90 
 lookup_fx_random          3,387       4,114               727   21.46%   x 0.82 
 lookup_fx_serial          3,197       3,833               636   19.89%   x 0.83 
```

Amanieu:
Looking at the disassembly, the only difference is the hash function (as expected):

AHash:
```
movabs    $0xab5ad457fe0d4dbd,%r13
xor       %rsi,%r13
movabs    $0x5851f42d4c957f2d,%rax
imul      %rax,%r13
rol       $0x9,%r13
imul      %rax,%r13
```

FxHash:
```
movabs    $0x517cc1b727220a95,%r13
imul      %rsi,%r13
```

Amanieu:
On my system the AHash instruction sequence has a latency of 9 cycles (imul is 3 cycles, all other instructions 1 cycle) while FxHash only has a latency of 4 cycles.

@tkaitchuck Do you think it might be possible to get away with just a single multiply? Possibly using a trick from Absl Hash which is to perform a long `u64 * u64 -> u128` multiply and then xor-ing the two words of the result. [link](https://github.com/abseil/abseil-cpp/blob/master/absl/hash/internal/hash.h#L734)

tkaitchuck:
I could but that would be x86 speffic. There aren't a lot of processors with that instruction that lack the AES instruction. So I can add it as a third path, but I'm not sure it will see much use.

tkaitchuck:
Update: I tried changing the update function in the fallback to
`(self.buffer as u128).wrapping_mul(MULTIPLE as u128).convert();` and then two xors.
It actually made it slower in my benchmarks. I don't really understand why.

tkaitchuck:
One thing I could do, is remove the final multiply in the finalize block. It would mean that primitives who's input only differed in the highest bit would have hashes that differ by only one bit. But with the rotation I could still guarantee that the difference would be in the low order byte. It would undermine the quality of the hash, but it may not be a problem in practice because longer inputs such as strings would not be materially affected.

tkaitchuck:
@Amanieu 
I tweaked the method you suggested and pushed it to aHash 0.2.2, try it out and let me know how it performs. It now seems to work fairly well for integers. For longer strings I am keeping the previous algorithm.
I may want to put it in a feature gate in the future depending on how fast it is on other platforms.

PR #105, opened 2019-07-19
Add doc-comment to check README file code examples

GuillaumeGomez:
Considering that there is actually only one example, don't hesitate to close it if you find it not worth it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,7 +18,8 @@
         ptr_offset_from,
         test,
         core_intrinsics,
-        dropck_eyepatch
+        dropck_eyepatch,
+        cfg_doctest,
     )
 )]
 #![warn(missing_docs)]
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -35,6 +36,15 @@ extern crate alloc;
 #[cfg(not(has_extern_crate_alloc))]
 extern crate std as alloc;
 
+#[cfg_attr(feature = "nightly", macro_use)]
+#[cfg(feature = "nightly")]
+#[cfg(doctest)]
+extern crate doc_comment;
```
Amanieu: Isn't `macro_use` and `extern crate` mostly unnecessary now that we have 2018 edition? Can't we just do:
```rust
doc_comment::doctest!("../README.md");
```

GuillaumeGomez: We absolutely can. If you only target 2018 edition, then I'll remove those lines.

Amanieu: Yes this crate only targets 2018 edition, please remove those lines.

Amanieu:
CI fails because rustfmt is sad.

GuillaumeGomez:
rustfmt, my nemesis! (updated btw)

Amanieu:
Thanks!

PR #110, opened 2019-08-05
Check in CI that the crate builds on targets without libstd

### LINKED ISSUE #109: hashbrown has a hard time compiling against libcore now
```rust
error[E0599]: no method named `entry` found for type `hashbrown::map::HashMap<rusttype::GlyphId, M>` in the current scope
  --> livesplit-core\src\rendering\glyph_cache.rs:31:21
   |
31 |         self.glyphs.entry(glyph).or_insert_with(|| {
   |                     ^^^^^
   |
   = note: the method `entry` exists but the following trait bounds were not satisfied:
           `hashbrown::map::DefaultHashBuilder : core::hash::BuildHasher`
```

ed0b240192c3a5c5c20995c7f8188453ce4e3a2f still builds. So it's either the raw API or more likely AHash that broke it.

Amanieu:
Fixes #109

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,6 +62,10 @@ matrix:
     - name: "aarch64-unknown-linux-gnu"
       env: TARGET=aarch64-unknown-linux-gnu CROSS=1
 
+    # Ensure that we successfully build without libstd
+    - name: "thumbv6m-none-eabi"
+      env: TARGET=thumbv6m-none-eabi CROSS=1 NO_STD=1
+
 install: travis_retry rustup target add "${TARGET}"
 script: sh ci/run.sh
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ matrix:
       rust: stable
       env: TARGET=x86_64-unknown-linux-gnu
     - name: "x86_64-unknown-linux-gnu (Rust 1.31.0)"
-      rust: 1.31.0
+      rust: 1.32.0
       env: TARGET=x86_64-unknown-linux-gnu
     - name: "i686-unknown-linux-gnu"
       env: TARGET=i686-unknown-linux-gnu CROSS=1
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ matrix:
     - name: "x86_64-unknown-linux-gnu (stable)"
       rust: stable
       env: TARGET=x86_64-unknown-linux-gnu
-    - name: "x86_64-unknown-linux-gnu (Rust 1.31.0)"
+    - name: "x86_64-unknown-linux-gnu (Rust 1.32.0)"
       rust: 1.32.0
       env: TARGET=x86_64-unknown-linux-gnu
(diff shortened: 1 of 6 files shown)


Amanieu:
cc @tkaitchuck: This is blocked on `const_random` adding `#![no_std]`.

Amanieu:
@tkaitchuck It seems that public dependencies (`public = true`) are only supported on nightly.

Amanieu:
Hmm I'm not sure why this worked before though...

Amanieu:
Oh it's because `const-random` 0.1.3 didn't use `public = true`.

Amanieu:
@tkaitchuck Can you please remove `public = true` from the `const-random` crate's Cargo.toml? And check that AHash compiles on stable.

tkaitchuck:
@Amanieu Does this still cause a problem with const-random 0.1.5

tkaitchuck:
Done. Const_random 0.1.6.

Amanieu:
@tkaitchuck `no_panic` seems to be causing build failures on Windows, can you investigate?

Amanieu:
@tkaitchuck Also, `ahash` seems to be importing the `alloc` crate, which is not actually used: https://travis-ci.com/rust-lang/hashbrown/jobs/223101557

Amanieu:
The `no_panic` failure can be reproduced with: 
```
cargo test --target x86_64-pc-windows-gnu --no-run
```
Make sure you have the mingw linker properly set up in ~/.cargo/config

tkaitchuck:
The issue is this: https://github.com/rust-lang/rust/issues/33465
I have confirmed that the work around posted there fixes the issue. I don't have permission to re-open that issue. 
@Amanieu Is this something we can easily work around? If we can't update the config bors uses to do the right thing, or modify the command to pass the flags to do it for this build, I could disable no-panic so that it does not attempt to use a linker. Obviously that is not ideal.

tkaitchuck:
I saw two issues with running that command the first was: 
https://github.com/rust-lang/rust/issues/33465
Which appears to still be an unresolved issue.

But I assume based on your comment about the config that this is not what you are referring to. The second issue was due to the optimizer not being aggressive enough. I've fixed this is 0.2.10. Hopefully that resolves the issue

Amanieu:
@tkaitchuck Windows builds are still failing, I think you need to mark the `#[no_panic]` test with `#[cfg(test)]`.

tkaitchuck:
Ok, done. 
But I am confused, was it building it in the 'dev' profile?

Amanieu:
The builds seem to be fixed now. I'm just waiting on https://github.com/rust-lang/rust/pull/63302 to land on nightly for the thumbv6m fix.

By the way, `no-panic` should be a dev-dependency now so it doesn't get pulled in outside of tests.

Amanieu:
>  But I am confused, was it building it with the 'dev' profile?

The reason it was still failing is because the profile flags in Cargo.toml are only taken into account when building that crate directly. They are ignored when a crate is built as a dependency.

PR #111, opened 2019-08-07
README: Notes on comparison

RalfJung:
When hashers with different design goals are compared, that should be noted. Otherwise this compares apples with oranges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,9 +34,9 @@ table implementation in the Rust standard library.
 
 ## Performance
 
-Compared to the previous implementation of `std::collections::HashMap` (Rust 1.35).
+Compared to the previous HashDoS-resistant implementation of `std::collections::HashMap` (Rust 1.35).
 
-With the hashbrown default AHash hasher:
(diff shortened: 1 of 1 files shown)


CryZe:
Doesn't ahash provides HashDOS resistance too? They use various sources of randomness and pull that into their seed and they even say it's HashDOS resistent like SipHash in their README.

Amanieu:
I don't think it has been cryptographically proven to be secure though...

CryZe:
Yeah that's probably true

RalfJung:
Yeah, to my knowledge aHash has not been subject to any independent cryptographic analysis. Without such an analysis, I wouldn't trust any cryptographic claims.

It is undoubtedly much closer to HashDos resistant than FxHash, but that's not the same thing.

RalfJung:
Looks like this note was lost at some point... the README now again advertises hashbrown in a way that does not mention that switching from the standard HashMap to this, one will lose DoS resistance. (I consider any such claims about aHash to be highly speculative, so each user should decide for themselves if they consider this 'good enough'.)

PR #116, opened 2019-09-05
Group::static_empty() change const to static.

### LINKED ISSUE #115: Declare Group::static_empty()'s value as static.
Presently inside of [ss2](https://github.com/rust-lang/hashbrown/blob/83ce3d5ab782dd7808394d435e7977e68f88325c/src/raw/sse2.rs#L32)/[generic.rs](https://github.com/rust-lang/hashbrown/blob/83ce3d5ab782dd7808394d435e7977e68f88325c/src/raw/generic.rs#L61), the method Group::static_empty() returns the address of a `const` value (promoted to `'static`) instead of an explicitly `static` variable.  As a result, the value ends up being created inside of each consuming dylib instead of having a singular location inside of `std`'s dylib (or hashbrown's dylib).  

In my use case, this is preventing me from reloading a dylib while any empty (`HashMap::new()`) instances exist, as their `ctrl` pointer ends up pointing to garbage in the unloaded dylib, instead of pointing to a static variable inside of `std`.

If people are cool with changing `const` -> `static`, I'd be happy to prepare the trivial PR.

rickvanprim:
Fixes #115.

### DIFF
--- a/src/raw/generic.rs
+++ b/src/raw/generic.rs
@@ -63,7 +63,7 @@ impl Group {
             _align: Group,
             bytes: [u8; Group::WIDTH],
         };
-        const ALIGNED_BYTES: AlignedBytes = AlignedBytes {
+        static ALIGNED_BYTES: AlignedBytes = AlignedBytes {
             bytes: [EMPTY; Group::WIDTH],
         };
         unsafe { &ALIGNED_BYTES.bytes }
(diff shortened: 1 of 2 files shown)


Amanieu:
Can you add a small comment explaining why we use a `static` instead of a `const`?

Amanieu:
Thanks!

@bors r+

PR #118, opened 2019-09-30
Add RustcVacantEntry::insert_entry for rust-lang/rust#64656

passcod:
See rust-lang/rust#64656.

~~This was based on v0.5.0 as that's what rustc uses; I can rebase onto master, but I'm not sure whether rustc wants v0.6.0 or if v0.6.0 is rustc-ready.~~

For context, this ultimately provides an API with this shape:

```rust
impl Entry<'a, K, V> {
    fn insert(self, value: V) -> OccupiedEntry<'a, K, V> {…}
}
```

to be used when one wants to insert a value without consuming the Entry, e.g. because one wants to keep a reference to the key around. There's more at the original (defunct) PR: rust-lang/rust#60142.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1433,6 +1433,33 @@ where
 }
 
 impl<'a, K, V, S> RawEntryMut<'a, K, V, S> {
+    /// Sets the value of the entry, and returns a RawOccupiedEntryMut.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use hashbrown::HashMap;
+    ///
+    /// let mut map: HashMap<&str, u32> = HashMap::new();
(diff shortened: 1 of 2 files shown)


Amanieu:
Please rebase onto master. Also you'll want to run rustfmt.

Amanieu:
There are a few CI failures, but they seem unrelated to this PR.

passcod:
I can implement the `insert` method for the other public Entry APIs in this crate if wanted?

passcod:
Do I need to do anything for this to go further?

Amanieu:
I'm just waiting for #120 to land before merging this.

passcod:
Ah, thanks!

PR #121, opened 2019-10-12
Allow clippy::doc_markdown

Amanieu:


Amanieu:
Blocked on https://github.com/rust-lang/rust/issues/65335

PR #123, opened 2019-10-15
Remove BuildHasher requirement from raw entry APIs.

goffrie:
This makes it possible to use HashMap with `S=()`, which forces the user to always supply hashes manually. This is an alternative to making the key type unhashable.

I'm not sure if this change is desirable but I'm putting it up for discussion anyway :)

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -958,10 +958,7 @@ where
     }
 }
 
-impl<K, V, S> HashMap<K, V, S>
-where
-    S: BuildHasher,
-{
+impl<K, V, S> HashMap<K, V, S> {
     /// Creates a raw entry builder for the HashMap.
     ///
     /// Raw entries provide the lowest level of control for searching and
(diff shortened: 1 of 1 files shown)


Amanieu:
Sorry about the late response, I somehow missed this PR.

Could you explain your use case for this functionality? Why can't you just use the default hasher? If your key is not `Hash` then you'll need to provide hashes manually anyways through `RawEntry`.

goffrie:
So, my use case is that I have a `HashMap<usize, V>` where `usize` is indexing into some other table and fetching the actual keys from that table. As such, I'm always using raw entry APIs to provide the hash/comparison function myself. This works fine, but it means that it's possible for me to accidentally use an API that hashes my `usize` and create wrong behaviour that way. With this patch, I can instead make my map `HashMap<usize, V, ()>`, and it becomes statically impossible for the hashmap to do that.

It's also possible for me to do this today using a wrapper type `HashMap<NonHashable<usize>, V>` where `NonHashable` doesn't implement Hash, but that seems more cumbersome to me.

If you have other suggestions I'm open to them too :)

Amanieu:
That actually sounds pretty reasonable.

@bors r+

goffrie:
Thanks!

Amanieu:
Could you fix the compile errors? I know they're not your fault, but they should be easy to fix.

goffrie:
ah, I missed that. Thanks!

PR #125, opened 2019-10-30
Introduce `ahash-compile-time-rng` feature.

### LINKED ISSUE #124: [Request] Update documentation to mention potential `no_std` problem.
**Disclaimer: This is actually not an issue with `hashbrown` but with `cargo`.**

---

**Topic: no_std support**

While `hashbrown` itself is able to compile successfully for `no_std` targets, and even uses the CI pipeline to verify its compatibility to the `thumbv6m-none-eabi` target, it might still trigger issues if used with other dependencies when the default `ahash` feature is enabled.

**Issue**

The dependency tree of `ahash` injects the `getrandom` crate if the default feature set is enabled. This crate is heavily platform dependent and does not support targets like the `thumbv6` or `thumbv7` families.

**How does this happen?**

The underlying issue is an a long standing and well known cargo bug (https://github.com/rust-lang/cargo/issues/5760) which automatically calculates a required feature set for each (sub-)dependency inside the dependency tree regardless of the dependency origin.

If `ahash` - with default feature set - is used, it will include the `const-random` crate and therefore the proc macro crate `const-random-macro` into the dependency tree. This will request the `getrandom` feature for the crates `rand v^0.7` and indirectly also `rand_core v^0.5.1`.

As dependencies of proc macros are compiled for the host and not the target architecture this works fine as long as the host architecture is contained [in the following list](https://github.com/rust-random/getrandom/blob/d661aa7e1b8cc80b47dabe3d2135b3b47d2858af/src/lib.rs#L196-L262) and the `no_std` build should be successful (as you can see inside the CI).

But it is going to fail as soon as one of the following two conditions are true:

 - The host architecture is not supported.

 - The target architecture is not supported and the dependency tree contains at least one non proc macro dependency, which requires either `rand v^0.7` or `rand_core v^0.5.1`, because than the underlying `cargo` issue kicks in. Cargo is going to inject the `getrandom` feature requirement for these crates, originating from the dependency list of `const-random-macro`, into the other dependency so that cargo is trying to compile `getrandom` not only for the host architecture but also for the target architecture.

**Request**

As long as https://github.com/rust-lang/cargo/issues/5760 is still a thing it would be great if at least the `hashbrown` documentation could get a warning paragraph explaining this topic including how to work around this issue.

Potential workarounds I can see:

 - `hashbrown` could introduce a way to disable the `compile-time-rng` feature of `ahash` if default features are disabled.

 - The relying crate could disable the `ahash` feature of `hashbrown` and reintroduce another hasher or `ahash` without `compile-time-rng` feature.

Both might result in `ahash` losing the compile time random seed for the hash values.

0ndorio:
**Content**

Disables the default features of `ahash` and reintroduces them
through a new feature called `ahash-compile-time-rng`, which is
enabled by default.

The new feature makes it possible for depended crates to rely on
`hashbrown` with `ahash` as default hasher and to disable the
`compile-time-rng` at the same time.

This might be useful for any depended crate targeting `no_std`,
which contains `rand` or `rand_core` somewhere inside the dependency
tree as a bug in cargo accidentally enables the underlying `getrandom`
feature if `compile-time-rng` is enabled [1].

... fixes #124

[1] https://github.com/rust-lang/cargo/issues/5760

---

**Warnings**

 (1) Compiling `ahash` with disabled `compile-time-rng` feature is currently broken and requires https://github.com/tkaitchuck/aHash/pull/25 to be merged in advance.

 (2) This introduces a hidden behavior change for all dependent crates, using hashbrown with `default-features = false` and `features = 'ahash'`. This happens as the `ahash` feature no longer implicitly enables the `compile-time-rng` feature of `ahash`.

---

Is the naming of the feature okay?  Do I need to add any additional  changes?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ build = "build.rs"
 
 [dependencies]
 # For the default hasher
-ahash = { version = "0.2.11", optional = true }
+ahash = { version = "0.2.11", optional = true, default-features = false }
 
 # For external trait impls
 rayon = { version = "1.0", optional = true }
@@ -37,7 +37,13 @@ serde_test = "1.0"
 doc-comment = "0.3.1"
 
 [features]
-default = ["ahash", "inline-more"]
+default = [
+    "ahash",
+    "ahash-compile-time-rng",
+    "inline-more",
+]
+
+ahash-compile-time-rng = [ "ahash/compile-time-rng" ]
 nightly = []
 rustc-internal-api = []
 rustc-dep-of-std = ["nightly", "core", "compiler_builtins", "alloc", "rustc-internal-api"]
--- a/README.md
+++ b/README.md
@@ -104,6 +104,10 @@ This crate has the following Cargo features:
 - `raw`: Enables access to the experimental and unsafe `RawTable` API.
 - `inline-more`: Adds inline hints to most functions, improving run-time performance at the cost
   of compilation time. (enabled by default)
+- `ahash`: Compiles with ahash as default hasher. (enabled by default)
+- `ahash-compile-time-rng`: Activates the `compile-time-rng` feature of ahash, to increase the
+   DOS-resistance, but can result in issues for `no_std` builds. More details in
+   [issue#124](https://github.com/rust-lang/hashbrown/issues/124). (enabled by default)
 
 ## License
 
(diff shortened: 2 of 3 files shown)


Amanieu:
Clippy seems to be failing due to a new lint. Could you add `clippy::must_use_candidate` to the whitelist in `lib.rs`?

Also, could you document the feature in `README.md` (there's a feature list near the bottom).

Amanieu:
You need to run `cargo fmt` to fix CI.

PR #138, opened 2020-01-10
Upgrade to autocfg 1

cuviper:


cuviper:
>     The command "rustup target add i686-apple-darwin" failed.

That's one of the [demoted targets](https://blog.rust-lang.org/2020/01/03/reducing-support-for-32-bit-apple-targets.html), no longer providing toolchains.

PR #147, opened 2020-03-17
Future-proof specialization code

Amanieu:
As per @Centril's [comment](https://github.com/rust-lang/rust/pull/70052#issuecomment-600092343)


Centril:
Syntax-related bits look good :)

PR #150, opened 2020-04-05
micro optimize repeat function

andjo403:
makes it easier for the compiler and the reader with less code :)

for "prof" https://godbolt.org/z/5bdD2b
when I tried cargo bench I got very large variance between runs was multiple test that one time had 2x improvement to the next time have a minor regression so have more trust in the llvm-mca result.


Amanieu:
Impressive results, thanks!

@bors r+

PR #157, opened 2020-04-29
Mark RawTable::par_iter unsafe

### LINKED ISSUE #156: Should RawTable::{into_iter, par_iter, into_par_iter, par_drain} be marked unsafe?
I think the current implementations of `RawTable::into_iter`, `RawTable::par_iter`, `RawTable::into_par_iter`, and `RawTable::par_drain` need to be marked `unsafe` for the same reason that `RawTable::iter` and `RawTable::drain` already are.

andersk:
It could be misused in the same way as `RawTable::iter`.

Fixes #156.


Amanieu:
Actually `drain` and `into_iter` are both safe since drain has a lifetime bound on the table and `into_iter` owns the table. Only plain `iter` is unsafe.

andersk:
Yeah okay, I buy that. Updated.

PR #167, opened 2020-07-01
impl Clone for RawIntoIter, and add table access methods

jonhoo:
This adds the ability to find and erase elements in a `RawTable` while
it is being consumed using `RawIntoIter`. This may be useful in contexts
where a `RawTable` is being destructed bit-by-bit, interleaved by user
operations.

Specifically, this is (I think) what I'd need for
https://github.com/jonhoo/griddle/issues/1 and
https://github.com/jonhoo/griddle/issues/2.

Please do double-check my implementation here, as I am not
completely sure I got the corner cases right.

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -1463,32 +1463,52 @@ impl<T: Clone> Clone for RawIntoIter<T> {
         } else {
             unsafe {
                 // First, we construct an empty table with the same layout as the current one.
-                let mut table =
+                let mut table = ManuallyDrop::new(
                     RawTable::new_uninitialized(self.table.buckets(), Fallibility::Infallible)
-                        .unwrap_or_else(|_| hint::unreachable_unchecked());
-                table.ctrl(0).write_bytes(EMPTY, table.num_ctrl_bytes());
-                // Then, we move over all the _remaining_ non-empty buckets.
+                        .unwrap_or_else(|_| hint::unreachable_unchecked()),
+                );
+
+                // And copy over all the control state. This is okay despite the values are holding
+                // invalid values since the table is wrapped in ManuallyDrop. We need to do this
+                // instead of filling with EMPTY, since the ctrl values of past buckets may affect
+                // lookups.
+                self.table
+                    .ctrl(0)
+                    .copy_to_nonoverlapping(table.ctrl(0), table.num_ctrl_bytes());
+
+                // Next, we move over all the _remaining_ non-empty buckets.
                 // We can do so without knowing the hash since the tables are identical.
                 for bucket_in_self in self.iter.clone() {
                     let index = self.table.bucket_index(&bucket_in_self);
-                    let old_ctrl = *table.ctrl(index);
                     let bucket_in_new = table.bucket(index);
-                    table.growth_left -= special_is_empty(old_ctrl) as usize;
-                    table.set_ctrl(index, *self.table.ctrl(index));
                     bucket_in_new.write(bucket_in_self.as_ref().clone());
-                    table.items += 1;
                 }
-                // We can now get a RawIntoIter for the new table, but it will have to re-iterate
-                // over all the empty buckets this RawIntoIter has already drained. To avoid that
-                // cost, we migrate over the state from the current iterator. Much of its state is
+                table.growth_left -= self.table.growth_left;
+                table.items = self.table.items;
+
+                // We can now get a RawIntoIter for the new table.
+                //
+                // However:
+                //
+                //  - We cannot start it from the beginning, since we don't have valid values for
+                //    anything prior to self.iter, even though the ctrl bits say otherwise.
+                //  - Even if it were safe, we don't _want_ it to re-iterate over all the empty
+                //    buckets this RawIntoIter has already drained.
+                //
+                // So, we migrate over the state from the current iterator. Much of its state is
                 // still valid, since the new table has the same setup as the old one.
-                let mut iter = table.into_iter();
+                //
+                // We wrap the RawIntoIter in ManuallyDrop until its state has been "fixed up", so
+                // that it doesn't get dropped due to a panic and start dropping invalid values.
+                // Note that it is _vital_ here that RawTable::into_iter does not panic.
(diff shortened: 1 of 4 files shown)


src/raw/mod.rs
```diff
@@ -1469,7 +1530,7 @@ impl<T> Drop for RawIntoIter<T> {
             }
 
             // Free the table
-            if let Some((ptr, layout)) = self.alloc {
+            if let Some((ptr, layout)) = ManuallyDrop::take(&mut self.table).into_alloc() {
```
jonhoo, line 1533: This isn't great, as `ManuallyDrop::take` only landed in Rust 1.42. I'm not quite sure how to work around this, except by making it an `Option`?

jonhoo:
I also wonder if it should be possible to `Clone` the remaining elements out of the `RawIntoIter`. Not entirely sure how to accomplish that though.

jonhoo:
Actually, never mind `Clone`. It can be done using `into_iter.iter().cloned().collect()`.

jonhoo:
I added a `Clone` impl, though it definitely needs to be looked over.

I think one remaining challenge is how to construct a `rayon` parallel iterator from a `RawIntoIter` (since we don't have a `RawTable` to give it). That can probably be punted for now though.

jonhoo:
A fun side-effect here is that this enables `hash_map::IntoIter` to be `Clone` if we wish it to be :)

jonhoo:
The test failures seem to be failure to pull some Docker image due to rate limiting?

jonhoo:
I think this is ready for review now. You can see how one might take advantage of this in https://github.com/jonhoo/griddle/pull/8. I'll add that I've done pretty extensive testing of this feature now, as it's gone through the ringer with griddle's quickcheck + its adapted indexmap tests.

Amanieu:
I can't help but feel that this is the wrong approach to take. I think that rather than adding more features to `RawIntoIter`, you should instead be using `RawIter` directly.

jonhoo:
That was my original instinct too, but there are so many invariants to uphold that it didn't feel feasible to present a more general API without exposing a lot of internal state. I'll do some more thinking tomorrow to see if I can come up with something better though.

I think there are two main "parts" to this:

First is to give users of the raw APIs the ability to mutate a `RawTable` while iterating over it. The user can always re-start the iterator, but if the way they mutate is, say, consume every element the iterator yields, then re-starting is _very_ expensive; it needs to search for the first non-empty bucket, which is closer and closer to the end. Unfortunately, implementing a general-purpose `RawIter::refresh` efficiently is difficult for two reasons:

 1. You don't know whether `items` should change — it should change only if there's an insert/remove after the position of the iterator, but not before.
 2. When reloading the current group, you can't distinguish between an insertion in the same group _prior_ to the current iteration point (which should not be yielded by the iterator), and an insertion in the same group _after_ the current iteration point (which should be yielded). This is because the iterator (currently) doesn't keep any state beyond the cached bitmask, and it only knows that the leftmost non-yielded 1 is the _next_ item it was _going_ to yield.

Thinking about it some more, I think maybe these could both be solved by having such a hypothetical `RawIter::refresh` take a `&Bucket` representing "the change to update the iterator to reflect". Using that, the iterator can check how that bucket changed by checking its current state (and assume it was toggled), and it can in turn use that to patch the current group state appropriately.

The second challenge is to find a way to enable users to efficiently "resume" an iterator after a `RawTable` is cloned. This requires rewiring a bunch of internal iterator state, and is what most of the code in the `RawIntoIter::clone` impl is doing. I suppose we could instead expose that through a `RawIter::resume_in(&mut self, &RawTable)`, though it would be _super_ unsafe (but maybe that's okay).

I think with those two added methods, I might be able to do what I need. It won't be quite as efficient as what's in this PR at the moment, but it might be workable. So, how would you feel about a PR that added the following:

```rust
impl RawIter {
  /// Refresh the iterator such that an insert or removal for the given bucket is reflected.
  ///
  /// If the bucket was inserted into, the iterator _may_ yield the given bucket next
  /// even if it already yielded a bucket with a lower index.
  ///
  /// This method must be called after one such modification is made.
  /// Calling this method after inserting *and* removing from the bucket,
  //// in either order, may corrupt the iterator state.
  pub unsafe fn refresh_reflect(&mut self, b: &Bucket);

  /// Resume this iterator in the same location in another table.
  ///
  /// The given table must have the same layout as the iterator's previous table.
  pub unsafe fn resume_in(&mut self, table: &RawTable);
}
```

I'll try to hack that together tomorrow and see if that's sufficient for what I need in griddle.

Amanieu:
I'm curious about the reasons why you would need this functionality.

- Why are you inserting elements into the table while iterating through it?
- Why do you need to resume an iterator after a table is cloned? In griddle you could just "flush" the second table on clone and reset the iterator.

jonhoo:
> Why are you inserting elements into the table while iterating through it?

I'm not, but I _am_ removing from the table while iterating over it. This is why I only implemented erase/remove for `RawIntoIter` in this PR — it's much simpler to handle just removals. The reason I have to be able to do removals is in case someone removes a key from the map which has not yet been moved from the old table to the new one. It needs to be removed from the old table, and the iterator I'm keeping over that old table also needs to be updated to reflect that removal.

If we _only_ want to support refreshing for removals, then we could _just_ have a `RawIter::refresh_removal(&Bucket)`. It would still need to take a `&Bucket` though to determine whether the `items` count should change.

> Why do you need to resume an iterator after a table is cloned? In griddle you could just "flush" the second table on clone and reset the iterator.

I suppose that's true. My thinking here was that a `Clone` should not require that you _also_ finish the resize, but maybe you're right that that's silly. If someone clones the map, then they'll have to walk the map to clone its elements _anyway_ (well, modulo `Copy`), so perhaps collapsing the second table into the first isn't such a bad incremental cost.

One unfortunate aspect of this is that `Clone` takes `&self`, so I would only be able to do the collapse in the newly created map, but not in the original map. So the "flush" still has to happen twice. But that's no worse than what happens with this PR.

jonhoo:
Closing in favor of #175.

Issue #168, opened 2020-07-01
Document erase_no_drop guarantees


jonhoo:
Following some discussion in https://github.com/bluss/indexmap/pull/131#discussion_r447965041, it'd be good to document the contract of `erase_no_drop`. In particular, there is code both inside `hashbrown` and in the pending PR to `indexmap` (https://github.com/bluss/indexmap/pull/131) along the lines of:
```rust
let v = unsafe {
    table.erase_no_drop(&bucket);
    bucket.read()
}
```
This, by first reading, seems like it shouldn't be okay. `erase_no_drop` erases the bucket, so why is it okay to read from it after the fact? It [turns out](https://github.com/bluss/indexmap/pull/131#discussion_r447980041), probably unsurprisingly, that `hashbrown` indeed currently supports this sequence of operations; `erase` only erases the bucket in the control bytes, but does not otherwise invalidate the bucket.

I think we have a choice to make: either we add this as a guarantee to the documentation of `erase_no_drop` so that callers can rely on that detail of the implementation, _or_ we do not give that guarantee, and require callers to always access the bucket _before_ they erase it. The latter gives us more leeway to change `erase_no_drop` later, at the cost of slightly less convenient access for callers. Of course _internally_ it's fine for `hashbrown` to rely on being able to access the bucket after `erase_no_drop` is called. This is a question of what the public contract (which we cannot break without a breaking release) should be.

cuviper:
Do we need a similar clarification on `clear_no_drop`?

In theory, you could have a outstanding `Bucket` live across this and still want to read it afterward, but I don't see anything that's actually doing that right now.

jonhoo:
Yes, I think the same clarification is needed there

Amanieu:
`erase_no_drop` is deprecated and replaced by `erase` and `remove` so I'm going to close this.

PR #171, opened 2020-07-01
Add RawTable::erase and remove

cuviper:
The two main uses of `erase_no_drop` are to then drop it in place or
read it out, which are now provided by `erase` and `remove`.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -578,9 +578,7 @@ impl<K, V, S> HashMap<K, V, S> {
             for item in self.table.iter() {
                 let &mut (ref key, ref mut value) = item.as_mut();
                 if !f(key, value) {
-                    // Erase the element from the table first since drop might panic.
-                    self.table.erase_no_drop(&item);
-                    item.drop();
+                    self.table.erase(item);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


cuviper:
I think we could now make `erase_no_drop` private, resolving its API ambiguity (cc #168 @jonhoo), but maybe that's still wanted for more pointer-like uses of `Bucket`.

jonhoo:
I love this. At least for my use-case I don't need access to the bucket from `erase_no_drop`.

cuviper:
Hmm, clippy is complaining "error: this argument is passed by value, but not consumed in the function body", but I took values on purpose as a semantic signal. Should I `drop` in the body, or just hide that lint?

jonhoo:
I like the semantic signal. I'd explicitly allow that lint on these functions.

jonhoo:
Should we also mark `erase_no_drop` as `deprecated` with this change?

Out of curiosity, what is the protocol for releasing a change like this to `hashbrown`? I guess since this is backwards-compatible, we could just cut 0.8.1? Or is there a more formal process given hashbrown's association with `std`?

cuviper:
> Or is there a more formal process given hashbrown's association with `std`?

I don't think so -- it's still a manual PR to update rust's `Cargo.lock`, even when it would be semver compatible. There's an open PR updating to 0.8.0 here: https://github.com/rust-lang/rust/pull/70052

jonhoo:
@Amanieu How do you feel about marking `erase_no_drop` as deprecated?
And what process should I follow to move towards a 0.8.1 with this, #174, and #175?

Amanieu:
I think deprecating `erase_no_drop` is reasonable.

Regarding 0.8.1 I would like to make a few additional changes on my end before releasing. I expect to release at some point this week.

jonhoo:
Filed https://github.com/rust-lang/hashbrown/pull/176.

PR #173, opened 2020-07-02
Make PartialEq for map and set more flexible

### LINKED ISSUE #172: Cannot compare maps with different BuildHashers
The current `PartialEq` implementation only permits comparing against the exact same type, which includes `S`:
https://github.com/rust-lang/hashbrown/blob/ace3ac2a63b12cc60a77b4a96e34f586816e3931/src/map.rs#L1102-L1108

If we wanted to allow `==` between maps that share `K` and `V`, but not `S`, we can do so by adapting the signature just slightly to:
```rust
impl<K, V, S, S2> PartialEq<HashMap<K, V, S2>> for HashMap<K, V, S>
where
    K: Eq + Hash,
    V: PartialEq,
    S: BuildHasher,
    S2: BuildHasher,
{
    fn eq(&self, other: &HashMap<K, V, S2>) -> bool {
```

Is there a compelling reason not to modify the implementation in this way?

jonhoo:
Fixes #172

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1099,13 +1099,14 @@ impl<K, V, S> HashMap<K, V, S> {
     }
 }
 
-impl<K, V, S> PartialEq for HashMap<K, V, S>
+impl<K, V, S, S2> PartialEq<HashMap<K, V, S2>> for HashMap<K, V, S>
 where
     K: Eq + Hash,
     V: PartialEq,
     S: BuildHasher,
+    S2: BuildHasher,
 {
-    fn eq(&self, other: &Self) -> bool {
+    fn eq(&self, other: &HashMap<K, V, S2>) -> bool {
         if self.len() != other.len() {
             return false;
         }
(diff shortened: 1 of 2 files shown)


jonhoo:
The test failures fixed by https://github.com/rust-lang/hashbrown/pull/173/commits/360d6a29c0378f74f9786fa6baeaf306ee1ad836 are a little awkward. I guess that makes this a backwards-incompatible change?

Amanieu:
I am concerned about the type inference implications. I feel that it's probably not worth it for such a niche feature. Do you have a specific use case for this?

jonhoo:
No, I don't, I just added it because it seemed like a straightforward enough capability to add. It mostly comes in handy in tests I think. But I agree with you that the inference implications are unfortunate. I'm okay with closing this if you'd prefer.

Amanieu:
I'd rather close this. Generally you're going to use the same hasher for all your hashmaps anyways.

PR #175, opened 2020-07-03
Improve RawIter re-usability

jonhoo:
Creating a new iterator for a table is occasionally expensive,
especially if a suffix of the table (since we iterate "backwards") is
sparsely populated. This PR adds methods that allow re-using an existing
`RawIter` in more situations.

Replaces #167.

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -1015,6 +1015,24 @@ impl<T> RawTable<T> {
         }
     }
 
+    /// Returns an iterator which consumes all elements from the table.
+    ///
+    /// Iteration starts at the provided iterator's current location.
+    ///
+    /// This method panics if the given iterator does not cover all items remaining in the table.
+    pub unsafe fn into_iter_from(self, iter: RawIter<T>) -> Option<RawIntoIter<T>> {
+        if iter.len() != self.len() {
+            return None;
+        }
+
+        let alloc = self.into_alloc();
+        Some(RawIntoIter {
+            iter,
+            alloc,
+            marker: PhantomData,
+        })
+    }
+
     /// Converts the table into a raw allocation. The contents of the table
     /// should be dropped using a `RawIter` before freeing the allocation.
     #[cfg_attr(feature = "inline-more", inline)]
@@ -1250,12 +1268,8 @@ impl<T> IntoIterator for RawTable<T> {
     fn into_iter(self) -> RawIntoIter<T> {
         unsafe {
             let iter = self.iter();
-            let alloc = self.into_alloc();
-            RawIntoIter {
-                iter,
-                alloc,
-                marker: PhantomData,
-            }
+            self.into_iter_from(iter)
+                .unwrap_or_else(|| hint::unreachable_unchecked())
         }
     }
 }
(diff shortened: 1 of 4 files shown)


src/raw/mod.rs
```diff
@@ -1015,6 +1015,24 @@ impl<T> RawTable<T> {
         }
     }
 
+    /// Returns an iterator which consumes all elements from the table.
+    ///
+    /// Iteration starts at the provided iterator's current location.
+    ///
+    /// This method panics if the given iterator does not cover all items remaining in the table.
```
Amanieu, line 1022: Use a `debug_assert` instead. This method is unsafe anyways.

src/raw/mod.rs
```diff
@@ -1408,6 +1422,126 @@ pub struct RawIter<T> {
...
+                        //
+                        // The item count must already be correct, since a removal or insert
+                        // "prior" to the iterator's position wouldn't affect the item count.
+                    }
+                    Ordering::Equal => {
+                        // The toggled bucket was coming up next, so we just skip it.
+                        self.iter.current_group = self.iter.current_group.remove_lowest_bit();
+                        // It must be a removal, since its bit was set.
+                        self.items -= 1;
+                        debug_assert!(!is_insert);
```
Amanieu, line 1510: It seems that this is just a simplified version of the case below? Why not just merge it with that case so we only need a single `<=` check.

Amanieu:
I like this much better!

jonhoo:
Glad to hear it! I have now also implemented the necessary changes in griddle, and it seems to be sufficient: https://github.com/jonhoo/griddle/pull/12. The `Clone` implementation isn't as pretty as I'd like it to be, but it's not _too_ bad.

The griddle test suite also passes, including a number of additional iterations of the quickcheck tests I've run locally, so I think this is ready now :tada:

jonhoo:
One _really_ awkward side-effect of this (and on https://github.com/jonhoo/griddle/pull/12 in particular) is that griddle now cannot implement `Clone` without `K: Hash, S: BuildHasher` bounds, since it must perform inserts. This, in turn, means that `derive(Clone)` will no longer work :cry:

Amanieu:
Clone probably should depend on `K: Hash, S: BuildHasher`. The only reason hashbrown doesn't is because it needs to be backward compatible with the libstd `HashMap` which doesn't have those bounds on its `Clone` impl.

jonhoo:
Is there anything more you'd like to see added to this PR? I've tried to document as I went, and have also added the appropriate "resume from `RawIter`" methods for `drain` and `into_iter`.

src/raw/mod.rs
Amanieu: This comment is incorrect now that the implementation has changed.

Maybe a separate parameter should be added to pass the number of items remaining since it may not necessarily match the item count in the map.

jonhoo: I'm not sure I follow --- I think the comment is _now_ correct whereas it wasn't before? Previously this returned an `Option`, which I removed in favor of just the `debug_assert`. Or do you mean that this isn't true in release mode? I could make it a regular `assert!` if you'd prefer?

I don't think passing in the number of items makes sense here — if the number of items left in the iterator doesn't match the number of items left in the table, then this is pretty much guaranteed to do weird things (most likely leak memory).

Amanieu: You should just update the comment to say that making sure the number of items matches is a pre-condition of calling the function. The current comment implies that the function will check it for you, which is false (in release mode).

jonhoo: How's that?

Amanieu:
Thanks!

PR #179, opened 2020-07-10
Add HashSet::drain_filter method

### LINKED ISSUE #178: drain_filter for HashSet?
I'd like to have drain_filter for sets as well as maps...

mbrubeck:
Fixes #178.


Amanieu:
Could you add inline annotation to all functions: `#[cfg_attr(feature = "inline-more", inline)]`

mbrubeck:
Rebased and added conditional inline attributes.

PR #187, opened 2020-08-11
Reverse `drain_filter` behavior to match libstd

### LINKED ISSUE #186: `drain_filter` has opposite behavior of standard library version
In methods like [`std::vec::Vec::drain_filter`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain_filter), the element is removed if the closure returns `true`.  But in `hashbrown::HashMap::drain_filter` (#135) and `hashbrown::HashSet::drain_filter` (#179), the element is removed if the closure returns `false`.  😦 

cc rust-lang/rust#59618

mbrubeck:
Fixes #186.  This is the worst sort of silent breaking change, but we don't have much choice here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+- `drain_filter` now removes and yields items that do match the predicate,
+  rather than items that don't.  This is a **breaking change** to match the
+  behavior of the `drain_filter` methods in `std`.
+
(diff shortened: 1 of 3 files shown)


mbrubeck:
Sorry, I forgot to update the docs to match the new behavior.  Pushed a follow-up commit.

mbrubeck:
Updated the tests.

mbrubeck:
Fixed the doc example to compile in Rust 1.32.  Sorry for the churn!

Amanieu:
No problem!

@bors r+

Issue #189, opened 2020-08-19
Removing OccupiedEntry and get back a VacantEntry


SkiFire13:
Right now there's no way to remove a value from an `OccupiedEntry` getting back both the value removed and a `VacantEntry` for the key just removed. I propose to add something like

```rust
fn take(self) -> (V, VacantEntry<'a, K, V, S>)
```

to `OccupiedEntry`. However I'm not sure how this could be implemented without changing how `VacantEntry` works.

The use I have for this would be temporary removing a value from the map without needing 2 lookups. Right now I have to `remove` (1st lookup) the value, doing what I want to do with that and then `insert` (2nd lookup) it again. If I could reuse the fact that the `OccupiedEntry` had a valid bucket for that key I could skip a lookup in the map.

Amanieu:
Have you tried using a crate like [`take_mut`](https://crates.io/crates/take_mut)?

SkiFire13:
I already knew about take_mut but it doesn't work in my case because:

- I need to provide a fallback value (and I can't) or in case of panics it will abort. With my proposal the map should be left in a consistent state so no `catch_unwind` or `abort` is needed.
- In my use case I may or may not use the `VacantEntry` depending on the result of a computation involving the value I previously took, but `take_mut` forces me to put back a value.

To be fair [a workaround](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ad76e9f45f5fbf98db91ea5d24cea45c) inspired by `take_mut` should be possible but it requires you to write `unsafe` code and still need `catch_unwind` because the map isn't left in a truly valid state. I'm also not sure that this code is correct (in particular possible shenanigans that may happen with `catch_unwind`), so I would prefer not having to write it myself.

PR #190, opened 2020-08-23
Implement `replace_entry_with`

SkiFire13:
Adds `OccupiedEntry::replace_entry_with` that consumes `self`, takes an `FnOnce(&K, V) -> Option<V>` and replaces/removes the value based on the value returned from the closure.

This function is intended to be used when just having a mutable reference is not enough and it is more efficient than just using `remove` optionally followed by `insert`. This is achieved by saving the index and the ctrl byte before removing the element from the bucket, allowing to skip the rehash and bucket search in case the element is reinserted after the function returns.

All of this relies on a new function `RawTable::replace_bucket_with` that does all exactly the same thing (replaces/removes the value based on the value returned from the closure) but at the `RawTable` level.

There's also a `Entry::and_replace_entry_with` (that does nothing when the entry is a  `VacantEntry`) and the analogous functions for `RawEntryMut` and `RawOccupiedEntryMut`.

To be settled: 

- [x] ~~Missing tests for `RawEntryMut::and_replace_entry_with` and `RawOccupiedEntryMut::replace_entry_with`~~

- [ ] I'm not sure if I should add this to the Rustc version of the entry API (I guess it should only mirror the stdlib version of HashMap, so no?).

- [x] ~~It's possible for these functions to return an `Entry`/`RawEntryMut`, but this requires to add an extra field to `OccupiedEntry`/`RawOccupiedEntryMut`. I've implemented it [in another branch](https://github.com/SkiFire13/hashbrown/tree/replace-entry-with-return), should I merge it?~~

Also see #189

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+- `drain_filter` now removes and yields items that do match the predicate,
+  rather than items that don't.  This is a **breaking change** to match the
+  behavior of the `drain_filter` methods in `std`.
+
 ## [v0.8.2] - 2020-08-08
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.8.2] - 2020-08-08
+
+### Changed
+- Avoid closures to improve compile times. (#183)
+- Do not iterate to drop if empty. (#182)
+
 ## [v0.8.1] - 2020-07-16
 
 ### Added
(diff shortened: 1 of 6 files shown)


SkiFire13:
Checks should have failed only because of #191

Amanieu:
I like the version that returns an `Entry`. It's fine to make a breaking change here, we can just do a semver bump.

SkiFire13:
Just adding the fields should be fine, they aren't public, but adding a generic type on `RawOccupiedEntryMut` is the breaking change.

Amanieu:
That's fine since `raw_entry` is still unstable in libstd. So we should be able to make the same changes to libstd when the hashbrown dependency is updated.

@bors r+

Amanieu:
@bors r-

You need to update the minimum Rust version in the CI config. It seems that crossbeam has bumped their MSRV.

SkiFire13:
> You need to update the minimum Rust version in the CI config. It seems that crossbeam has bumped their MSRV.

I feel like it should be a separate PR. I'll open one when I get back home.

PR #202, opened 2020-09-12
Add more safe methods to RawTable

cuviper:
These methods combine lookup and bucket operations so the caller doesn't
have to deal with unsafe bucket methods.

- `get`: `find` and `as_ref`
- `get_mut`: `find` and `as_mut`
- `insert_entry`: `insert` and `as_mut`
- `remove_entry`: `find` and `remove`
- `erase_entry`: `find` and `erase`

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -810,8 +810,8 @@ where
         Q: Hash + Eq,
     {
         // Avoid `Option::map` because it bloats LLVM IR.
-        match self.get_key_value(k) {
-            Some((_, v)) => Some(v),
+        match self.get_inner(k) {
+            Some(&(_, ref v)) => Some(v),
             None => None,
         }
     }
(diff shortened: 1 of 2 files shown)


Amanieu:
What is the rationale for adding these methods instead of just using the safe API in `HashMap` directly?

I have been considering going in the other direction: exposing more functionality in `HashMap` so that we can eventually deprecate `RawTable`.

cuviper:
I was motivated for this by the `cached` crate's `SizedCache`, which was using `HashMap<K, usize>` and `LRUList<(K, V)>`, where the latter is indexable by the `usize` values from the map. I thought it would be nice to remove the `K` duplication by using `RawTable<usize>` instead of the map, reading the keys from the list -- https://github.com/jaemk/cached/pull/59. However, dealing with `Bucket` would introduce the first `unsafe` in that crate, avoided by this PR's safe methods.

That proposed use case is quite a bit like `indexmap`, where key storage is managed independently. Another example I've been playing with is this [`autohash`](https://github.com/cuviper/autohash) crate, where keys can report their own "hash" without actually going through a `Hasher`, as a better solution than https://github.com/rust-lang/rust/pull/76233 when the key is already hash-like. So I think `RawTable` has really nice flexibility, and I'm skeptical that the same uses could be served by `HashMap` extensions.

cuviper:
If anything, I might like to see `RawTable` pulled out into its own crate, for cases where the map/set implementations are not used. That could be forked out of `hashbrown`, but it would be nice to continue sharing that implementation.

cuviper:
@Amanieu -- any more thoughts on this? We can close if you think it's not viable, but it seems simple enough to me.

Amanieu:
Sorry, this slipped my mind. I think these are probably fine to add as is.

cuviper:
And I see you also published 0.9.1, thanks!

PR #203, opened 2020-09-13
Add MSRV in the README.md

mominul:
This adds an MSRV badge in the `README.md` similar to the [`regex` project](https://github.com/rust-lang/regex/blob/master/README.md).
I got the MSRV from the `CHANGELOG.md`

Thanks!


Amanieu:
Thanks!

PR #204, opened 2020-09-29
fix: Only instantiate RawTable's reserve functions once

Marwes:
...per key-value.

Each of the previous closures would cause an entirely new reserve/resize
function to be instantiated. By using this trick (which std uses for
iterator adaptors), we always get a single instantiatiation per key
(modulo the insert_with_hasher method).

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -218,7 +218,7 @@ pub(crate) fn make_hasher<K: Hash, V>(
 /// Ensures that a single closure type across uses of this which, in turn prevents multiple
 /// instances of any functions like RawTable::reserve from being generated
 #[cfg_attr(feature = "inline-more", inline)]
-fn equivalent<Q, K, V>(k: &Q) -> impl Fn(&(K, V)) -> bool + '_
+fn equivalent_key<Q, K, V>(k: &Q) -> impl Fn(&(K, V)) -> bool + '_
 where
     K: Borrow<Q>,
     Q: ?Sized + Eq,
@@ -229,7 +229,7 @@ where
 /// Ensures that a single closure type across uses of this which, in turn prevents multiple
 /// instances of any functions like RawTable::reserve from being generated
 #[cfg_attr(feature = "inline-more", inline)]
-fn equivalent_single<Q, K>(k: &Q) -> impl Fn(&K) -> bool + '_
+fn equivalent<Q, K>(k: &Q) -> impl Fn(&K) -> bool + '_
 where
     K: Borrow<Q>,
     Q: ?Sized + Eq,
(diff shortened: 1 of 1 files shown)


src/map.rs
```diff
@@ -206,6 +206,29 @@ impl<K: Clone, V: Clone, S: Clone> Clone for HashMap<K, V, S> {
     }
 }
 
+#[cfg_attr(feature = "inline-more", inline)]
```
Amanieu, line 209: Can you add comments here explaining why these wrappers are used.

src/map.rs
```diff
@@ -206,6 +206,29 @@ impl<K: Clone, V: Clone, S: Clone> Clone for HashMap<K, V, S> {
...
+pub(crate) fn make_hasher<K: Hash, V>(
+    hash_builder: &impl BuildHasher,
+) -> impl Fn(&(K, V)) -> u64 + '_ {
+    move |val| make_hash(hash_builder, &val.0)
+}
+
+fn equivalent<Q, K, V>(k: &Q) -> impl Fn(&(K, V)) -> bool + '_
```
Amanieu, line 216: Add `#[cfg_attr(feature = "inline-more", inline)]`.

src/map.rs
```diff
@@ -664,8 +687,7 @@ where
     #[cfg_attr(feature = "inline-more", inline)]
     pub fn reserve(&mut self, additional: usize) {
         let hash_builder = &self.hash_builder;
-        self.table
-            .reserve(additional, |x| make_hash(hash_builder, &x.0));
+        self.table.reserve(additional, make_hasher(hash_builder));
```
Amanieu, line 690:
```suggestion
        self.table.reserve(additional, make_hasher(&self.hash_builder));
```
The `let` above was only needed because the closure would otherwise capture all of `self` instead of just the `hash_builder` field.

src/map.rs
```diff
@@ -1974,16 +2002,16 @@ impl<'a, K, V, S> RawVacantEntryMut<'a, K, V, S> {
         S: BuildHasher,
     {
         let hash_builder = self.hash_builder;
```
Amanieu, line 2004: This line is unnecessary.

src/map.rs
```diff
@@ -206,6 +206,37 @@ impl<K: Clone, V: Clone, S: Clone> Clone for HashMap<K, V, S> {
...
+    K: Borrow<Q>,
+    Q: ?Sized + Eq,
+{
+    move |x| k.eq(x.0.borrow())
+}
+
+/// Ensures that a single closure type across uses of this which, in turn prevents multiple
+/// instances of any functions like RawTable::reserve from being generated
+#[cfg_attr(feature = "inline-more", inline)]
+fn equivalent_single<Q, K>(k: &Q) -> impl Fn(&K) -> bool + '_
```
Amanieu, line 232: This should probably be called `equivalent` and the other one should be renamed to `equivalent_key`.

Amanieu:
Thanks!

@bors r+

PR #205, opened 2020-09-29
Reduce the amount of llvm IR instantiated

Marwes:
This works to improve the generated code in a similar way to #204 , however it is entirely orthogonal to it but also far more invasive.

The main change here is the introduction of `RawTableInner` which is contains all the fields that `RawTable` has but without being parameterized by `T`. Methods have been moved to this new type in parts or their entirety and `RawTable` forwards to these methods.

For this small test case with 4 different maps there is a reduction of the number of llvm lines generated by -17% (18088 / 21873 =0.82695560737) .

```rust
fn main() {
    let mut map1 = hashbrown::HashMap::new();
    map1.insert(1u8, "");
    map1.reserve(1000);
    let mut map2 = hashbrown::HashMap::new();
    map2.insert(1i16, "");
    map2.reserve(1000);
    let mut map3 = hashbrown::HashMap::new();
    map3.insert(3u16, "");
    map3.reserve(1000);
    let mut map4 = hashbrown::HashMap::new();
    map4.insert(3u64, "");
    map4.reserve(1000);
    dbg!((
        map1.iter().next(),
        map2.iter().next(),
        map3.iter().next(),
        map4.iter().next()
    ));
}

```

The commits are almost entirely orthogonal (except the first which does the main refactoring to support the rest) and if some are not desired they can be removed. If it helps, this PR could also be split up into multiple.

For most commitst I don't expect any performance degradation (unless LLVM stops inlining some function as it is now called more), but there are a couple of commits that does slow parts down however these should only be in the cold parts of the code (for instance, panic handling).


src/raw/mod.rs
```diff
@@ -1053,144 +879,454 @@ impl<T> RawTable<T> {
...
+                unsafe { NonNull::new_unchecked(self.table.ctrl.as_ptr().sub(ctrl_offset)) },
+                layout,
+            ))
+        };
+        mem::forget(self);
+        alloc
+    }
+}
+
+impl RawTableInner {
```
Amanieu, line 995: All methods here should be `#[inline]` since they are non-generic.

Amanieu (commented):
Could you try running the benchmarks (`cargo bench`) to see if any of the changes have a performance impact?

Marwes:
Didn't bother to inline the private functions since I think it is redundant within the same compilation unit(?) But I guess they are used from generic functions which causes them to be used from another compilation unit.

That seemed to be enough to restore performance though.

```
 name                         old ns/iter  new ns/iter  diff ns/iter   diff %  speedup
 clone_from_large             1,272        1,325                  53    4.17%   x 0.96
 clone_from_small             17           17                      0    0.00%   x 1.00
 clone_large                  1,371        1,362                  -9   -0.66%   x 1.01
 clone_small                  32           32                      0    0.00%   x 1.00
 insert_ahash_highbits        8,753        8,957                 204    2.33%   x 0.98
 insert_ahash_random          8,982        8,872                -110   -1.22%   x 1.01
 insert_ahash_serial          8,462        8,141                -321   -3.79%   x 1.04
 insert_erase_ahash_highbits  20,480       20,212               -268   -1.31%   x 1.01
 insert_erase_ahash_random    19,826       19,715               -111   -0.56%   x 1.01
 insert_erase_ahash_serial    19,543       19,168               -375   -1.92%   x 1.02
 insert_erase_std_highbits    38,157       38,102                -55   -0.14%   x 1.00
 insert_erase_std_random      38,685       38,385               -300   -0.78%   x 1.01
 insert_erase_std_serial      31,856       32,867              1,011    3.17%   x 0.97
 insert_std_highbits          17,661       17,811                150    0.85%   x 0.99
 insert_std_random            17,952       17,970                 18    0.10%   x 1.00
 insert_std_serial            15,414       15,504                 90    0.58%   x 0.99
 iter_ahash_highbits          1,153        1,201                  48    4.16%   x 0.96
 iter_ahash_random            1,060        1,203                 143   13.49%   x 0.88
 iter_ahash_serial            1,076        1,141                  65    6.04%   x 0.94
 iter_std_highbits            1,112        1,087                 -25   -2.25%   x 1.02
 iter_std_random              1,239        1,111                -128  -10.33%   x 1.12
 iter_std_serial              1,171        1,117                 -54   -4.61%   x 1.05
 lookup_ahash_highbits        5,998        5,831                -167   -2.78%   x 1.03
 lookup_ahash_random          5,611        5,674                  63    1.12%   x 0.99
 lookup_ahash_serial          5,405        5,393                 -12   -0.22%   x 1.00
 lookup_fail_ahash_highbits   5,407        5,167                -240   -4.44%   x 1.05
 lookup_fail_ahash_random     5,182        5,480                 298    5.75%   x 0.95
 lookup_fail_ahash_serial     5,014        5,475                 461    9.19%   x 0.92
 lookup_fail_std_highbits     14,310       14,504                194    1.36%   x 0.99
 lookup_fail_std_random       14,117       14,548                431    3.05%   x 0.97
 lookup_fail_std_serial       12,181       12,272                 91    0.75%   x 0.99
 lookup_std_highbits          14,178       14,635                457    3.22%   x 0.97
 lookup_std_random            14,449       14,862                413    2.86%   x 0.97
 lookup_std_serial            12,059       12,254                195    1.62%   x 0.98

```

Amanieu:
The changes mostly look good. However they are quite invasive so I would feel more comfortable if we could test the impact of this change in compilation performance.

Could you try creating a PR in rust-lang/rust which switches libstd to use your branch of hashbrown so that we can do a perf run?

Marwes:
Rebased and squashed history into a decent state. Perf results at https://github.com/rust-lang/rust/pull/77566#issuecomment-765529457 looks good so I believe this should be good for review again now!

src/raw/mod.rs
```diff
@@ -1055,224 +889,566 @@ impl<T, A: Allocator + Clone> RawTable<T, A> {
...
+            }
+            probe_seq.move_next(self.bucket_mask);
+        }
+    }
+
+    #[allow(clippy::mut_mut)]
+    unsafe fn prepare_rehash_in_place<'s>(
+        &'s mut self,
+        needs_drop: bool,
+        drop: fn(&mut Self, usize),
```
Amanieu, line 1204: This looks very suspicious to me. Are you sure this is actually necessary to improve compile times?

I would have expected the guard to be optimized away most of the time since `Hash::hash` almost never panics (and the compiler knows this since it inlines the hash function).

Also, this function is non-generic so the lack of a `#[inline]` makes it impossible to inline.

src/raw/mod.rs
```diff
@@ -259,20 +294,7 @@ fn calculate_layout<T>(buckets: usize) -> Option<(Layout, usize)> {
...
-        .checked_mul(buckets)?
-        .checked_add(ctrl_align - 1)?
-        & !(ctrl_align - 1);
-    let len = ctrl_offset.checked_add(buckets + Group::WIDTH)?;
-
-    Some((
-        unsafe { Layout::from_size_align_unchecked(len, ctrl_align) },
-        ctrl_offset,
-    ))
+    TableLayout::new::<T>().calculate_layout_for(buckets)
```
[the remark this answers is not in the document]
Marwes: The only place `calculate_layout<T>` is used now is in `into_allocation` so I removed the nightly specialization

src/raw/mod.rs
Marwes: > This looks very suspicious to me. Are you sure this is actually necessary to improve compile times?

Can't say for sure since `llvm-lines` only indirectly measures compile times, in the llvm-lines I do not see if the guard itself would get optimized away.  However I use a `fn` argument here since the `drop` function would practically never be called since the non-panic path will always `forget` the guard.

Amanieu: However with this you are forcing LLVM to perform full codegen on the drop function, whereas it could previously have been eliminated in one of the early LLVM passes as unreachable code.

Could you try moving the guard creation back into `rehash_in_place` to see how much it affects performance?

Marwes: Moving the guard out forces the guard to be specialized for each `T` and increases the number of llvm lines by ~1% . Wall time is to small and has to much variance to be measured, possibly a rust perf run could show something but probably not.

I suppose llvm may codegen the small drop closure eagerly but I think that on balance, having the rest of the `drop` closure in the drop guard being only instantiated once per allocator to outweigh that (my test only has 4 different hashmaps, the wins on grow as you have more hashmaps).

```
               |self_: &mut RawTableInner<A>, index| {
                    self_.bucket::<T>(index).drop();
                },
```

Amanieu: The `drop` closure still needs to be instantiated for each `T`. The only thing we are saving is that `guard` and `Scopeguard::drop` are only instantiated for each `A` instead of for each `T`.

Since `Scopeguard::drop` is likely to be optimized away as unreachable, I expect this change to have negligible effect on compile times while improving the performance of `rehash_in_place`.

Could you try a perf run with the guard moved back into `rehash_in_place`? If the result is worse than before then I'm happy to merge this PR as it is.

Marwes: ## guard directly in rehash_in_place

https://perf.rust-lang.org/compare.html?start=84864bfea9c00fb90a1fa6e3af1d8ad52ce8f9ec&end=3ca26a5875f0c4d0522408aa4284b042697b23e4

## guard in prepare_rehash_in_place

https://perf.rust-lang.org/compare.html?start=25f39fe80293f77bd86f64a1261a3e2c0ca23847&end=97c7b0501b746bb7191e480b2076cf601f6d0ca4

Seems to be a bit better on average to keep the guard directly in `rehash_in_place` (even if the biggest improvement is a bit smaller).

Marwes: Considering `rehash_in_place` isn't tested I'd be inclined to remove it entirely.

Amanieu:
Thanks for the great work @Marwes!

Marwes:
Still need to fix the debug info issue in rust-lang/rust but do you know when this might be released? In case I should push hard on fixing that issue or take it easy.

Amanieu:
I just need to get #225 fixed then I can publish a release. Most likely some time later this week.

src/set.rs
```diff
@@ -452,8 +452,6 @@ impl<T, S> HashSet<T, S, Global> {
 
 impl<T, S, A> HashSet<T, S, A>
 where
-    T: Eq + Hash,
-    S: BuildHasher,
```
pickfire, line 456: Why is this changed?

Amanieu: The bounds for `with_hasher` were removed. This removes them for `with_hasher_in` as well.

PR #207, opened 2020-10-12
Enable specialization with aHash

tkaitchuck:
This uses the specialization feature in nightly to improve performance for hashing some classes (u8, u16, u32, u64, u128, [u8], and String) For these classes it improves performance by 15% the test `insert_ahash` benchmarks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,45 +39,44 @@ Compared to the previous implementation of `std::collections::HashMap` (Rust 1.3
 
 With the hashbrown default AHash hasher:
 
-```text
- name                       oldstdhash ns/iter  hashbrown ns/iter  diff ns/iter   diff %  speedup 
- insert_ahash_highbits        20,846              7,397                   -13,449  -64.52%   x 2.82 
- insert_ahash_random          20,515              7,796                   -12,719  -62.00%   x 2.63 
- insert_ahash_serial          21,668              7,264                   -14,404  -66.48%   x 2.98 
- insert_erase_ahash_highbits  29,570              17,498                  -12,072  -40.83%   x 1.69 
- insert_erase_ahash_random    39,569              17,474                  -22,095  -55.84%   x 2.26 
- insert_erase_ahash_serial    32,073              17,332                  -14,741  -45.96%   x 1.85 
- iter_ahash_highbits          1,572               2,087                       515   32.76%   x 0.75 
- iter_ahash_random            1,609               2,074                       465   28.90%   x 0.78 
- iter_ahash_serial            2,293               2,120                      -173   -7.54%   x 1.08 
- lookup_ahash_highbits        3,460               4,403                       943   27.25%   x 0.79 
- lookup_ahash_random          6,377               3,911                    -2,466  -38.67%   x 1.63 
- lookup_ahash_serial          3,629               3,586                       -43   -1.18%   x 1.01 
- lookup_fail_ahash_highbits   5,286               3,411                    -1,875  -35.47%   x 1.55 
- lookup_fail_ahash_random     12,365              4,171                    -8,194  -66.27%   x 2.96 
- lookup_fail_ahash_serial     4,902               3,240                    -1,662  -33.90%   x 1.51 
-```
+| name                    |  oldstdhash ns/iter | hashbrown ns/iter | diff ns/iter |  diff %  | speedup |
+|:------------------------|:-------------------:|------------------:|:------------:|---------:|---------|
+| insert_ahash_highbits       | 18,865      | 8,020         |               -10,845 | -57.49%  | x 2.35 |
+| insert_ahash_random         | 19,711      | 8,019         |               -11,692 | -59.32%  | x 2.46 |
+| insert_ahash_serial         | 19,365      | 6,463         |               -12,902 | -66.63%  | x 3.00 |
+| insert_erase_ahash_highbits | 51,136      | 17,916        |               -33,220 | -64.96%  | x 2.85 |
+| insert_erase_ahash_random   | 51,157      | 17,688        |               -33,469 | -65.42%  | x 2.89 |
+| insert_erase_ahash_serial   | 45,479      | 14,895        |               -30,584 | -67.25%  | x 3.05 |
+| iter_ahash_highbits         | 1,399       | 1,092         |                  -307 | -21.94%  | x 1.28 |
+| iter_ahash_random           | 1,586       | 1,059         |                  -527 | -33.23%  | x 1.50 |
+| iter_ahash_serial           | 3,168       | 1,079         |                -2,089 | -65.94%  | x 2.94 |
+| lookup_ahash_highbits       | 32,351      | 4,792         |               -27,559 | -85.19%  | x 6.75 |
+| lookup_ahash_random         | 17,419      | 4,817         |               -12,602 | -72.35%  | x 3.62 |
+| lookup_ahash_serial         | 15,254      | 3,606         |               -11,648 | -76.36%  | x 4.23 |
+| lookup_fail_ahash_highbits  | 21,187      | 4,369         |               -16,818 | -79.38%  | x 4.85 |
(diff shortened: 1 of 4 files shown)


tkaitchuck:
This uses aHash 0.5.1 which no longer depends on const-random by default. See https://github.com/tkaitchuck/aHash/commit/980159d060f932531583c27f407c21020124a76c

Amanieu:
CI is failing.

tkaitchuck:
I fixed the formatting / unused imports. However I am not sure what to do about this one:
https://travis-ci.com/github/rust-lang/hashbrown/jobs/398173552
```
error[E0463]: can't find crate for `std`
 --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/lazy_static-1.4.0/src/inline_lazy.rs:9:1
  |
9 | extern crate std;
  | ^^^^^^^^^^^^^^^^^ can't find crate
  |
  = note: the `thumbv6m-none-eabi` target may not be installed
```
It appears that lazy_static depends on std and it is a dev dependency in hashbrown because it is used in the rayon test. I am not sure why this is failing now, but wasn't before.

tkaitchuck:
@Amanieu Does this approach make sense? 
I have added a `std` flag to HashBrown which is off by default. So by default it uses aHash with constant keys, but if `std` or `ahash-compile-time-rng` it will use those for randomization. (With preference given to runtime randomization) 

My concern is that it is something of a downgrade for `no-std` environments which by default previously were getting the week randomization based on memory address and now will be getting just fixed constants. Obviously a memory address is in a `no-std` environment might be far from random, but at least one couldn't do this: https://accidentallyquadratic.tumblr.com/post/153545455987/rust-hash-iteration-reinsertion

tkaitchuck:
Right now if both flags are off I have it doing:
`pub type DefaultHashBuilder = core::hash::BuildHasherDefault<ahash::AHasher>;`
I could change this behavior in aHash to not always produce an identical build hasher when using the default instantiation. That would resolve the issue I mentioned above. However I am reluctant to do this as it is dis-similar to the way the standard library works. Specifically [DefaultHasher's new method](https://doc.rust-lang.org/std/collections/hash_map/struct.DefaultHasher.html#method.new) says the produced value is "the same as all other DefaultHasher instances created through new or default".

src/map.rs
```diff
@@ -239,9 +249,19 @@ where
 
 #[cfg_attr(feature = "inline-more", inline)]
 pub(crate) fn make_hash<K: Hash + ?Sized>(hash_builder: &impl BuildHasher, val: &K) -> u64 {
-    let mut state = hash_builder.build_hasher();
-    val.hash(&mut state);
-    state.finish()
+    #[cfg(feature = "ahash")]
```
Amanieu, line 252: A comment explaining what is going on here would be helpful (specialization to improve performance).

Amanieu:
I think the current behavior is fine. Using constant keys should also improve performance a bit.

README.md
```diff
@@ -96,19 +96,22 @@ use hashbrown::HashMap;
...
 map.insert(1, "one");
 ```
-
+## Flags
 This crate has the following Cargo features:
 
+- `inline-more`: Adds inline hints to most functions, improving run-time performance at the cost
+   of compilation time. (enabled by default)
+- `ahash`: Compiles with ahash as default hasher. (enabled by default)
+- `std`: Enables use of features that depend on the standard library.
```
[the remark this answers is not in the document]
tkaitchuck: What about `std` for `serde`. Is that never of any use here?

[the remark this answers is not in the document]
Amanieu: We only use serde to provide trait impls. It is up to the users of serde to decide whether they need `std` or not.

src/map.rs
tkaitchuck: Done

README.md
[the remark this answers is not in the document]
tkaitchuck: Done

README.md
```diff
@@ -96,19 +96,22 @@ use hashbrown::HashMap;
...
+   If `ahash` is used this includes using random keys to provide DOS resistance. (disabled by default)
+- `ahash-compile-time-rng`: This is an alternative to `std` which still allows for some degree of DOS-resistance.
+   However, it can result in issues for certain platforms.
```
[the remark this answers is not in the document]
tkaitchuck: Done

tkaitchuck:
This [failed on thumbv6m-none-eabi](https://travis-ci.com/github/rust-lang/hashbrown/jobs/402651379) apparently because atomics are not available on `thumbv6m-none-eabi` per https://github.com/rust-lang/compiler-builtins/issues/114 
Is there a way to detect this availability via a cfg flag?

README.md
```diff
@@ -96,19 +96,22 @@ use hashbrown::HashMap;
...
 
-- `nightly`: Enables nightly-only features: `#[may_dangle]`.
+- `inline-more`: Adds inline hints to most functions, improving run-time performance at the cost
+   of compilation time. (enabled by default)
+- `ahash`: Compiles with ahash as default hasher. (enabled by default)
+- `ahash-run-time-rng`: Uses randomly generated keys for each hashmap to provide DOS resistance.
+   This requires the standard library. (disabled by default)
+- `ahash-compile-time-rng`: This is an alternative to `ahash-run-time-rng` that works by pre-generating keys at 
+   compile time and embedding them as constants. The avoids the dependency on the standard library but means the 
+   binary will be slightly different each time it is compiled. (disabled by default)
```
Amanieu, line 109: A thought occurs: is it really necessary to have these two flags at all? In `hashbrown` we make no guarantees about DoS resistance so we can simply always use constant keys. I would expect anyone needing DoS resistance to use the `ahash` crate directly (enabling any features they need).

tkaitchuck: That would work if we didn't need to think about version upgrades. A user could depend of HashBrown 0.9.1 and aHash 0.5.6 and set the flags on aHash that they want, and everything would work as expected. However, if they upgrade to HashBrown 0.9.2 and it happens to depend on aHash 0.6.0 and they don't update the version of their direct dependency on aHash then cargo will pull in two versions of aHash. HashBrown will use one without the flags set, and the other will be unused, unless the code is manually specifying the hasher each time. So suddenly the flags will have no effect in some places and they won't notice.

Amanieu: Anyone relying on those flags should explicitly use `HashMap<K, V, AHasher>` where `AHasher` is imported from their version of the `ahash` crate, not the one hashbrown uses.

tkaitchuck: What about the case where there is an intermediate library? For example: Suppose there is a library for data processing that depends on HashBrown. If the library is generic, and does not know if it will be used in a context that where `std` is available or if DOS resistance is required. 

From your above answer it sounds like such a library not depend on HashBrown. Is this the case? I was attempting to model this case on the way many crates handle `std` or `serde` where it's a flag which is propagated to the application via intermediate dependencies. 

Ideally there would be a way to do this without all the intermediate dependencies being "aware" or it or the end application being tied to a specific version of its dependencies. But I am not aware of any way to do that within Cargo.

tkaitchuck: As thing are, there is no performance advantage to fixed keys. However when dos resistance is not a concern: AHasher has a method `short_finish()` which is a faster alternative to `finish` (at some cost to quality). It is mutually exclusive with specialization for primitives though.

Amanieu: Maybe this could be exposed on the `ahash` side as two separate hasher types: one with dos resistance and one without. This would neatly solve the issue.

tkaitchuck: I have created https://github.com/tkaitchuck/aHash/issues/58 to track that. (as doing that properly will take some time.)

tkaitchuck:
@Amanieu Please take another look. I believe this approach is a lot better.

tkaitchuck:
Remaining build issue is a pre-existing problem: https://github.com/rust-lang/hashbrown/pull/214

Amanieu:
LGTM once the CI issues are fixed.

Amanieu:
Needs a rebase.

tkaitchuck:
Done

PR #209, opened 2020-10-30
Better branch likelyness on stable

a1phyr:
It uses uses the `#[cold]` attribute to get a similar effect as `core::intrinsics::{likely, unlikely}` on stable (10-15% on some benchmarks)

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -43,14 +43,23 @@ use self::imp::Group;
 // consistently improves performance by 10-15%.
 #[cfg(feature = "nightly")]
 use core::intrinsics::{likely, unlikely};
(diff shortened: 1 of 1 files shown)


src/raw/mod.rs
```diff
@@ -43,14 +43,23 @@ use self::imp::Group;
 // consistently improves performance by 10-15%.
 #[cfg(feature = "nightly")]
 use core::intrinsics::{likely, unlikely};
+
+#[cfg(not(feature = "nightly"))]
```
Amanieu, line 47: Can you add a comment here explaining that LLVM always treats calls to cold functions as unlikely?

PR #223, opened 2021-01-16
Export AllocError as well as Allocator.

cfallin:
This PR re-exports the `AllocError` type alongside `Allocator` and others from the `raw` submodule.

This seems to be necessary to implement the `Allocator` trait from outside the crate (and I want to use exported types from this crate so that I don't have to depend on the unstable allocator API directly). It allowed me to instantiate hashbrown hashmaps using `bumpalo` over in bytecodealliance/regalloc.rs#115; I couldn't work out another way to do this. If there is a more direct way, please do let me know! (Thanks for this excellent crate and the configurable allocator interface, in any case!)

### DIFF
--- a/src/raw/alloc.rs
+++ b/src/raw/alloc.rs
@@ -3,7 +3,7 @@ pub use self::inner::*;
 #[cfg(feature = "nightly")]
 mod inner {
     use crate::alloc::alloc::Layout;
-    pub use crate::alloc::alloc::{Allocator, Global};
+    pub use crate::alloc::alloc::{AllocError, Allocator, Global};
     use core::ptr::NonNull;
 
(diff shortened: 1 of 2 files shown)


Amanieu:
CI is failing on nightly.

cfallin:
Thanks! Would it be possible for you to consider doing a point-release that includes this?

Amanieu:
Published 0.10.0

Amanieu:
I've had to yank 0.10.0 due to #225 and #226. I'm hoping to have these issues resolved soon. One notable change is that we cant't export a custom `Allocator` trait, however we can have bumpalo-specific variants of `HashMap`, `HashSet` and `RawTable` supported.

Issue #224, opened 2021-01-16
Feature request: insertion that bails if reallocation would occur


cole-miller:
For a custom data structure I'm implementing that uses `hashbrown::HashMap` internally, I'd like to access something like the following interface:

```rust
// returns the value of `map.insert(key, val)` (wrapped in `Ok`) if this call would not cause `map` to reallocate
// if reallocation would occur, returns `Err((key, val))` without performing the insertion
fn try_insert_no_grow<K: Eq + Hash, V, S: BuildHasher>(
    map: &mut HashMap<K, V, S>,
    key: K,
    val: V,
) -> Result<Option<V>, (K, V)> {
    unimplemented!()
}
```

Preemptively comparing `map.len()` to `map.capacity()` will not achieve this precisely due to the "lower bound" nature of `HashMap::capacity` mentioned in the docs. I want this power so that I can determine precisely when reallocation occurs and "rebalance" the containing data structure at the same time.

I asked about how to achieve this on URLO [here](https://users.rust-lang.org/t/checking-whether-hashmap-insert-would-reallocate/54205) and didn't get any leads on how to do it using the current interface (including `RawTable`), but if there *is* a way of course I'd love to hear about it.

cole-miller:
closed by 4c3ba70a4eeef951d1841a5a2879849cfc9f62dc

PR #229, opened 2021-01-19
Add `try_insert_no_grow` method on `RawTable`

cole-miller:
This is my attempt to answer the need described in #224 by adding a method on `RawTable`:

```rust
impl<T> RawTable<T, Global> {
    pub fn try_insert_no_grow(&mut self, hash: u64, value: T) -> Result<Bucket<T>, T>;
}
```

This method follows the logic of `RawTable::insert`, but if the table would need to grow to accomodate the new element it "signals" by returning `Err(value)` instead of performing the reallocation and insertion.

Marked as draft for now because

- I'm not sure whether there's a better way to achieve this (advice welcome)
- some tests are probably needed
- there's an unused warning (despite the `pub`) that I haven't figured out how to get rid of

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## Added
+- Added safe `try_insert_no_grow` method to `RawTable`. (#229)
+
 ## Changed
 - The minimum Rust version has been bumped to 1.49.0. (#230)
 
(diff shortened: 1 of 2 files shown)


cole-miller:
Fixed the unused warning by adding `#[cfg(feature = "raw")]` -- not sure if that's the best solution.

The latest CI failure seems to be due to a (new?) Clippy lint `ptr_as_ptr` triggering on other parts of the code.

Amanieu:
> Fixed the unused warning by adding #[cfg(feature = "raw")] -- not sure if that's the best solution.

That's the correct solution.

Code LGTM, just need to wait for #230 to land first.

cole-miller:
Thanks!

PR #231, opened 2021-01-21
Add support for bumpalo

Amanieu:
Allows `Bump` and `&Bump` to be used as allocators without the nightly feature.

This is currently blocked on https://github.com/fitzgen/bumpalo/pull/92 for support on nightly.

### DIFF
--- a/src/raw/alloc.rs
+++ b/src/raw/alloc.rs
@@ -13,6 +13,19 @@ mod inner {
             .map(|ptr| ptr.as_non_null_ptr())
             .map_err(|_| ())
     }
+
+    #[cfg(feature = "bumpalo")]
+    unsafe impl Allocator for crate::BumpWrapper<'_> {
+        #[inline]
+        fn allocate(&self, layout: Layout) -> Result<NonNull<[u8]>, core::alloc::AllocError> {
+            match self.0.try_alloc_layout(layout) {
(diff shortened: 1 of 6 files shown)


src/raw/alloc.rs
```diff
@@ -47,4 +47,20 @@ mod inner {
     pub fn do_alloc<A: Allocator>(alloc: &A, layout: Layout) -> Result<NonNull<u8>, ()> {
         alloc.allocate(layout).map_err(|_| ())
     }
+
+    #[cfg(feature = "bumpalo")]
+    unsafe impl Allocator for bumpalo::Bump {
+        fn allocate(&self, layout: Layout) -> Result<NonNull<u8>, AllocError> {
```
baloo, line 53: Just curious, I'm comparing this with the implementation I made upstream, and the signature looks wrong.
`allocator_api` signature should be `NonNull<[u8]>`. That makes me require the `nonnull_slice_from_raw_parts` feature.

It feels like I'm missing something here.

Amanieu: That's just an internal `Allocator` trait, it is completely private and can't be implemented outside this crate. As such, it doesn't *have* to match the standard library `Allocator` trait.

baloo: ha! I missed that, thank you!

PR #233, opened 2021-01-27
test: Add a test for rehash_in_place

Marwes:
No tests ran the `rehash_in_place` method before and I could not come up with a way to trigger the rehash from the higher level structures so explicitly calling the method is the best I could do.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -9,7 +9,7 @@ extern crate test;
 use test::{black_box, Bencher};
 
 use hashbrown::hash_map::DefaultHashBuilder;
-use hashbrown::HashMap;
+use hashbrown::{HashMap, HashSet};
 use std::{
     collections::hash_map::RandomState,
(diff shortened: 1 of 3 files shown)


Amanieu:
You should be able to trigger `rehash_in_place` by continuously adding and removing elements in such a way that the number of elements in the map stays the same:
```rust
for i in 0..100 {
    set.insert(i);
}
for i in 100..10000 {
    set.remove(i - 100);
    set.insert(i);
}
```

Marwes:
That worked after carefully adjusting the capacity. Tried adding a benchmark as well (since rustc seems to spend a decent amount of time in rehashing, unless I misremember).

Tried using the benchmark to test simd searching for `DELETED` elements in `rehash_in_place` but I don't see any difference in either direction :( . https://github.com/Marwes/hashbrown/commit/46aba3e4e0ce9500fb036d41e8f3b77b8944dacc

PR #235, opened 2021-02-04
Implement `From<HashMap<T, ()>>` for `HashSet<T>`.

### LINKED ISSUE #219: Implement `From<HashMap<T, ()>>` for `HashSet<T>`

smmalis37:
Closes #219.

### DIFF
--- a/src/set.rs
+++ b/src/set.rs
@@ -1107,6 +1107,15 @@ where
     }
 }
 
+impl<T, S, A> From<HashMap<T, (), S, A>> for HashSet<T, S, A>
+where
+    A: Allocator + Clone,
+{
+    fn from(map: HashMap<T, (), S, A>) -> Self {
+        Self { map }
+    }
+}
+
 impl<T, S, A> FromIterator<T> for HashSet<T, S, A>
 where
     T: Eq + Hash,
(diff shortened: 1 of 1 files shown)


smmalis37:
Those clippy CI failures don't look related to this?

Amanieu:
This happens when clippy adds a new lint. Can you add the `# Panics` section to the doc comments that need it? You can do it either in this PR or in a separate one.

smmalis37:
I could if I understood why these functions can panic. I attempted to dig into the codebase, but I can't understand why https://github.com/rust-lang/hashbrown/blob/83ac6fd0d364bc220f7d24cc234bc0c4ab30b3ae/src/map.rs#L3162 sets Key to None. As far as I can see this is the only way to get `key` set to None, which will cause these panics later. Example code that panics:

```rust
let mut m: HashMap<i32, i32> = HashMap::new();
let o = m.entry(0).insert(1);
let (k, v) = o.replace_entry(2);
```

Why does OccupiedEntry have `key` as an Option? Could it be changed to always be present? If it can't, can these two unwraps be changed to handle this case better?

Amanieu:
`key` is the key that was used to perform the lookup. It is the value passed to `HashMap::entry`. However it is `None` when the `OccupiedEntry` is created from `VacantEntry::insert_entry` since the `key` is then moved into the `HashMap` to create the entry.

smmalis37:
I suppose I can go add `Panics if this OccupiedEntry was created through Entry::insert` to both of those functions, but that feels like a really weird and unfortunate restriction to have. I'm completely new to this codebase, so forgive me for asking, but is there really no better option?

Amanieu:
I don't think there is a better option.

smmalis37:
This should be good to go now I think.

PR #237, opened 2021-02-14
Implement `Default` for `RawTable`

cole-miller:
This seems straightforward and useful. (I found myself wanting it so I could use `core::mem::take` on a `RawTable` field.) Hopefully I'm not missing something.


cole-miller:
Thanks!

PR #239, opened 2021-02-15
Add `get_each_mut` methods on `RawTable` and `HashMap`

### LINKED ISSUE #151: Features request: Mutably lookup multiple distinct values
In the current interface looking up a single key will borrow the entire map and in particular a mutable lookup requires a unique borrow of the entire map. This is very safe but also overly restrictive. In comparison, slices (as a very primitive key-value-map) permit this same operation with a `split_mut` method which for example allows defining an entirely safe method for acquiring elements at two different indices at once:

```rust
fn at2<T>(slice: &mut [T], a: usize, b: usize) -> (&mut T, &mut T) {
    let (_, slice) = slice.split_at_mut(cmp::min(a, b));
    let (head, tail) = slice.split_at_mut(cmp::max(a, b) - cmp::min(a, b));
}
```

If we try to replace the slice with a HashMap, for example as a sparse data structure, however we hit a brickwall. There is a bad workaround: Use an iterator over *all* mutable reference, filter it by the two elements we are interested in and return them. This is both inefficient, unergonomic, and brittle.

```rust
fn at2<T>(map: &mut HashMap<usize, T>, a: usize, b: usize) -> (&mut T, &mut T) {
    let mut values: HashMap<_, _> = map.iter_mut().collect();
    (values.remove(&a).unwrap(), values.remove(&b).unwrap())
}
```

This is embarrisingly expensive, cloning the entire map and recalculating all hashes, while the existence of the safe workaround should also make it clear that there is no formal issue preventing the simultaneous use of multiple values (the backing memory is allocated similar to a vector) while querying keys.

## A few remarks on a first draft

I'm not quite sure how the interface should look like. It could be some simple interface returning raw pointers to elements which requires the caller to implement correct safety and uniqueness guarantees.

There is also a more ambitious solution. Do note that `raw::RawTable` has the interfaces needed to build a less expensive safe abstraction but it is too `raw`. It is both very complex in that almost everything is `unsafe` and not exposed in the standard library :slightly_smiling_face: 

Let's think about how the overhead of the workaround could be lessened by using internal interface. Such a virtual clone would only need to support removal, and no insertion. It would borrow most data structures from the original map. By not allowing insertion, we would not have to recalculate the hashes or do rehashing. It also need not actually remove elements but only mark them borrowed, which can be even cheaper than cloning and modifying the control bytes. This marking could even use a small-vec optimization if only few elements have already been borrowed.

cole-miller:
This PR adds a new method `get_each_mut` on `RawTable` and new methods `get_each_mut` and `get_each_key_value_mut` on `HashTable`. These methods enable looking up mutable references to several table or map entries at once. They take advantage of min_const_generics, which is available without a feature gate on recent nightly (but not stable) compilers. Hence everything is behind `#[cfg(feature = "nightly")]` for now.

Closes #151 (cc @HeroicKatora).

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -1,11 +1,15 @@
 use crate::alloc::alloc::{handle_alloc_error, Layout};
 use crate::scopeguard::guard;
 use crate::TryReserveError;
+#[cfg(feature = "nightly")]
+use crate::UnavailableMutError;
 use core::hint;
 use core::iter::FusedIterator;
 use core::marker::PhantomData;
 use core::mem;
 use core::mem::ManuallyDrop;
+#[cfg(feature = "nightly")]
+use core::mem::MaybeUninit;
 use core::ptr::NonNull;
 
 cfg_if! {
@@ -316,12 +320,12 @@ impl<T> Bucket<T> {
         }
     }
     #[cfg_attr(feature = "inline-more", inline)]
-    pub unsafe fn as_ptr(&self) -> *mut T {
+    pub fn as_ptr(&self) -> *mut T {
         if mem::size_of::<T>() == 0 {
             // Just return an arbitrary ZST pointer which is properly aligned
             mem::align_of::<T>() as *mut T
         } else {
-            self.ptr.as_ptr().sub(1)
+            unsafe { self.ptr.as_ptr().sub(1) }
         }
     }
     #[cfg_attr(feature = "inline-more", inline)]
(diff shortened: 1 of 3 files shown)


cole-miller:
Okay, fixed the problems I could find on my own, so this should be ready for review.

src/raw/mod.rs
```diff
@@ -944,6 +962,62 @@ impl<T, A: Allocator + Clone> RawTable<T, A> {
...
+            unsafe { MaybeUninit::uninit().assume_init() };
+        for i in 0..N {
+            out[i] = MaybeUninit::new(
+                #[allow(clippy::never_loop)]
+                'outer: loop {
+                    for j in 0..i {
+                        match (&buckets[j], &buckets[i]) {
+                            // These two buckets are the same, and we can't safely return a second
+                            // mutable reference to the same entry.
+                            (Some(prev), Some(cur)) if unsafe { prev.as_ptr() == cur.as_ptr() } => {
```
Amanieu, line 1002: Is the unsafe really necessary here? I think `as_ptr` could probably be changed to be a safe method instead.

Amanieu (commented):
Great work! I think this is probably one of the most awaited features on `HashMap`!

src/raw/mod.rs
cole-miller: I was wondering about that. If there's no reason for it to be `unsafe` I can submit a separate PR removing that annotation (or just do it in this one if that's easier).

src/raw/mod.rs
```diff
@@ -388,6 +392,20 @@ struct RawTableInner<A> {
     alloc: A,
 }
 
+/// We need this workaround because `transmute` can't be used on arrays of generic length.
+///
+/// Nightly has a `MaybeUninit::array_assume_init` function, but it's behind a feature gate (see
+/// rust-lang/rust#80908), so for now we just open-code it here.
```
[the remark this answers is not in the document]
cole-miller: Fair enough, I'll do that.

src/raw/mod.rs
Amanieu: Just do it here.

cole-miller:
Thanks!

PR #242, opened 2021-02-23
always inline `ScopeGuard::drop`

quininer:
I accidentally discovered that some minor changes in a production project will cause a lot of size increase and some performance regression. After comparison, I found a lot of `ScopeGuard::drop` symbols appeared in the poorer version.

I suspect this is because the compiler missed some inline. for `ScopeGuard`, it should always be beneficial to inline it.


Amanieu:
This should probably apply to all the methods in this file.

quininer:
I added more inline attributes to the file.

PR #243, opened 2021-03-01
Use `criterion` for benches

### LINKED ISSUE #4: Making benchmarks better
The benchmarks should really be using [criterion](https://github.com/japaric/criterion.rs)
It would be interesting to benchmark it against [bytell-hash-map](https://github.com/WaDelma/bytell-hash-map) which I did rust implementation for (Haven't optimised it that much and should really make it usable library).

henryboisdequin:
Closes #4


rust-toolchain
```diff
@@ -0,0 +1 @@
+1.49.0
```
Amanieu, line 1: What is the purpose of this file?

.gitignore
```diff
@@ -1,3 +1,7 @@
+# Rust
 /target
 **/*.rs.bk
 Cargo.lock
+
+# Editors
+.vscode
```
Amanieu, line 7: You should put this in `.gitignore` in your home directory rather than in project gitignore configurations.

rust-toolchain
henryboisdequin: This file enforces everyone working on hashbrown to use the 1.49 Rust tool chain.

Amanieu: Unfortunately this breaks our CI. Could you remove this file?

Many people, including myself, work with Rust nightly and need support for nightly features in our work. This includes working with hashbrown which uses several nightly-only features when the `"nightly"` feature is enabled.

henryboisdequin: @Amanieu Done ✅

benches/bench.rs
```diff
@@ -1,331 +1,101 @@
...
-
-bench_suite!(
-    bench_grow_insert,
-    grow_insert_ahash_serial,
-    grow_insert_std_serial,
-    grow_insert_ahash_highbits,
-    grow_insert_std_highbits,
-    grow_insert_ahash_random,
-    grow_insert_std_random
-);
```
Amanieu, line 131: Please don't remove all our benchmarks, we need those to accurately measure the performance of hashbrown in different situations.

henryboisdequin: @Amanieu Where do these come from? As when I test locally, all of these functions are not found.

Amanieu: Just run `cargo bench` with a nightly toolchain.

Amanieu:
You need to fix the CI failures. And restore the benchmark code that you deleted (the function are created by the `macro_rules!`, but you deleted it).

PR #251, opened 2021-03-19
Add `pub` modifier to `BumpWrapper`.

DavidVonDerau:
Without this change, I'm seeing the following error when trying to use the `BumpWrapper` feature in my own code.

```
error[E0423]: cannot initialize a tuple struct which contains private fields
   |
   |             HashSet::new_in(BumpWrapper(arena))
   |                             ^^^^^^^^^^^ constructor is not visible here due to private fields
```

`arena` is of type `&Bump`, matching the example given in the `test_bumpalo` test: `HashMap::new_in(BumpWrapper(&bump));`.


Amanieu:
Nice catch! I'll publish a new release with this fix.

PR #254, opened 2021-03-29
Make `RawTable::insert_no_grow` unsafe

### LINKED ISSUE #253: `RawTable::insert_no_grow` needs safety checks, or should be `unsafe`
This program segfaults:

```rust
use hashbrown::raw::RawTable;
fn main() {
    let mut table = RawTable::new();
    loop {
        table.insert_no_grow(0, 0);
    }
}
```

If I change this to have some initial capacity, then in debug mode it panics decrementing `growth_left`:

```
thread 'main' panicked at 'attempt to subtract with overflow', /home/jistone/.cargo/registry/src/github.com-1ecc6299db9ec823/hashbrown-0.11.2/src/raw/mod.rs:882:13
```

In release mode that loops forever. I added debug prints for len/capacity, and that gets up to 16/14 before hanging entirely.

I suppose panic/infinite-loop is somewhat acceptable from a safety perspective, as long as we're sure to do one of those without corrupting memory, but the empty case is definitely a problem. Maybe we should assert the same condition that the regular `insert` uses when it decides to resize? We could even use/mimic `try_insert_no_grow` and then panic on error.

cuviper:
For performance reasons, this method _assumes_ that there is sufficient
capacity for the new element, and it misbehaves otherwise, breaking
invariants or even segfaulting. The necessary conditions could be
checked, but if we're to keep it lean, it should be `unsafe`, so the
burden is on the caller to ensure capacity.

Fixes #253.

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -872,19 +872,17 @@ impl<T, A: Allocator + Clone> RawTable<T, A> {
     /// This does not check if the given element already exists in the table.
     #[cfg_attr(feature = "inline-more", inline)]
     #[cfg(any(feature = "raw", feature = "rustc-internal-api"))]
-    pub fn insert_no_grow(&mut self, hash: u64, value: T) -> Bucket<T> {
-        unsafe {
-            let (index, old_ctrl) = self.table.prepare_insert_slot(hash);
-            let bucket = self.table.bucket(index);
+    pub unsafe fn insert_no_grow(&mut self, hash: u64, value: T) -> Bucket<T> {
+        let (index, old_ctrl) = self.table.prepare_insert_slot(hash);
+        let bucket = self.table.bucket(index);
 
-            // If we are replacing a DELETED entry then we don't need to update
-            // the load counter.
-            self.table.growth_left -= special_is_empty(old_ctrl) as usize;
+        // If we are replacing a DELETED entry then we don't need to update
+        // the load counter.
+        self.table.growth_left -= special_is_empty(old_ctrl) as usize;
 
-            bucket.write(value);
-            self.table.items += 1;
-            bucket
-        }
+        bucket.write(value);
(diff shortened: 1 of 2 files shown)


src/raw/mod.rs
```diff
@@ -872,19 +872,17 @@ impl<T, A: Allocator + Clone> RawTable<T, A> {
...
+    pub unsafe fn insert_no_grow(&mut self, hash: u64, value: T) -> Bucket<T> {
+        let (index, old_ctrl) = self.table.prepare_insert_slot(hash);
+        let bucket = self.table.bucket(index);
 
-            // If we are replacing a DELETED entry then we don't need to update
-            // the load counter.
-            self.table.growth_left -= special_is_empty(old_ctrl) as usize;
+        // If we are replacing a DELETED entry then we don't need to update
+        // the load counter.
+        self.table.growth_left -= special_is_empty(old_ctrl) as usize;
```
Amanieu, line 881: Since this function is already unsafe, we could add an extra `assume_no_deleted` argument which assumes that the table is fresh and does not contain any `DELETED` entries? This would skip the check on `old_ctrl` and might improve performance in some cases.

However it might be exposing too much of the hash table internals. What do you think?

cuviper: So `assume_no_deleted` would still be a condition there, but one that you would hope is constant-propagated?

I think it's too much of a micro-optimization, especially for exposing internals. I ran some `indexmap` benchmarks that are heavy on `rebuild_hash_table`, which is a `clear` followed by a loop of `insert_no_grow`. About 50% of that function's perf profile is simply the `bucket.write`, mostly due to cache misses. Most of the remaining time is spread around finding the insert slot, and less than 4% on updating `growth_left`. I even tried hard-coding that to `growth_left -= 1`, but the difference was lost in the noise.

PR #261, opened 2021-05-04
Upgrade to rand 0.8.3

PsiACE:
Just upgrade rand to 0.8.3.

Signed-off-by: Chojan Shang <psiace@outlook.com>


PsiACE:
Because different platforms have different requirements for from_seed, use seed_from_u64 instead

PsiACE:
cc @Amanieu

PR #267, opened 2021-05-22
Introduce more try methods to the HashMap

Kerollmops:
Hello,

I am working with hashbrown and bumpalo, trying to implement a text indexer using a HashMap combined with a custom allocator, the idea is to create a custom hashmap that is memory limited, that's a useful feature when you need to index a lot of documents and be sure to use the maximum memory you can and avoid going OOM and get killed.

The hashbrown API is quite good when it comes to avoiding hashing the keys, but there is real API when it comes to detecting potential OOM apart from the `try_reserve` one. Using the `try_reserve` method (with 1 as the parameter) before calling the `raw_entry_mut` method every time we need to look up into the map is not quite valid because the entry can already be occupied and we don't need to grow it, but maybe it's enough and it can work well.

I am aware that those three methods add a lot of complexity to the crate and I would completely agree with you if you tell me that you don't want to add them to the crate for this reason.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -2249,6 +2249,21 @@ impl<'a, K, V, S, A: Allocator + Clone> RawVacantEntryMut<'a, K, V, S, A> {
         self.insert_hashed_nocheck(hash, key, value)
     }
 
+    /// Try to set the value of the entry with the VacantEntry's key,
+    /// and returns a mutable reference to it.
+    ///
+    /// Returns the key and value if growing the map was mandatory and
+    /// the allocator returns an error.
+    #[cfg_attr(feature = "inline-more", inline)]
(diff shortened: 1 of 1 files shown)


Amanieu:
The performance impact of using `try_reserve` before `insert` is negligible. I would strongly encourage you to explore that option first.

There is an issue with the name of `try_insert`: libstd's `HashMap` already has a method with this name (https://doc.rust-lang.org/nightly/std/collections/hash_map/struct.HashMap.html#method.try_insert). It is unrelated to allocation failure, it simply returns an error if the key already exists.

Kerollmops:
Thank you for your time and reply, I will try that first and just close this PR.
Have a nice day!

PR #270, opened 2021-05-26
Replace Travis with Github actions

Amanieu:


.github/workflows/rust.yml
```diff
@@ -0,0 +1,80 @@
...
+      with:
+        toolchain: ${{ matrix.channel }}
+        override: true
+    - env:
+        TARGET: x86_64-pc-windows-msvc
+        CHANNEL: ${{ matrix.channel }}
+      run: sh ci/run.sh
+    strategy:
+      matrix:
+        channel: [1.49.0, stable, beta, nightly]
```
nagisa, line 80: Consider using matrix for the runner/runs-on variable too in order to reduce duplication. I find `strategy.matrix.include` and `strategy.matrix.exclude` to be useful in these situations: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-new-combinations

PR #276, opened 2021-07-01
adjustment to make clippy happy

striezel:
It seems clippy does not like `cloned()` when `copied()` can be used.
See <https://github.com/rust-lang/hashbrown/runs/2967101005> for
the details.


striezel:
And clippy is happy again: https://github.com/rust-lang/hashbrown/pull/276/checks?check_run_id=2967200663 :smiley_cat:

Amanieu:
Thanks!

PR #280, opened 2021-07-15
Inline `static_empty`

Zoxc:
This ended up not being inlined into `reserve_rehash`.


Amanieu:
I would have expected rustc's const-eval to always resolve this, but apparently not.

@bors r+

PR #282, opened 2021-07-20
Make rehashing and resizing less generic

Zoxc:
This makes the code in `rehash_in_place`, `resize` and `reserve_rehash` less generic on `T`. It also improves the performance of rustc. That performance increase in partially attributed to the use of `#[inline(always)]`.

This is the effect on rustc runtime:
```
clap:check                        1.9523s   1.9327s  -1.00%
hashmap-instances:check           0.0628s   0.0624s  -0.57%
helloworld:check                  0.0438s   0.0436s  -0.50%
hyper:check                       0.2987s   0.2970s  -0.59%
regex:check                       1.1497s   1.1402s  -0.82%
syn:check                         1.7004s   1.6851s  -0.90%
syntex_syntax:check               6.9232s   6.8546s  -0.99%
winapi:check                      8.3220s   8.2857s  -0.44%

Total                            20.4528s  20.3014s  -0.74%
Summary                           4.0000s   3.9709s  -0.73%
```
`rustc_driver`'s code size is increased by 0.02%.

This is the effect on compile time this has on my [HashMap compile time benchmark](https://github.com/rust-lang/hashbrown/pull/277#issuecomment-881033607):
```
hashmap-instances:check           0.0636s   0.0632s  -0.61%
hashmap-instances:release        33.0166s  32.2487s  -2.33%
hashmap-instances:debug           7.8677s   7.2012s  -8.47%

Total                            40.9479s  39.5131s  -3.50%
Summary                           1.5000s   1.4430s  -3.80%
```
The `hashmap-instances:debug` compile time could be further improved if there was a way to apply `#[inline(always)]` only on release builds.

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -3,7 +3,6 @@ use crate::scopeguard::guard;
 use crate::TryReserveError;
 #[cfg(feature = "nightly")]
 use crate::UnavailableMutError;
-use core::hint;
 use core::iter::FusedIterator;
 use core::marker::PhantomData;
 use core::mem;
@@ -11,6 +10,7 @@ use core::mem::ManuallyDrop;
 #[cfg(feature = "nightly")]
 use core::mem::MaybeUninit;
 use core::ptr::NonNull;
+use core::{hint, ptr};
 
 cfg_if! {
     // Use the SSE2 implementation if possible: it allows us to scan 16 buckets
@@ -359,6 +359,7 @@ impl<T> Bucket<T> {
     pub unsafe fn as_mut<'a>(&self) -> &'a mut T {
         &mut *self.as_ptr()
     }
+    #[cfg(feature = "raw")]
     #[cfg_attr(feature = "inline-more", inline)]
     pub unsafe fn copy_from_nonoverlapping(&self, other: &Self) {
         self.as_ptr().copy_from_nonoverlapping(other.as_ptr(), 1);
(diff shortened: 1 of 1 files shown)


src/raw/mod.rs
```diff
@@ -1312,6 +1224,19 @@ impl<A: Allocator + Clone> RawTableInner<A> {
         Bucket::from_base_index(self.data_end(), index)
     }
 
+    #[cfg_attr(feature = "inline-more", inline)]
+    unsafe fn bucket_ptr(&self, index: usize, size_of: usize) -> *mut u8 {
```
Amanieu, line 1228: This function absolutely needs to be `#[inline]` at minimum, it is not generic.

src/raw/mod.rs
```diff
@@ -708,76 +699,15 @@ impl<T, A: Allocator + Clone> RawTable<T, A> {
     /// allocation).
     ///
     /// If `hasher` panics then some the table's contents may be lost.
+    #[cfg(test)]
```
Amanieu, line 702: Move this function to the test module at the bottom of the file.

src/raw/mod.rs
```diff
@@ -1457,6 +1382,180 @@ impl<A: Allocator + Clone> RawTableInner<A> {
...
+        hasher: &dyn Fn(&mut Self, usize) -> u64,
+        size_of: usize,
+        drop: fn(*mut u8),
+        drops: bool,
+    ) {
+        // If the hash function panics then properly clean up any elements
+        // that we haven't rehashed yet. We unfortunately can't preserve the
+        // element since we lost their hash and have no way of recovering it
+        // without risking another panic.
+        self.prepare_rehash_in_place();
```
Amanieu, line 1490: There's no reason for `prepare_rehash_in_place` to be a separate function any more now that `rehash_in_place` is moved to `RawTableInner`.

src/raw/mod.rs
```diff
@@ -1457,6 +1382,180 @@ impl<A: Allocator + Clone> RawTableInner<A> {
...
+    /// code generated, but it is eliminated by LLVM optimizations when inlined.
+    #[allow(clippy::inline_always)]
+    #[inline(always)]
+    unsafe fn reserve_rehash_inner(
+        &mut self,
+        additional: usize,
+        hasher: &dyn Fn(&mut Self, usize) -> u64,
+        fallibility: Fallibility,
+        layout: TableLayout,
+        drop: fn(*mut u8),
```
Amanieu, line 1397: This should be an `Option` instead of having a separate `drops`.

src/raw/mod.rs
```diff
@@ -1312,6 +1224,19 @@ impl<A: Allocator + Clone> RawTableInner<A> {
         Bucket::from_base_index(self.data_end(), index)
     }
 
+    #[cfg_attr(feature = "inline-more", inline)]
+    unsafe fn bucket_ptr(&self, index: usize, size_of: usize) -> *mut u8 {
+        debug_assert_ne!(self.bucket_mask, 0);
+        debug_assert!(index < self.buckets());
+        let base: *mut u8 = self.data_end().as_ptr();
+        if size_of == 0 {
+            // FIXME: Check if this `data_end` is aligned with ZST?
```
Zoxc, line 1233: Just making a note to fix this.

src/raw/mod.rs
Zoxc: It is generic over `A`, so I figured this should match the inlining of `fn bucket`. Though I would put `#[inline(always)]` on both of them (on release mode).

Zoxc: There's a bunch of small functions which are beneficial to inline. I'll open a PR for those. It seems the compile time benefit of https://github.com/rust-lang/hashbrown/pull/119 mostly comes from removing inline on `get`, `insert`, etc.

src/raw/mod.rs
Zoxc: I find `rehash_in_place` to be sufficiently large and complex already, so I don't mind this being in a separate function for code clarity. I'm less sure about `prepare_resize` staying a separate function though.

PR #286, opened 2021-08-13
Apply s̶o̶m̶e  all clippy suggestions found in tests

steffahn:
Apply some clippy suggestions to tests, involving
* removing calls to `.clone()` on `&str`s,
* removing calls to `.clone()` on `const`s,
* a transformation `x = x * 2` →  `x *= 2`.

All of these seem like definite improvements to me.

In particular the `clone()`s on `&str` are a deny-by-default clippy lint ([`clippy::clone_double_ref`](https://rust-lang.github.io/rust-clippy/master/index.html#clone_double_ref)) and I don’t like the red marks in my IDE.

### DIFF
--- a/ci/tools.sh
+++ b/ci/tools.sh
@@ -30,7 +30,7 @@ if retry rustup component add rustfmt ; then
 fi
 
 if retry rustup component add clippy ; then
-    cargo clippy --all -- -D clippy::all -D clippy::pedantic
+    cargo clippy --all --tests -- -D clippy::all -D clippy::pedantic
 fi
 
 if command -v shellcheck ; then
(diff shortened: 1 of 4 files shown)


steffahn:
(The CI failure `rustfmt_clippy` seems unrelated to this PR, right?)

Amanieu:
Please update the CI script so that tests are also checked against clippy lints.

steffahn:
Rebasing after #287 was merged…

steffahn:
> Please update the CI script so that tests are also checked against clippy lints.

Let me see if I can figure out how that works…

steffahn:
Doing more commits than necessary here to double-check that the lints I’m seeing locally also turn up in PR…

steffahn:
@Amanieu Done now. I hope that just adding the `--tests` to the `cargo clippy` doesn’t somehow prevent any lints from happening. As long as there are no `cfg(not(test))` there shouldn’t be any problem though.

Amanieu:
Great work!

@bors r+

PR #288, opened 2021-08-13
Activate more features in clippy CI and fix new warnings

steffahn:
~~This PR is WIP,~~ the goal is to also fix the arising clippy warnings. Already opening a PR in order to test that the new CI command actually works.


steffahn:
CI should pass now

PR #291, opened 2021-09-01
Rename get_each_mut to get_many_mut and align API with the stdlib

shepmaster:
The standard library will be gaining a similar [ability for slices][]. This updates Hashbrown's API to match that PR and use only stable `MaybeUninit` functions.

This change requires a MSRV bump to Rust 1.51.

[ability for slices]: https://github.com/rust-lang/rust/pull/83608

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1,7 +1,5 @@
 use crate::raw::{Allocator, Bucket, Global, RawDrain, RawIntoIter, RawIter, RawTable};
 use crate::TryReserveError;
-#[cfg(feature = "nightly")]
-use crate::UnavailableMutError;
 use core::borrow::Borrow;
 use core::fmt::{self, Debug};
 use core::hash::{BuildHasher, Hash};
@@ -1172,18 +1170,14 @@ where
 
     /// Attempts to get mutable references to `N` values in the map at once.
     ///
-    /// Returns an array of length `N` with the results of each query. For soundness,
-    /// at most one mutable reference will be returned to any value. An
-    /// `Err(UnavailableMutError::Duplicate(i))` in the returned array indicates that a suitable
-    /// key-value pair exists, but a mutable reference to the value already occurs at index `i` in
-    /// the returned array.
-    ///
-    /// This method is available only if the `nightly` feature is enabled.
+    /// Returns an array of length `N` with the results of each query. For soundness, at most one
+    /// mutable reference will be returned to any value. `None` will be returned if any of the
+    /// keys are duplicates or missing.
     ///
     /// # Examples
     ///
     /// ```
-    /// use hashbrown::{HashMap, UnavailableMutError};
+    /// use hashbrown::HashMap;
     ///
     /// let mut libraries = HashMap::new();
     /// libraries.insert("Bodleian Library".to_string(), 1602);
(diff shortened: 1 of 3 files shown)


src/map.rs
```diff
@@ -1103,109 +1101,140 @@ where
...
     where
         K: Borrow<Q>,
         Q: Hash + Eq,
     {
-        self.get_each_inner_mut(ks)
-            .map(|res| res.map(|(k, v)| (&*k, v)))
+        self.get_many_inner_mut(ks).map(|res| {
+            // TODO use `array::map` here instead once that's stable.
+            unsafe {
+                use core::mem::MaybeUninit;
```
Amanieu, line 1221: Move this import to the top of the file.

Amanieu (commented):
Could you add `get_many_unchecked_mut` as well?

Also bump the Rust version in CI to 1.51 so the tests pass.

shepmaster:
> Could you add `get_many_unchecked_mut` as well?

Done

> Also bump the Rust version in CI to 1.51 so the tests pass.

Done

---

I also changed the implementation of the raw `get_many_mut` to avoid looping twice, which should be a bit more performant (but I did not test that).

src/map.rs
```diff
@@ -1103,109 +1101,174 @@ where
...
+    /// let got = libraries.get_many_key_value_mut([
+    ///     "Bodleian Library",
+    ///     "Herzogin-Anna-Amalia-Bibliothek",
+    ///     "Herzogin-Anna-Amalia-Bibliothek",
+    /// ]);
+    /// assert_eq!(got, None);
     /// ```
-    #[cfg(feature = "nightly")]
-    pub fn get_each_key_value_mut<Q: ?Sized, const N: usize>(
+    pub fn get_many_key_value_mut<Q: ?Sized, const N: usize>(
```
Amanieu, line 1255: Can you add an unchecked version of this as well?

src/raw/mod.rs
```diff
@@ -855,60 +852,49 @@ impl<T, A: Allocator + Clone> RawTable<T, A> {
...
     /// The `eq` argument should be a closure such that `eq(i, k)` returns true if `k` is equal to
     /// the `i`th key to be looked up.
-    ///
-    /// This method is available only if the `nightly` feature is enabled.
-    #[cfg(feature = "nightly")]
-    pub fn get_each_mut<const N: usize>(
+    pub fn get_many_mut<const N: usize>(
         &mut self,
         hashes: [u64; N],
+        validate_hashes_are_unique: bool,
```
Amanieu, line 867: This needs to be an unsafe function if validation can be disabled. I would prefer if there was a separate `unchecked` function which is unsafe.

shepmaster: Updated

src/map.rs
shepmaster: Added

src/raw/mod.rs
```diff
@@ -855,60 +852,59 @@ impl<T, A: Allocator + Clone> RawTable<T, A> {
...
     ///
-    /// Returns an array of length `N` with the results of each query. For soundness,
-    /// at most one mutable reference will be returned to any entry. An
-    /// `Err(UnavailableMutError::Duplicate(i))` in the returned array indicates that a suitable
-    /// entry exists, but a mutable reference to it already occurs at index `i` in the returned
-    /// array.
+    /// Returns an array of length `N` with the results of each query.
+    ///
+    /// At most one mutable reference will be returned to any entry. `None` will be returned if any
+    /// of the hashes are duplicates. `None` will be returned if the hash is not found.
```
Amanieu, line 858:
```suggestion
    /// At most one mutable reference will be returned to any entry. `None` will be returned if any
    /// of the entries are duplicates. `None` will be returned if an entry is not found.
```

Amanieu:
LGTM, I was waiting for a reply on https://github.com/rust-lang/rust/pull/83608 about returning a `Result` instead of an `Option` but I guess we can always change this later if needed.

Amanieu:
Sorry, this slipped off my review queue. Could you rebase so I can merge this? The only conflict should be the CI version change.

shepmaster:
Rebased and now using `array::map` as it has stabilized since the original PR was opened.

PR #294, opened 2021-09-20
Add try_entry to HashMap.

mwkmwkmwk:
This closes a minor feature hole in the HashMap API. Suppose you want to:

1. Given only a borrowed key, look up a map entry.
2. Based on the entry's value, decide to either mutate or remove it.

Right now this isn't quite possible.  There are two alternatives:

1. First do `get_mut` to examine and perhaps mutate the value, then do `remove` if the item should be removed.  This results in a double lookup on the map in the "remove" case.
2. Clone the key into an owned key, then use the `entry` API.  This results in unnecessary memory allocation.

This PR closes the hole by adding a `try_entry` function that is similar to `entry`, but returns `None` instead of `VacantEntry` and only needs a borrowed key.  The returned `OccupiedEntry` also has the same limitation of not owning a key as an `OccuppiedEntry` created via `VacantEntry::insert`, and so cannot be used with some APIs (that are quite obscure in the first place).

---

I'd actually like to propose the same method for `std::collections::HashMap`, but I'm not quite sure how to go about it, procedurally — an equivalent PR cannot be made against rust-lang/rust, as the `HashMap` type is effectively outsourced here.  Help?

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -942,6 +942,68 @@ where
         }
     }
 
+    /// Gets the given key's corresponding entry in the map for in-place manipulation,
+    /// if it exists.
+    ///
+    /// The key may be any borrowed form of the map's key type, but
+    /// [`Hash`] and [`Eq`] on the borrowed form *must* match those for
+    /// the key type.
+    ///
+    /// This is a version of `entry` that takes a borrowed key, but cannot be used to insert
+    /// a new entry, returning `None` when the key is not present in the map.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use hashbrown::HashMap;
+    ///
+    /// let mut fruits = HashMap::new();
+    /// fruits.insert("apple".to_string(), 1);
+    /// fruits.insert("orange".to_string(), 3);
+    /// fruits.insert("pear".to_string(), 2);
+    ///
+    /// // Eats a fruit, returns true if succeeded.
+    /// let mut eat_one = |k| {
+    ///     if let Some(mut e) = fruits.try_entry(k) {
+    ///         let num = e.get_mut();
+    ///         *num -= 1;
+    ///         if *num == 0 {
+    ///             e.remove();
+    ///         }
+    ///         true
+    ///     } else {
+    ///         false
+    ///     }
+    /// };
+    ///
+    /// assert!(eat_one("apple"));
+    /// assert!(eat_one("orange"));
+    /// assert!(!eat_one("plum"));
+    ///
+    /// assert_eq!(fruits.get("apple"), None);
+    /// assert_eq!(fruits["orange"], 2);
+    /// assert_eq!(fruits["pear"], 2);
+    /// assert_eq!(fruits.get("plum"), None);
+    /// ```
+    #[cfg_attr(feature = "inline-more", inline)]
+    pub fn try_entry<Q: ?Sized>(&mut self, k: &Q) -> Option<OccupiedEntry<'_, K, V, S, A>>
+    where
+        K: Borrow<Q>,
+        Q: Hash + Eq,
+    {
+        let hash = make_hash::<K, Q, S>(&self.hash_builder, k);
+        if let Some(elem) = self.table.find(hash, equivalent_key(k)) {
+            Some(OccupiedEntry {
+                hash,
+                key: None,
+                elem,
+                table: self,
+            })
+        } else {
+            None
+        }
+    }
+
     /// Returns a reference to the value corresponding to the key.
     ///
     /// The key may be any borrowed form of the map's key type, but
@@ -3171,7 +3233,8 @@ impl<'a, K, V, S, A: Allocator + Clone> OccupiedEntry<'a, K, V, S, A> {
     ///
     /// # Panics
     ///
-    /// Will panic if this OccupiedEntry was created through [`Entry::insert`].
+    /// Will panic if this OccupiedEntry was created through [`Entry::insert`] or
+    /// [`HashMap::get_entry`].
     ///
     /// # Examples
     ///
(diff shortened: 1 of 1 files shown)


SafariMonkey:
I'm an onlooker in this space, so take this with a grain of salt. That said, seems worth mentioning that the [`raw_entry`](https://docs.rs/hashbrown/0.11.2/hashbrown/hash_map/struct.HashMap.html#method.raw_entry_mut) API is the one that `std` had been considering stabilising for this purpose, but apparently they are now more interested in a simpler API, in my eyes a little more like this one. To quote:

> We'd like to see an implementation of `entry_ref` which allows an `Entry` to be created from just a reference to a key, which would be held in a `Cow` internally and upgraded to an owned key when necessary for an insertion.
> 
> We feel that any more advanced use cases are better off using `hashbrown`'s `RawTable` API directly which is much more flexible than the `HashMap` API.

(source: https://github.com/rust-lang/rust/issues/56167#issuecomment-910742027)

This is not to comment on the merits of this particular proposal - I really don't have the expertise to do so - but to give a pointer towards where these discussions are happening (in the context of `std`, at least).

ZoopOTheGoop:
I was looking into this myself, but for `HashMap`. The reason I don't like this is it's not really an improvement in the number of times you need to hash if `key` is missing. This is basically the same case as:

```rust
if let Some(location) = map.get_mut(&my_key) {
    location
} else {
    map.entry.or_insert(my_key.clone())
}
```

Which will re-look up the key.

This proposal has a couple problems I see, one is that it's harder to use `if let...else`, which is what a lot of people would intuitively do

```rust
if let Some(entry) = map.try_entry(&my_key) {
    entry
} else {
    map.entry(my_key.clone())
} // Type error: first branch has type `OccupiedEntry` second has type `Entry`
```

This could be fixed by changing it to `Entry`, whose branch happens to always be `OccupiedEntry`.

I'm not 100% sure the borrow checker will allow this, which is another intuitive way to handle this (assuming it's changed to `Entry`), I'd have to check:

```rust
map.try_entry(&my_key).or_else(|| map.entry(my_key)).insert()
```

Either way, it hashes and looks up `my_key` twice on failure.

Personally the way I had two different ways of doing it in mind:

1. An `EntryRef` (based on the name in https://github.com/rust-lang/rust/issues/56167#issuecomment-910742027) which is created from an `entry_ref(&mut self, key: Q)`, this has two methods: `to_entry(&mut self) -> Entry`, `to_entry_with(&mut, with: FnOnce()->K) -> Entry`. The first is only implemented for types where `Q: ToOwned<K>`
2. Directly putting these on `XMap`, where `fn entry_ref(&mut self, key: Q) -> Entry` will do either a lookup OR `to_owned` automatically, and `fn entry_ref_with(&mut self, key: Q, with: FnOnce(Q) -> K) -> Entry`, which will call the function to provide a key as needed.

Making it return an `Entry` I think it best, in principle, because it allows people to reuse familiar code without API surface explosion, in particular different functions for different purposes could call `entry` or `entry_ref`/`_with` as needed and return the resulting `Entry` and use the same method chain on them.

... in *theory*. In practice, at least for `HashMap` this is much easier said than done. The problem is `OccupiedEntry` wants to own `K`, which means we cannot simply construct an `OccupiedEntry` off a borrow of it as it stands.

Given this, I only really see a couple potential proposals one could make:

1. Make an `EntryRef` enum/struct and API, probably using `RawEntryMut` internally, that essentially duplicates `Entry` altogether, possibly find a way to share some code with some internal refactoring.
    - This seems like a bad solution just because of the maintenance, effort, and confusing API surface area, not to mention it doesn't allow reusing stuff that takes `Entry` unless we add some traits too, which is more work for something that's almost implemented already. (That said an `Entry` trait could be a good add someday to allow people to abstract over collections in some way, but I don't think it should be needed here).
2. Do one of my two options above, accepting this fact and just stating that `key` will be cloned no matter what. 
    - This defeats the purpose entirely, IMO.
3. Add a third enum member to `Entry` (and probably therefore `RustcEntry` in `hashbrown`) which is functionally identical to `OccupiedEntry`, but for a borrowed Q.
    - Unfortunately, due to the decision to expose `Entry`'s type internals, this is also a potentially breaking change if anything is directly doing a `match` or `if let` on `Entry`. I'm not sure why you would, given the API as it stands, but it would be an issue. 
    - Not even something like `#[non_exhaustive]` can get us past this, since it forces matches to add a `_`.
4. Retrofit `OccupiedEntry` (and probably upstream into `hashbrown::Entry` and such in `HashBrown`) to allow for `Borrow<K>`, or make that field `Option<K>`, or similar.
    - *Maybe* we could do this, if we use sufficient care, given `Borrow<T>` is implemented for any `T`, sized or not (i.e. everything is its own Borrow), and neither `OccupiedEntry` nor `Entry` put any type constraints on `K`. Accepting `Borrow<K>` instead of just `K` should *only* widen the API, not close it off, and the public-facing method `entry` itself would still be under `K`.
    - I'm not 100% sure about this, however.
    - Just from a quick scan of the code, there seem to be maybe one or two places, mostly internal, that rely on `OccupiedEntry`'s key field being owned, I think most of them can be worked around either by simply using the key already available in `entry`, or deferring to a helper method (from a simple internal trait maybe?) that flips its behavior depending on whether it's actually owned or not.
    - This could probably be aided with `unsafe` too but it's best to avoid that.

That said, doing almost any of this requires doing at least part of the work in `hashbrown` *first*, so we'd need to coordinate putting what is a relatively useless change in `hashbrown` (since they have the `raw_entry` API already, which is what the `entry_ref` idea is replacing, none of the changes will be directly useful in that crate itself), and then sync that work after that's done.

IDK, that's just my analysis after a few hours of poking at this and looking at this implementation, thoughts?

Amanieu:
I think that `EntryRef` added in #301 solves this use case.

Issue #297, opened 2021-10-29
Implement From<[T; N]> for HashSet<T>


4LT:
Hashbrown's HashSet does not implement the `From` trait for arrays, while `std::collections::HashSet` (as of 1.56) does.

As such, HashSet is currently not a drop-in replacement.

Amanieu:
I would be happy to accept a PR for this if you are interested in contributing!

4LT:
Yeah, I might as well take a crack at it

4LT:
Closing because #298  was merged

PR #298, opened 2021-11-05
Issue #297: Implement From<array> on HashSet and HashMap

4LT:
Implemented From<[T; N]> on HashSet and From<[(K, V); N]> on HashMap.  In both cases, the default hasher was used to match the implementations from the std crate (see https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#1161-1190)

Edition was updated to 2021 to take advantage of the change in array's `.into_iter` method from iterating over `&T` to iterating over `T`.

Hopefully the implementations are trivial enough that tests aren't necessary ;)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -60,10 +60,10 @@ jobs:
           channel: nightly
         - os: macos-latest
           target: x86_64-apple-darwin
-          channel: 1.49.0
+          channel: 1.56.1
         - os: windows-latest
           target: x86_64-pc-windows-msvc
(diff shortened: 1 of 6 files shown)


Amanieu:
Since we're due for major version bump, could you update the CI and README to point to the latest stable instead of 1.49?

4LT:
Okay, I've done that, but now builds are failing with `--no-default-features` set b/c I used the concrete `DefaultHashBuilder` in my implementations, but that doesn't use the traits `BuildHasher` and `Default` when default features are not used.

PR #316, opened 2022-03-17
mask prepare_resize inline attribute with inline-more feature

### LINKED ISSUE #315: hide inline attribute for prepare_resize behind inline_more
The generated code size for prepare_resize is kinda large, consider mask it behind inline_more feature?

```
__RNvMs9_NtCs4rqddLBFH3d_9hashbrown3rawINtB5_13RawTableInnerNtNtCs9zLCrCdkmxP_5alloc5alloc6GlobalE14prepare_resizeCs3Uva5kdYy2w_9(crate_large) 275
__RNvMs9_NtCs4rqddLBFH3d_9hashbrown3rawINtB5_13RawTableInnerNtNtCs9zLCrCdkmxP_5alloc5alloc6GlobalE14prepare_resizeCs17qWP1ufcKj_13(crate_small) 115
```

shuoli84:
fix #315

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -1346,7 +1346,7 @@ impl<A: Allocator + Clone> RawTableInner<A> {
     }
 
     #[allow(clippy::mut_mut)]
-    #[cfg_attr(feature = "inline-more", inline)]
+    #[inline]
     unsafe fn prepare_resize(
         &self,
         table_layout: TableLayout,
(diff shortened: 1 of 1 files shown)


Amanieu:
I don't think `prepare_resize` is the issue here since it is only called from one place (`resize_inner`). Perhaps the `#[inline(always)]` should be removed from `resize_inner` instead?

shuoli84:
> I don't think `prepare_resize` is the issue here since it is only called from one place (`resize_inner`). Perhaps the `#[inline(always)]` should be removed from `resize_inner` instead?

Yes, I think so, resize_inner is the better place to do the modify. Thx.

PR #321, opened 2022-04-27
Correct the implementation of Debug for ValuesMut and IntoValues structures

JustForFun88:
Previously the implementation of Debug trait for ValuesMut was
```
impl<K, V> fmt::Debug for ValuesMut<'_, K, V>
where
    K: fmt::Debug,
    V: fmt::Debug,
{
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.debug_list().entries(self.inner.iter()).finish()
    }
}
```
It is strange, because `self.inner.iter()` return an iterator with element of type `(&’a K, &’a V)`.

The same is true when we look at the old implementation of Debug trait for the `IntoValues` structure. Here we have mistake
```
impl<K: Debug, V: Debug, A: Allocator + Clone> fmt::Debug for IntoValues<K, V, A> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.debug_list()
            .entries(self.inner.iter().map(|(k, _)| k))
            .finish()
    }
}
```
because with the function `self.inner.iter().map(|(k, _)| k)` we return iterator with element of type 'a K.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -3151,7 +3151,9 @@ impl<K, V> FusedIterator for ValuesMut<'_, K, V> {}
 
 impl<K, V: Debug> fmt::Debug for ValuesMut<'_, K, V> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_list().entries(self.inner.iter().map(|(_, val)| val)).finish()
+        f.debug_list()
+            .entries(self.inner.iter().map(|(_, val)| val))
+            .finish()
     }
 }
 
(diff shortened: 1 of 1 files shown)


Amanieu:
Can you run `cargo fmt` to fix the formatting?

Amanieu:
CI failure should be fixed by #325

JustForFun88:
> Can you run `cargo fmt` to fix the formatting?

Sorry, forgot about it

PR #328, opened 2022-05-08
Improving documentation.

JustForFun88:
Improved hashbrown documentation. Changed some examples so that they become full-fledged tests.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -995,7 +995,6 @@ where
     ///
     /// ```
     /// use hashbrown::HashMap;
-    /// use hashbrown::TryReserveError;
     ///
     /// let mut map: HashMap<&str, isize> = HashMap::new();
     /// // Map is empty and doesn't allocate memory
@@ -1005,7 +1004,13 @@ where
     ///
     /// // And now map can hold at least 10 elements
     /// assert!(map.capacity() >= 10);
-    ///
+    /// ```
+    /// If the capacity overflows, or the allocator reports a failure, then an error
+    /// is returned:
+    /// ```
+    /// # fn test() {
+    /// use hashbrown::HashMap;
+    /// use hashbrown::TryReserveError;
     /// let mut map: HashMap<i32, i32> = HashMap::new();
     ///
     /// match map.try_reserve(usize::MAX) {
@@ -1016,13 +1021,18 @@ where
     ///     _ => panic!(),
     /// }
     ///
-    /// match map.try_reserve(usize::MAX / 32) {
+    /// match map.try_reserve(usize::MAX / 64) {
     ///     Err(error) => match error {
     ///         TryReserveError::AllocError { .. } => {}
     ///         _ => panic!("TryReserveError::CapacityOverflow ?"),
     ///     },
     ///     _ => panic!(),
     /// }
+    /// # }
+    /// # fn main() {
+    /// #     #[cfg(not(miri))]
+    /// #     test()
+    /// # }
     /// ```
     #[cfg_attr(feature = "inline-more", inline)]
     pub fn try_reserve(&mut self, additional: usize) -> Result<(), TryReserveError> {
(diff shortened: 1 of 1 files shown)


JustForFun88:
I really don't know how change example that must return error

src/map.rs
```diff
@@ -766,6 +849,8 @@ impl<K, V, S, A: Allocator + Clone> HashMap<K, V, S, A> {
     ///
     /// assert_eq!(evens, vec![0, 2, 4, 6]);
     /// assert_eq!(odds, vec![1, 3, 5, 7]);
+    /// // Map capacity is equal to old one.
+    /// assert!(map.capacity() == capacity_before_drain_filter);
```
Amanieu, line 853:
```suggestion
    /// assert_eq!(map.capacity(), capacity_before_drain_filter);
```

src/map.rs
```diff
@@ -791,8 +876,14 @@ impl<K, V, S, A: Allocator + Clone> HashMap<K, V, S, A> {
...
     /// let mut a = HashMap::new();
     /// a.insert(1, "a");
+    /// let capacity_before_clear = a.capacity();
+    ///
     /// a.clear();
+    ///
+    /// // Map is empty.
     /// assert!(a.is_empty());
+    /// // But map capacity is equal to old one.
+    /// assert!(a.capacity() == capacity_before_clear);
```
Amanieu, line 886:
```suggestion
    /// assert_eq!(a.capacity(), capacity_before_clear);
```

src/map.rs
```diff
@@ -888,8 +995,44 @@ where
...
+    ///     _ => panic!(),
+    /// }
+    ///
+    /// match map.try_reserve(usize::MAX / 64) {
+    ///     Err(error) => match error {
+    ///         TryReserveError::AllocError { .. } => {}
+    ///         _ => panic!("TryReserveError::CapacityOverflow ?"),
+    ///     },
+    ///     _ => panic!(),
+    /// }
```
Amanieu, line 1030: I would just remove this, it's very difficult to get it to trigger reliably and can cause OOM errors in other processes on some systems.

Amanieu (commented):
LGTM! One thing I noticed is that we are somewhat inconsistent in the doc headers: sometimes we use `Examples` and other times we use `Example`. It would be nice to consistently use one or the other.

Also, do you have any plans to submit these doc changes upstream to rust-lang/rust so that they appear in the standard library docs?

JustForFun88:
> LGTM! One thing I noticed is that we are somewhat inconsistent in the doc headers: sometimes we use `Examples` and other times we use `Example`. It would be nice to consistently use one or the other.
> 
> Also, do you have any plans to submit these doc changes upstream to rust-lang/rust so that they appear in the standard library docs?

I think so, but it will take time. I will need to parse the existing `hashbrown` and `std::collections::HashMap` documentation.
I'll try to finish within a week.

src/map.rs
JustForFun88: But if I delete, then there are will not be shown all variants of the error.  This is just an example, no one will actually try to allocate 16 million terabytes of memory in a real program :-)

JustForFun88: Can I just write a comment that running this part of the example could cause OOM errors in other processes on some systems?

JustForFun88:
Ready! Update as per your request. I also looked at the existing documentation for `std::collections::HashMap` and tried to get the two documentations as close to each other as possible (without considering of course my own changes).

> One thing I noticed is that we are somewhat inconsistent in the doc headers: sometimes we use `Examples` and other times we use `Example`. It would be nice to consistently use one or the other.

I'll bring this up in the next pull request #331.

src/map.rs
Amanieu: The examples are tested by CI, so this would cause CI tests to fail. Maybe you can mark the example as ````no_run` so that it isn't executed when running tests.

Issue #330, opened 2022-05-08
API for removing by key hash and value


qrnch-jan:
Would it be possible to add a function that allows an entry to be remove based on a key hash and a closure, sort of like how `RawEntryBuilder::from_hash()` works?

I'm having some trouble with a piece of code that uses a stored key hash to manipulate a `HashMap`.  Simplified, I need to remove an entry from a `HashMap`, but I do not have access to the entry's key.  However, I do have access to its hash and its value (which is uniquely identifiable).  Something like this would be _perfect_:

```
map.remove_by_hash(key_hash, |_k, v| {
   *v == the_value_i_am_looking_for
});
```

I run into trouble if I try to translate the key hash to a key reference by via `raw_entry()`, then `from_hash()`, and finally use the regular `remove()` method because the returned reference to the key will disallow mutable access to the `HashMap`.

Amanieu:
Since you're using `HashMap` in non-standard ways, have you considered just using `RawTable` directly? It provides a much better interface for dealing with custom hashmaps.

I'm hesitant to add yet another extension to the `HashMap` API since it is already quite bloated. Also the libs team consensus on `RawEntry` is that it's unlikely to ever be stabilized and may be removed in the future.

qrnch-jan:
The vast majority of what I do is plain `HashMap` stuff.  Using the `RawTable` would mean I would need to duplicate a lot of standard functionality, which I'd like to avoid.

I would prefer to make a convincing argument for accepting my pull request, but I'd be lying if I said I didn't understand the argument for removing `RawEntry`..

qrnch-jan:
Update: I'm going the `RawTable` as suggested, no need for the `remove_by_hash()` method.

PR #335, opened 2022-05-15
Adding function for getting access to map `table: RawTable<(K, V), A>` field

JustForFun88:
I was a little interested in issues "API for removing by key hash and value" #330. The original question was about adding a fancy API, and of course it wasn't merged (#334).

However, it seems to me that this shows that it would be nice to just add the ability to access to map's fields. And let the user himself do what he pleases.

The user will still not be able to change the existing API. But at the same time, he will have the opportunity to add small extensions. Otherwise, when using `RawTable`, he will have to not only add one or two functions, but generally recreate all the `HashMap`’s APIs, which can be difficult.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,6 @@ rustc-dep-of-std = [
     "rustc-internal-api",
 ]
 raw = []
-map-inner = ["raw"]
 
 # Enables usage of `#[inline]` on far more functions than by default in this
 # crate. This may lead to a performance increase but often comes at a compile
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ rustc-dep-of-std = [
     "rustc-internal-api",
 ]
 raw = []
+map-inner = ["raw"]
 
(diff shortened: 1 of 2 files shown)


JustForFun88:
What's wrong with this test? I'm getting through everything. And it has nothing to do with my change at all.

JustForFun88:
I understood. It's not my fault. It is [rust-lang/rust#96150](https://github.com/rust-lang/rust/pull/96150).

Amanieu:
Could you expose these using functions instead? The hash builder is already exposed via the `hasher()` method, so you'd just need to add an unsafe `raw_table(_mut)` method which exposes the underlying raw table. I don't think that a separate feature is needed, you can include it under the existing `raw` feature.

JustForFun88:
> Could you expose these using functions instead? The hash builder is already exposed via the `hasher()` method, so you'd just need to add an unsafe `raw_table(_mut)` method which exposes the underlying raw table. I don't think that a separate feature is needed, you can include it under the existing `raw` feature.

Ready. Fixed as requested.

PR #336, opened 2022-05-15
 Allow unused macro rules for macro `cfg_if`.

JustForFun88:
The unused macro rules lint is an new lint for the rust compiler: [rust-lang/rust#96150](https://github.com/rust-lang/rust/pull/96150)


JustForFun88:
Network failure :-)

PR #342, opened 2022-06-16
Add an Entry API for HashSet

### LINKED ISSUE #274: Entry API for HashSet
It would in some cases be useful to have an equivalent of the entry API available for `HashSet`. The case I ran into was wanting to remove values from two sets only if both sets contain the values. With the current API, this would be something like:

```rust
if set1.contains(&value1) && set2.remove(&value2) {
    set1.remove(&value1);
}
```

which does two lookups in `set1` in the case that the values should be removed: as far as I can see there is no way to avoid doing the duplicate lookup. A similar situation would arise in the case where values should only be inserted if neither set contains the values.

It is possible to emulate this with a `HashMap<T, ()>`, e.g.

```rust
if let Entry::Occupied(entry) = map1.entry(value1) {
    if map2.remove(&value2).is_some() {
        entry.remove();
    }
}
```

but would be a nice-to-have for `HashSet` as well.

cuviper:
This adds `HashSet::entry` and `rustc_entry`, mimicking the API of `HashMap`. The "rustc" variants are without the `S` type parameter, just like the difference on maps. Set entries have fewer methods since there's no mutable value like in maps, but the basic insert/remove/replace functionality is there.

Resolves #274.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -5573,7 +5573,7 @@ impl<'a, K, V, S, A: Allocator + Clone> VacantEntry<'a, K, V, S, A> {
     }
 
     #[cfg_attr(feature = "inline-more", inline)]
-    fn insert_entry(self, value: V) -> OccupiedEntry<'a, K, V, S, A>
+    pub(crate) fn insert_entry(self, value: V) -> OccupiedEntry<'a, K, V, S, A>
     where
         K: Hash,
         S: BuildHasher,
--- a/src/rustc_entry.rs
+++ b/src/rustc_entry.rs
@@ -56,10 +56,10 @@ where
 
 /// A view into a single entry in a map, which may either be vacant or occupied.
 ///
-/// This `enum` is constructed from the [`entry`] method on [`HashMap`].
+/// This `enum` is constructed from the [`rustc_entry`] method on [`HashMap`].
 ///
 /// [`HashMap`]: struct.HashMap.html
-/// [`entry`]: struct.HashMap.html#method.rustc_entry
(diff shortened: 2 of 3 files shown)


Amanieu:
Could we avoid introducing another `rustc_entry`? This was a hack for compatibility with `std` which didn't have a `S` parameter on the `Entry` type. I would prefer if we did this "correctly" and had a `S` parameter for the version used by `std`.

cuviper:
I expect that `std::collections::hash_set::Entry<T>` should look like a simple wrapper on `hash_map::Entry<T, ()>`, even if that's not really direct because it goes through hashbrown. I think it would be unfortunate if `std` were inconsistent about whether these `Entry` types have an `S` parameter, but that's ultimately a libs-api decision.

If you still want to push for that `S`, then I'll drop `rustc_entry` here and keep it in a branch on the side, just in case. :)

Amanieu:
I'll repeat that `Entry` not having `S` was a mistake that we shouldn't carry over to new APIs. The `rustc_entry` in the hashbrown crate is actually slightly less efficient because it has to reserve space for a value ahead of time before knowing whether a value will be inserted into the map. It is better to defer the reservation until a value is explicitly inserted, which requires the `S` parameter.

cuviper:
OK, I've dropped the `S`-less one.

SUPERCILEX:
Any idea when this will be released?

Amanieu:
I just published a release.

SUPERCILEX:
Thanks!

Issue #345, opened 2022-07-12
Lookup value without `Borrow` implementation


timothee-haudebourg:
I have a recurring problem with this library (and the standard library) on the signature of lookup methods such as `HashMap::get(&self, key: &Q)` that require the key type `K` to implement `K: Borrow<Q>`. I came across many cases where this is a limitation. For instance consider a map where the key type is as follows:
```rust
enum Key {
  String(String)
  Number(i32)
}
```
This type owns the data. But if you want to build a string key by only borrowing the string (which itself may not be in a `Key`), a nice solution is to define a `KeyRef` type as follows
```rust
enum KeyRef<'a> {
  String(&'a str)
  Number(i32)
}
```
Now because the string is wrapped inside an `enum` type, there is no way to have `Key: Borrow<KeyRef>`, even though both types are semantically very close, can have the same hash function, have a straight forward comparison function, etc. But because we can't implement `Key: Borrow<KeyRef>`, we can't use `KeyRef` as a key in `HashMap::get`, which is very annoying. The only solution in this case is to clone the data from `KeyRef` to build a `Key` and then use `HashMap::get`.

There are some successful hash map implementations out there that have solved this issue already. For instance, the [`indexmap`](https://crates.io/crates/indexmap) library introduces an [`Equivalent`](https://docs.rs/indexmap/latest/indexmap/trait.Equivalent.html) trait that replaces the `Borrow` bound and instead requires `Q: Equivalent<K>`. The only provided implementation is based on `Borrow`:
```rust
impl<Q: ?Sized, K: ?Sized> Equivalent<K> for Q where
    Q: Eq,
    K: Borrow<Q>
```
This makes this little difference transparent when the `K: Borrow<Q>` implementation exists, but also allows for more complicated scenarios to work by adding an `Equivalent` implementation.

I know introducing a new trait is sometimes not wanted, but in this case it could solve a recurring problem, that I'm sure have be encountered by many people, justifying its introduction in the `indexmap` library.

Amanieu:
I'd be happy to include support for `Equivalent` in hashbrown! Would you be interested in writing up a PR for this?

timothee-haudebourg:
The only issue is with the `EntryRef` API, which requires `K: Borrow<Q>` not only for comparison. I could adapt the API, in particular by exposing the `KeyOrRef` type, but not without major changes, so for now I'll just keep it as it is.

timothee-haudebourg:
[...]

PR #350, opened 2022-07-22
`Equivalent` trait

timothee-haudebourg:
This PR introduces an `Equivalent` trait, as requested in #345, to customize lookup operations without requiring a `Borrow` implementation. It provides a blanket implementation that covers the `Borrow` case so it should not break anything.

The `hash_map::EntryRef` API, although it uses the `Borrow` trait, remains unchanged as this would introduce some breaking changes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -117,6 +117,39 @@ pub mod hash_set {
 pub use crate::map::HashMap;
 pub use crate::set::HashSet;
 
+/// Key equivalence trait.
+///
+/// This trait defines the function used to compare the input value with the
+/// map keys (or set values) during a lookup operation sur as [`HashMap::get`]
+/// or [`HashSet::contains`].
+/// It is provided with a blanket implementation based on the
+/// [`Borrow`](core::borrow::Borrow) trait.
+///
+/// # Correctness
+///
+/// Equivalent values must hash to the same value.
+pub trait Equivalent<K: ?Sized> {
+    /// Checks if this value is equivalent to the given key.
+    ///
+    /// Returns `true` if oth values are equivalent, and `false` otherwise.
+    ///
+    /// # Correctness
+    ///
+    /// When this function returns `true`, both `self` and `key` must hash to
+    /// the same value.
+    fn equivalent(&self, key: &K) -> bool;
+}
+
+impl<Q: ?Sized, K: ?Sized> Equivalent<K> for Q
+where
+    Q: Eq,
+    K: core::borrow::Borrow<Q>,
+{
+    fn equivalent(&self, key: &K) -> bool {
+        self == key.borrow()
+    }
+}
+
 /// The error type for `try_reserve` methods.
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub enum TryReserveError {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -120,7 +120,7 @@ pub use crate::set::HashSet;
 /// Key equivalence trait.
 ///
 /// This trait defines the function used to compare the input value with the
-/// map keys (or set values) during a lookup operation sur as [`HashMap::get`]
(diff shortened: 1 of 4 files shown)


timothee-haudebourg:
I said that the `EntryRef` API is unchanged, but the `HashMap::entry_ref` method uses the `Equivalent` trait. Maybe it would be better to continue using the `Borrow` trait for this method, since the resulting `EntryRef` will probably be unusable without a `Borrow` implementation anyway.

Amanieu (approved):
LGTM! Let me know once you have fixed the typo so I can merge this.

Amanieu:
ping @timothee-haudebourg, is this ready to merge?

timothee-haudebourg:
Sorry I forgot about this, I just pushed some typo fixes, it should be ready to merge!

Amanieu:
Could you add a test for this? Perhaps just copy one from indexmap: https://github.com/bluss/indexmap/blob/master/tests/equivalent_trait.rs

Amanieu:
@bors r+

Thanks!

dhedey:
Just to give a heads up for anyone else hitting this problem.

When I upgraded `HashBrown` past 1.3.1 which released this feature, I got some compile errors on code I'd previously written. The error was:
```
the trait `Borrow<&X>` is not implemented for `X`
= note: required for `&X` to implement `hashbrown::Equivalent<X>`
```

It transpires that we previously had double references: `let x = &abc; hashmap.get(&x)` - and the code previously was happy with them, but the inference with the blanket impl of `Equivalent` isn't as clever as with the direct `Borrow` impl.

I could fix this in my code by removing the extra `&` (or equivalently, adding `*` before the reference).

----

Separately, from a hashbrown compatibility perspective, is this an issue? For us at least it means the build passed fine using `std::HashMap`, but broke using `hashbrown` as a stand in replacement when building for no-std.

mrcnski:
Thanks @dhedey, we hit the same error, and simply removing an extra `&` worked to resolve it!

mattklein123:
@Amanieu sorry for the late drive by on this PR: would you accept a PR which makes the default implementation of Equivalent gated behind a feature which defaults to on? I have a use case where I would like to implement Equivalent on top of a generic structure and I can't figure out a way to prevent this from conflicting with the default implementation. Thank you!

Amanieu:
Note that in the next release we will use the trait from the `equivalent` crate instead (https://github.com/rust-lang/hashbrown/pull/442). Would this help in your case?

Amanieu:
Unfortunately I don't think we can disable the `Equivalent` trait: it's needed by `HashMap` to support `Borrow`.

mattklein123:
> Note that in the next release we will use the trait from the equivalent crate instead (https://github.com/rust-lang/hashbrown/pull/442). Would this help in your case?

I'm not sure. I can take a look once it releases.

> Unfortunately I don't think we can disable the Equivalent trait: it's needed by HashMap to support Borrow.

I wasn't suggesting disabling the equivalent trait, I was thinking more that we could allow the blanket implementation to be disabled and force the user to implement it however they want? It's the blanket implementation that is getting me.

Amanieu:
Can you show the impl you are trying to write? I would think the Rust coherence rules would always prevent you from implementing a conflicting impl.

mattklein123:
> Can you show the impl you are trying to write? I would think the Rust coherence rules would always prevent you from implementing a conflicting impl.

Let me try to come up with a working example when I have some time. Basically I just want this code removed from my build: https://github.com/rust-lang/hashbrown/blob/40ab9e6cb9458f2920ff40530b42245db0c12787/src/lib.rs#L149-L157

The coherence rules are preventing me from implementing a conflicting impl. :)

Amanieu:
I believe the coherence rules would prevent you regardless of whether this impl exists in hashbrown. The problem is that your crate doesn't "own" the trait, so you cannot write blanket impls.

PR #351, opened 2022-08-10
Add an `into_ref` method, an immutable version of the `into_mut` method

Aceeri:
Was looking around for a way to return the reference from a function that inserts it using the entry and am currently just using the `into_mut()` because it will just coerce down into a `&'a V` anyways. But would be nice to have the type safety of it.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -3587,6 +3587,27 @@ impl<'a, K, V, S, A: Allocator + Clone> RawOccupiedEntryMut<'a, K, V, S, A> {
         unsafe { &self.elem.as_ref().1 }
     }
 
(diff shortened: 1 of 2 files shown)


Amanieu:
I don't see the advantage of this method over just calling `into_mut`.

Aceeri:
It is relatively niche but I mostly just don't want the thing to be modified after I insert it. I could just do
```rust
let immutable: &V = my_map.entry("thing").insert("my value").into_mut();
```
but does end up slightly confusing in the case of returning from the method.

Amanieu:
Sorry, but I think the benefits here are too small to justify increasing the API surface. An additional confusion is that the `HashMap` is still mutably borrowed while the returned reference is active.

PR #353, opened 2022-08-19
add support for architectures with 16-bit pointer width Closes #352

### LINKED ISSUE #352: hashbrown fails to compile on avr architectures because target_pointer_width is 16
I am porting an embedded app to AVR, and hashbrown is giving me several compile errors:
```
error[E0433]: failed to resolve: use of undeclared type `GroupWord`
  --> /home/thoth/.cargo/registry/src/github.com-1ecc6299db9ec823/hashbrown-0.3.1/src/raw/generic.rs:33:18
   |
33 |     let repeat = GroupWord::from(byte);
   |                  ^^^^^^^^^ use of undeclared type `GroupWord`

error[E0425]: cannot find value `K` in this scope
  --> /home/thoth/.cargo/registry/src/github.com-1ecc6299db9ec823/hashbrown-0.3.1/src/fx.rs:41:69
   |
41 |         self.hash = self.hash.rotate_left(5).bitxor(i).wrapping_mul(K);
   |                                                                     ^ help: a local variable with a similar name exists: `i`
```

it appears hashbrown has cfg checks based on target_pointer_width, but no clause cover the 16-bit pointer case (AVR has 16-bit pointers)

mutantbob:
One example is AVR-based microcontrollers like the atmega328 or atmega2560

### DIFF
--- a/src/raw/generic.rs
+++ b/src/raw/generic.rs
@@ -19,8 +19,8 @@ type GroupWord = u64;
     not(target_arch = "wasm32"),
 ))]
 type GroupWord = u32;
-#[cfg(target_pointer_width="16")]
-type GroupWord=u16;
+#[cfg(target_pointer_width = "16")]
+type GroupWord = u16;
 
 pub type BitMaskWord = GroupWord;
 pub const BITMASK_STRIDE: usize = 8;
--- a/src/raw/generic.rs
+++ b/src/raw/generic.rs
@@ -19,6 +19,8 @@ type GroupWord = u64;
     not(target_arch = "wasm32"),
 ))]
 type GroupWord = u32;
+#[cfg(target_pointer_width="16")]
+type GroupWord=u16;
 
 pub type BitMaskWord = GroupWord;
 pub const BITMASK_STRIDE: usize = 8;


mutantbob:
This is my attempt at fixing https://github.com/rust-lang/hashbrown/issues/352 .  Now to fix the ahash crate...

Amanieu:
LGTM, but you need to run `cargo fmt`.

mutantbob:
Closes #352

Amanieu:
After thinking about this for a bit, this may not be the right thing to do here. Using a group size that is too small will have a negative impact on workloads which frequently insert & remove elements since it makes it more likely to completely fill a group with tombstones. It might be better to use `u32` like the 32-bit architectures here.

Another issue is that AVR (RISC-V too) doesn't support an efficient implementation of `u32::trailing_zeros` and unaligned memory access which are crucial for the performance of the hash table. This can probably be looked over for now but in the future we may want a different implementation for these architectures.

Amanieu:
ping @mutantbob Can you change this PR to use `u32` on 16-bit architectures?

Amanieu:
Superseded by #368

PR #354, opened 2022-08-21
Add RawTable#is_full

braddunbar:
There's currently no way to tell  if a particular bucket is full. This can be handy for choosing a random element or iterating over a hash table in a specific order of buckets, so I've added the `is_full` function to RawTable.

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -796,7 +796,7 @@ impl<T, A: Allocator + Clone> RawTable<T, A> {
     {
         let index = self.bucket_index(&bucket);
         let old_ctrl = *self.table.ctrl(index);
-        debug_assert!(self.is_full(index));
+        debug_assert!(self.is_bucket_full(index));
         let old_growth_left = self.table.growth_left;
         let item = self.remove(bucket);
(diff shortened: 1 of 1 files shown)


braddunbar:
Ok, so I changed this up so the build is passing and also refactored a bit. Hope you like it!

Also, I'm considering using `is_bucket_full` instead of `is_full` because [`is_empty`](https://docs.rs/hashbrown/0.12.3/hashbrown/raw/struct.RawTable.html#method.is_empty) seems like it's the opposite of `is_full`, but it most certainly is not. Maybe it's only confusing to me though! 🤷‍♀️ 

I would love any feedback on the name, the general idea, or anything else you'd like to talk about! 😺

Amanieu:
I think `is_bucket_full` is a better name, `is_full` might be confused with asking whether the table is full to capacity.

Otherwise I think this is a good addition, but it should be `unsafe` since there is no bounds checking on the bucket index.

braddunbar:
Done and done. Thank you for the feedback @Amanieu!

braddunbar:
Thank you!

PR #355, opened 2022-08-22
make with_hasher_in const (consistent with with_hasher)

TennyZhuang:
I found that the following four methods have additional comments about their allocation:

* HashMap::with_hasher
* HashMap::with_hasher_in
* HashSet::with_hasher
* HashSet::with_hasher_in

There is no reason to make such a difference; in this PR, I make them consistent as the following behavior:

* All methods clarify that they will not allocate memory before the first insertion
* All methods are marked as const function.

We also bump the MSRV to 1.61.0 for the const_fn_trait_bound feature. Currently, 1.64.0 has been released, and I guess it's acceptable to break the 1.61.0 users.

Signed-off-by: TennyZhuang <zty0826@gmail.com>

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -50,7 +50,7 @@ jobs:
           thumbv6m-none-eabi,
           x86_64-pc-windows-gnu,
         ]
-        channel: [1.58.1, nightly]
+        channel: [1.61.0, nightly]
         include:
         - os: macos-latest
           target: x86_64-apple-darwin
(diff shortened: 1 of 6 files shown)


TennyZhuang:
https://github.com/rust-lang/rust/issues/93706, seems that it depends a feature which was stabilized?

Amanieu:
You need to bump the minimum Rust version in CI to 1.61 which is when the `const_fn_trait_bound` feature was stabilized.

TennyZhuang:
> You need to bump the minimum Rust version in CI to 1.61 which is when the `const_fn_trait_bound` feature was stabilized.

I prefer to do that after 1.64 stabilized, just hold the PR.

TennyZhuang:
Why were most actions canceled without errors? Are there some limitations?

Amanieu:
The runs are canceled once the first runner fails. In this case it seems that the 1.56.1 runner is failing, because you missed that one when changing the workflows (line 69).

TennyZhuang:
> The runs are canceled once the first runner fails. In this case it seems that the 1.56.1 runner is failing, because you missed that one when changing the workflows (line 69).

Oh, I'm curious why I miss it because I find & replace it globally. Thank you very much!

Amanieu:
Thanks!

Amanieu:
I can't reproduce this failure locally, let's try again?
@bors retry

Amanieu:
Actually this seems to be caused by: https://github.com/rust-lang/rust/pull/102748

Let's wait for that fix to get merged first.

JustForFun88:
I think it can be merged now, but please don't bump the version of crate because I want to document a raw part of crate

PR #358, opened 2022-09-03
Add HashSet#raw_table

braddunbar:
Just like #335 did for `HashMap`, I'd like to add access to the underlying `RawTable` for `HashSet`. I intend to use it in conjunction with #354 to pull random elements from a `HashSet`.

Let me know if I've missed something here or you'd like things implemented differently. I'll be happy to change it up!

### DIFF
--- a/src/set.rs
+++ b/src/set.rs
@@ -1,3 +1,5 @@
+#[cfg(feature = "raw")]
+use crate::raw::RawTable;
 use crate::{Equivalent, TryReserveError};
 use alloc::borrow::ToOwned;
 use core::fmt;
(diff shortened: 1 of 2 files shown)


src/set.rs
```diff
@@ -1135,6 +1137,26 @@ where
             None => None,
         }
     }
+
+    /// Returns a mutable reference to the [`RawTable`] used underneath [`HashSet`].
+    /// This function is only available if the `raw` feature of the crate is enabled.
+    ///
+    /// # Note
+    ///
+    /// Calling the function safe, but using raw hash table API's may require
```
Amanieu, line 1146:
```suggestion
    /// Calling the function is safe, but using raw hash table API's may require
```
This should probably be fixed in the `HashMap` version as well.

braddunbar: Fixed both places in 6041b5aef6ad990a379fd624b953c7ec26d62e18, along with a little tweaking of the grammar in that sentence. Let me know if you'd like something different!

Amanieu:
@bors r+

Thanks!

braddunbar:
Thank you!

PR #359, opened 2022-09-06
Fix issue with field retagging in scopeguard

Nilstrieb:
With field retagging in SB, this causes UB because the read out value is invalidated by the move into `mem::forget`. Using `ManuallyDrop<T>` avoids the problem. With that issue fixed, we can now enable field retagging in CI. Field retagging is required to justify some of the `noalias` optimizations that rustc
performs currently.

<details>
<summary>Miri error</summary>

```rust
Undefined Behavior: trying to retag from <565074> for Unique permission at alloc212719[0x20], but that tag does not exist in the borrow stack for this location
    --> src/scopeguard.rs:40:13
     |
40   |             value
     |             ^^^^^
     |             |
     |             trying to retag from <565074> for Unique permission at alloc212719[0x20], but that tag does not exist in the borrow stack for this location
     |             this error occurs as part of retag at alloc212719[0x20..0x40]
     |
     = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
     = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <565074> was created by a Unique retag at offsets [0x20..0x40]
    --> src/scopeguard.rs:33:25
     |
33   |             let value = ptr::read(&guard.value);
     |                         ^^^^^^^^^^^^^^^^^^^^^^^
help: <565074> was later invalidated at offsets [0x20..0x40] by a Unique retag
    --> src/scopeguard.rs:39:25
     |
39   |             mem::forget(guard);
     |                         ^^^^^
     = note: backtrace:
     = note: inside `scopeguard::ScopeGuard::<&mut raw::RawTable<(i32, i32)>, [closure@src/raw/mod.rs:1637:45: 1637:52]>::into_inner` at src/scopeguard.rs:40:13
note: inside `<raw::RawTable<(i32, i32)> as core::clone::Clone>::clone_from` at src/raw/mod.rs:1671:17
    --> src/raw/mod.rs:1671:17
     |
1671 |                 ScopeGuard::into_inner(self_);
     |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `<map::HashMap<i32, i32> as core::clone::Clone>::clone_from` at src/map.rs:202:9
    --> src/map.rs:202:9
     |
202  |         self.table.clone_from(&source.table);
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #360, opened 2022-09-07
Fix leak of memory if dropping function panic

JustForFun88:
This PR is similar to #348, but we are considering panic during table dropping. If the element from the `dropping table` panics from its `Drop impl`, the table will not be deallocated, leading to a memory leak. This pull request fixes that.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -8379,4 +8379,33 @@ mod test_map {
 
         map2.clone_from(&map1);
     }
+
+    // Test that map do not leak memory if dropping function panic
+    #[test]
+    #[should_panic = "panic in drop"]
+    fn test_panic_in_drop() {
+        #[derive(Clone)]
+        struct CheckedDrop {
+            panic_in_drop: bool,
+        }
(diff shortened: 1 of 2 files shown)


JustForFun88:
Looks like fail caused by [rust-lang/rust#101474](https://github.com/rust-lang/rust/issues/101474) as mentioned in #359

Amanieu:
I'm concerned that this increases code size (and thus compilation time) for little benefit: a panic in a Drop usually means that you're going to leak memory anyways. IMO the proper solution is to completely disallow unwinding from Drop (https://github.com/rust-lang/rfcs/pull/3288).

JustForFun88:
> I'm concerned that this increases code size (and thus compilation time) for little benefit: a panic in a Drop usually means that you're going to leak memory anyways. IMO the proper solution is to completely disallow unwinding from Drop ([rust-lang/rfcs#3288](https://github.com/rust-lang/rfcs/pull/3288)).

Hmm... yes, I think you're right. Complications do not bring any benefit. Closing this pull request.

PR #361, opened 2022-09-07
Small improvement in `RawTableInner::rehash_in_place` function

JustForFun88:
Small improvement on rehashing. There is no point in calculating the memory place first if we exit the loop in the next step.


JustForFun88:
Again test  failing caused by [rust-lang/rust#101474](https://github.com/rust-lang/rust/issues/101474) as mentioned in #359

Amanieu:
I would expect this to be handled by the compiler, but I guess it doesn't hurt.

@bors r+

PR #362, opened 2022-09-18
 Check that layout size fits in isize in Layout

JustForFun88:
Since [rust-lang/rust#95295](https://github.com/rust-lang/rust/pull/95295) we need check that Layout `size`, when rounded up to the nearest multiple of `align`, must not overflow isize (i.e., the rounded value must be less than or equal to `isize::MAX`)


JustForFun88:
This test is awesome:-) Looks like I forgot that our capacity is smaller than buckets number on 12.5%

PR #367, opened 2022-10-12
Rewrite `get_many_mut` methods

JustForFun88:
The previous form of the `HashMap::get_many_mut_inner` and `HashMap::get_many_unchecked_mut_inner` functions was not very optimal:

- firstly, the old RawTable API was not very convenient and did not combine well with each other (`get_mut` requires that `eq: impl FnMut(&T) -> bool`, get_many_mut requires that `eq` argument should be a closure such that `eq(i, k)` returns true if `k` is equal to the `i`th key to be looked up.).
- secondly, the hash of all keys was calculated, regardless of whether there are any of given keys in the table at all.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1730,9 +1730,15 @@ where
     where
         Q: Hash + Equivalent<K>,
     {
-        let hashes = self.build_hashes_inner(ks);
-        self.table
-            .get_many_mut(hashes, |i, (k, _)| ks[i].equivalent(k))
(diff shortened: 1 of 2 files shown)


Amanieu:
I'm not actually sure this is an improvement, have you benchmarked this? It adds a lot of branches in the generated code to handle lazy hash computation. Straight up computing the hashes and doing multiple lookups might result is faster code. In practice, this API is used when you have few keys (usually just 2) and you're already pretty sure all are in the table.

PR #374, opened 2022-11-15
Remove drain-on-drop behavior from DrainFilter

the8472:
This is a breaking change for hashbrown but not for std because in std the drain impl is still unstable.

This is part of stdlib [ACP 136](https://github.com/rust-lang/libs-team/issues/136).


the8472:
The ACP has been approved so this should be ready to review.
Note that it'll need a bump to 0.14.

the8472:
Ah, I forgot that there was also a request to rename the API in std. So this could be made non-breaking in hashbrown by adding the new behavior under the new name and keeping the old impl.

Amanieu:
I think it's fine to just make it a breaking change. I am planning to make a major version bump anyways.

the8472:
Ready for review.

PR #378, opened 2022-12-30
chore: update version in README

sno2:


sno2:
It seems CI is broken. Hopefully the diff is enough to validate this change?

Issue #384, opened 2023-01-09
hashbrown is not following strict provenance


RalfJung:
hashbrown is using int-to-ptr casts in some places, instead of the strict provenance APIs:

https://github.com/rust-lang/hashbrown/blob/09dc17e0d686cdeba15920a4f77462c572e7cad1/src/raw/mod.rs#L306

https://github.com/rust-lang/hashbrown/blob/09dc17e0d686cdeba15920a4f77462c572e7cad1/src/raw/mod.rs#L326

https://github.com/rust-lang/hashbrown/blob/09dc17e0d686cdeba15920a4f77462c572e7cad1/src/raw/mod.rs#L334

(there might be more)

The cases I have found so far are all inside `if Self::IS_ZERO_SIZED_TYPE {`, so I assume these are "fake" pointers that are never dereferenced, and could be using [`ptr::invalid_mut`](https://doc.rust-lang.org/nightly/core/ptr/fn.invalid_mut.html).

The `fuzzy_provenance_casts` lint can be used to statically check for such casts. Currently however the strict provenance functions are still [unstable](https://github.com/rust-lang/rust/issues/95228).

zakarumych:
Should hashbrown use `sptr` crate for the time Strict Provenance API is unstable?

RalfJung:
@Amanieu is there a usual cadence for syncing hashbrown changes to the standard library?

Looks like the version shipped with current nightly still has the old int2ptr casts in it.

Amanieu:
We currently have several breaking changes on master, and I'm finishing up some new feature which will be released in hashbrown 0.14. I think the release should happen in the next week or two.

RalfJung:
Cool, thanks for the update!

PR #390, opened 2023-02-04
use strict provenance APIs on nightly

RalfJung:
This makes hashbrown compatible with strict provenance, which helps Miri in ensuring pointers are used correctly.

Fixes https://github.com/rust-lang/hashbrown/issues/384


RalfJung:
Clippy has a bunch of complaints but those are not introduced by this PR.

RalfJung:
Strangely I don't see any permissive provenance warnings in the CI job for the first commit. Looks like these ZST code paths do not get executed in the test suite. I added a test for one of them, not sure how to reach the other ones.

RalfJung:
@Amanieu I have rebased and resolved the conflicts.

PR #392, opened 2023-02-10
Doc `new_uninitialized` and `fallible_with_capacity`

JustForFun88:


src/raw/mod.rs
```diff
@@ -1072,6 +1072,11 @@ where
 }
 
 impl<A> RawTableInner<A> {
+    /// Creates a new empty hash table without allocating any memory.
+    ///
+    /// In effect this returns a table with exactly 1 bucket. However we can
+    /// leave the data pointer dangling since that bucket is never written to
```
Amanieu, line 1078:
```suggestion
    /// leave the data pointer dangling since that bucket is never accessed
```

src/raw/mod.rs
```diff
@@ -1086,6 +1091,18 @@ impl<A> RawTableInner<A> {
 }
 
 impl<A: Allocator + Clone> RawTableInner<A> {
+    /// Allocates a new [`RawTableInner`] with the given number of buckets.
+    /// The control bytes are left uninitialized.
```
Amanieu, line 1095: The buckets are also uninitialized.

PR #400, opened 2023-02-16
Fix `HashSet::get_or_insert_with`

### LINKED ISSUE #399: Violating the set invariant with `HashSet::get or_insert_with`
The implementation of `HashSet::get_or_insert_with` is not good. This methods make it quite easy to break the invariant of set that all elements are repeated in set only once. Let's say this test passes:
```rust
#[test]
fn some_invalid_insert() {
    let mut set = HashSet::new();
    set.insert(1);
    set.get_or_insert_with(&2, |_| 1);
    set.get_or_insert_with(&3, |_| 1);
    assert_eq!(vec![&1, &1, &1], set.iter().collect::<Vec<_>>());
}
```
May be we need document it?

JustForFun88:
Fix #399. Tried to do it without additional overhead. Unless there are additional comparisons.


yyny:
Should an `unsafe _unchecked` function also be added? Hashing and comparing the `insert_value` should be unnecessary if this function is used properly.

JustForFun88:
> Should an `unsafe _unchecked` function also be added? Hashing and comparing the `insert_value` should be unnecessary if this function is used properly.

Well, strictly speaking, rehashing was present in the old version of the code.  Only two comparisons and a panic were added here.  Let's see what @Amanieu says :-).

In theory, as I understand it, there is ***nothing*** that violates memory or causes **UB** in the fact that there are two or more identical elements in a Hashmap or HashSet.  It just increases the collision, the elements after the first one will be lost and will only show up when iterating. The documentation for `HashMap::insert_unique_unchecked` scares people a little. Anyway, I can't even imagine when `insert_unique_unchecked` «operation may panic, loop forever, or any following operation with the map may panic, loop forever or return arbitrary result».

clarfonthey:
Going to close this since this appears to have been fixed since this PR was opened: this now does an equivalence check and panics when it's not satisfied.

Issue #382, opened 2023-01-06
Add a `take_any` method


timothee-haudebourg:
It is not the first time I'm missing a method to remove an arbitrary element from a set or map. This issue has been solved in the standard library for `BTreeSet` and `BTreeMap` with the recent introduction of the `pop_first` and `pop_last` methods, but there is no equivalent for `HashMap`/`HashSet`.

What I'm looking for is as simple as that:

/// Removes an arbitrary element from the set.
fn take_any(&mut self) -> Option<T>

It is currently impossible to remove an element from a `HashMap`/`HashSet` through `remove` or `take` without a preexisting reference to an element/key of the `set`/`map`. So to have such reference the only way is to clone one beforehand:

```rust
fn take_any<T: Clone>(set: &mut HashSet<T>) -> Option<T> {
  let result = set.iter().next()?.clone();
  set.remove(&result);
  result
}
```

It is possible without the `Clone` bound with the `drain_filter` that allows to drain the first item and keep the rest, but it seems expansive to me.

```rust
fn take_any<T>(set: &mut HashSet<T>) -> Option<T> {
  let mut taken = false;
  set.drain_filter(|_| {
    let take = !taken;
    taken = true;
    take
  }).next()
}
```

As for the name, it could also be named `pop_any` to reflect the  `pop_first` and `pop_last` methods on `BTreeSet` and `BTreeMap`.

timothee-haudebourg:
Following the [related thread on the internals forum](https://internals.rust-lang.org/t/add-a-take-any-method-to-hashmap-hashset/18088/2) it appears that it would be a O(n) operation, which removes the advantages of using a `HashMap`. But on the eventual case where it cannot be avoided, I still think it would be nice to have this function to avoid the `Clone` bound and the cost of using `drain_filter` that iterates over the whole map/set.

Amanieu:
My recommendation in these cases is to use `IndexMap` instead, where the `pop` operation is guaranteed to be `O(1)`.

timothee-haudebourg:
[...]

PR #420, opened 2023-04-12
Implement iterator which starts at arbitrary location

alfa07:
This PR implements iterator which starts at arbitrary location in the table. There are several motivations for this change:

- For security reasons one may wish to start iterations at arbitrary locations in the table. That what golang is doing
- For implementing efficient pseudo-random sampling from the table (e.g for use in probabilistic algorithms)
- For being able to port golang code with 100% semantic preservation

See this [issue](https://github.com/al8n/stretto/issues/37) for a real-world motivating example.

That is just POC, if the idea in principle is acceptable I would be happy to shape it with some guidance from maintainers.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -824,7 +824,7 @@ impl<K, V, S, A: Allocator + Clone> HashMap<K, V, S, A> {
     /// assert_eq!(map.len(), 3);
     /// let mut vec: Vec<(&str, i32)> = Vec::new();
     ///
-    /// for (key, val) in map.iter_at(0x517cc1b727220a95_usize) {
+    /// for (key, val) in map.iter_at(0x517cc1b727220a95u64 as usize) {
     ///     println!("key: {} val: {}", key, val);
     ///     vec.push((*key, *val));
     /// }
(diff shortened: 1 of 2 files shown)


phantomhker:
it does not work properly, that's my testing code
```
    fn print_map_random(m: &HashMap<&str, &str>,times: usize) {
        for _ in 0..times {
            for (k, v) in m.iter_at(0) {
                println!("{k}-{v}");
            }
            println!("=========")
        }
    }

    #[test]
    fn test_map() {
        let mut m = HashMap::new();
        m.insert("k1", "v1");
        m.insert("k2", "v2");
        m.insert("k3", "v3");
        m.insert("k4", "v4");
        m.insert("k5", "v5");
        m.insert("k6", "v6");
        print_map_random(&m, 10);
    }
```

the8472:
> for (k, v) in m.iter_at(0) {

You need to randomize the hint to start at random offsets of the table.

Amanieu:
This feels awkward for similar reasons as #382: you really want to be using `IndexMap` instead if you are doing any kind of operation that depends on the ordering/distribution of the keys.

alfa07:
@Amanieu Using IndexMap is not the right choice for this use case. We want to access elements of the HashMap with O(1) per access in pseudo random order. Hashing of keys in the HashMap provides this order for free. IndexMap preserves insertion order which we don't want here as it is not pseudo random. Also IndexMap is plainly more expensive.

dzmitry-lahoda:
We have similar(same?) use case. We store view state (cache like) in Hashbrown map. Sometimes we want to pick any random value out of map and do some validation on its current state.

PR #422, opened 2023-04-20
Relax the trait bounds of `HashSet::raw_table{,_mut}`

kawadakk:
Removes the trait bounds `T: Eq + Hash, S: BuildHasher` from `HashSet<T, S, A>::raw_table{,_mut}`.

Considering that the `HashMap` counterpart doesn't have these bounds, they are likely accidental.

### DIFF
--- a/src/set.rs
+++ b/src/set.rs
@@ -1221,7 +1221,9 @@ where
             None => None,
         }
     }
+}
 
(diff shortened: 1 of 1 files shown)


kawadakk:
Didn't know that renaming a branch would automatically close associated PRs. Sorry for the noise.

PR #428, opened 2023-05-01
Special case `clear()` on empty tables

fornwall:
Special case `clear()` to avoid `O(capacity)` runtime even on an empty set.

There are workloads which currently needs `if !map.is_empty() { map.clear(); }` for optimal performance, which looks odd (and that it's necessary may be surprising).

Special casing empty table aligns with what [java.util.HashMap](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/HashMap.java#L867) is doing.

After discussion with @forny.


Amanieu:
LGTM! However I don't think a benchmark is necessary here, it will just clutter the benchmark list.

fornwall:
@Amanieu Thanks! Have removed the benchmark now.

PR #436, opened 2023-06-08
Avoid reporting a capacity of zero when there is still space in the map.

### LINKED ISSUE #79178: HashMap reports zero capacity without freeing memory
It's generally expected that when std collections return a capacity of zero, that they are not backed by any allocation. I believe this was also true for the HashMap prior to hashbrown, although I could be wrong.

In any case, whilst `HashMap::capacity()` is documented as a *lower bound* I still find it extremely surprising that it reports a capacity of zero without freeing its memory.

You can reproduce the issue with the following program (using mockalloc to detect the leaked memory):

```rust
use std::alloc::System;
use std::collections::hash_map::{DefaultHasher, HashMap};
use std::hash::BuildHasherDefault;

use mockalloc::Mockalloc;

#[global_allocator]
static ALLOCATOR: Mockalloc<System> = Mockalloc(System);

const NUMBERS: &[usize] = &[
    25, 38, 41, 42, 89, 115, 184, 237, 273, 286, 300, 326, 377, 413, 482, 536, 602, 650, 702, 746,
    750, 807, 810, 836, 960, 979, 982, 1007,
];

fn main() {
    let mut guard =
        HashMap::<String, (), _>::with_hasher(BuildHasherDefault::<DefaultHasher>::default());
    mockalloc::assert_allocs(|| {
        for &n in NUMBERS {
            let k = n.to_string();
            guard.insert(k, ());
        }
        for &n in NUMBERS {
            let k = n.to_string();
            guard.remove(&k);
            if guard.len() * 3 < guard.capacity() {
                guard.shrink_to_fit();
            }
        }
        eprintln!("Capacity: {}", guard.capacity());
        //guard.shrink_to_fit();
    });
}
```

Given these lines: 
```rust
if guard.len() * 3 < guard.capacity() {
    guard.shrink_to_fit();
}
```

I expected all the memory to be reclaimed by the end of the loop. However, this is not the case: the memory is only reclaimed if `shrink_to_fit()` is called an additional time.

IMO, the fix should be two-fold:
1) When the last element of a HashMap is removed, it should automatically clear out any tombstones so that the `capacity()` method is accurate again.

2) A new method should be introduced to obtain the upper bound on the capacity. At the moment code which attempts to shrink under-utilized hashmaps is broken, and it's not possible to fix this code without a new method.

Diggsey:
Fixes https://github.com/rust-lang/rust/issues/79178

There are two changes in this PR:

1) Add a `max_capacity()` method. The `capacity()` method provides a lower bound on the capacity, this complements it by providing an upper bound. There are some use-cases (such as deciding when to call `shrink_to_fit()`) when using the upper bound on the capacity may be more appropriate.

2) Recover "lost" capacity when the map is fully emptied if there are a large number of tombstones. Currently this happens when the reported capacity would be less than or equal to half the actual capacity of the map. The most important case is when the reported capacity would be zero, but we haven't deallocated the backing memory: with this change a positive capacity will be reported.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -619,6 +619,15 @@ impl<K, V, S, A: Allocator + Clone> HashMap<K, V, S, A> {
         self.table.capacity()
     }
 
+    /// Returns the maximum number of elements the map could potentially hold
+    /// without reallocating.
+    ///
+    /// This number is an upper bound.
+    #[cfg_attr(feature = "inline-more", inline)]
+    pub fn max_capacity(&self) -> usize {
+        self.table.max_capacity()
+    }
+
     /// An iterator visiting all keys in arbitrary order.
     /// The iterator element type is `&'a K`.
     ///
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -1389,6 +1389,15 @@ impl<T, A: Allocator + Clone> RawTable<T, A> {
         self.table.items + self.table.growth_left
     }
 
+    /// Returns the maximum number of elements the map could potentially hold
+    /// without reallocating.
+    ///
+    /// This number is an upper bound.
+    #[inline]
+    pub fn max_capacity(&self) -> usize {
+        bucket_mask_to_capacity(self.table.bucket_mask)
+    }
+
     /// Returns the number of elements in the table.
     #[inline]
     pub fn len(&self) -> usize {
@@ -2485,6 +2494,15 @@ impl<A: Allocator + Clone> RawTableInner<A> {
         // SAFETY: the caller must uphold the safety contract for `erase` method.
         self.set_ctrl(index, ctrl);
         self.items -= 1;
+
+        // If the last item was removed we can do some cleanup
+        if unlikely(self.items == 0) {
+            let desired_capacity = bucket_mask_to_capacity(self.bucket_mask);
+            if self.growth_left <= desired_capacity / 2 {
+                // Restore an accurate view of the capacity
+                self.clear_no_drop()
+            }
+        }
     }
 }
 


Amanieu:
I don't mind adding a `max_capacity` method, but I am against adding the extra code in `erase`. My main concern is that this increases code size for a rare condition, and that tombstones will naturally be cleared as more items are inserted in the map.

Have you considered using `shrink_to` instead of `shrink_to_fit` to limit the size of a table to a given capacity? This doesn't rely on `growth_left` for calculations and is therefore independent of the number of tombstones in the table.

Diggsey:
> My main concern is that this increases code size for a rare condition, and that tombstones will naturally be cleared as more items are inserted in the map.

This is intended to restore the invariant that `(map.capacity() == 0) => no memory allocated`. This invariant was true prior to the implementation of `std::hash_map::HashMap` being replaced by hashbrown, and is true for the other std collections, but is not currently true in hashbrown.

My solution could be restricted further to target *only* this specific case, rather than the more general solution here, but I think it would still require code in `erase`.

Are you worried about the total code size, or the code size in `erase` specifically? I don't think this should add more than a single instruction to `erase` since the condition being checked will already be generated by the previous line (at least on x86).

> Have you considered using shrink_to instead of shrink_to_fit to limit the size of a table to a given capacity? This doesn't rely on growth_left for calculations and is therefore independent of the number of tombstones in the table.

I don't think that helps with my specific use-case, since I was using `capacity()` to decide *whether* to try to shrink the table - `max_capacity()` alone does solve the problem for me, but it left this footgun for others due to the broken invariant, which I wanted to also fix.

Amanieu:
> This is intended to restore the invariant that (map.capacity() == 0) => no memory allocated. This invariant was true prior to the implementation of std::hash_map::HashMap being replaced by hashbrown, and is true for the other std collections, but is not currently true in hashbrown.

This has never been something that was guaranteed. The documentation specifically says that the capacity is the number of elements that can be inserted without reallocation, and makes no mention that 0 has a special meaning.

> I don't think that helps with my specific use-case, since I was using `capacity()` to decide _whether_ to try to shrink the table - `max_capacity()` alone does solve the problem for me, but it left this footgun for others due to the broken invariant, which I wanted to also fix.

I believe just replacing this:

```rust
if guard.len() * 3 < guard.capacity() {
    guard.shrink_to_fit();
}
```

with this:

```rust
guard.shrink_to(guard.len() * 3);
```

should just work. It will also properly free the backing memory of the hashmap when `len` is 0.

Diggsey:
> This has never been something that was guaranteed. The documentation specifically says that the capacity is the number of elements that can be inserted without reallocation, and makes no mention that 0 has a special meaning.

According to the documentation it would be valid for `capacity()` to always return 0. That doesn't make it a reasonable implementation choice though.

If hashbrown has memory allocated and isn't storing any elements, then there's definitely room to insert at least one element without reallocating.

Another way to avoid this would be to special-case it in `capacity()`, which would avoid bloating `erase()` at least...

> I believe just replacing this ... with this ... should just work. It will also properly free the backing memory of the hashmap when len is 0.

This would cause unnecessary extra resizing, and wouldn't reclaim all the memory in the "normal" case though.

PR #440, opened 2023-07-01
Update Cargo.toml

ISwearImABot:
Patches include bug fixes and improvements.

Result `cargo test`:
```
test result: ok. 221 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 101.92s
```

Some improvements are even significant.

[Update `aHash 0.8.1 [Nov 1, 2022]`](https://github.com/tkaitchuck/aHash/compare/v0.8.0...v0.8.1):
```
- Fix compiler error on wasm with default features.
- Significant improvements to speed on unusual architectures (Things other than: x86_64, aarch64, mips64, powerpc64, riscv64gc, or s390x)
...
- Fix AVR support by using atomic-polyfill
- Simplify the way that flags work
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,11 +14,11 @@ rust-version = "1.64.0"
 
 [dependencies]
 # For the default hasher
-ahash = { version = "0.8.0", default-features = false, optional = true }
+ahash = { version = "0.8.3", default-features = false, optional = true }
 
 # For external trait impls
 rayon = { version = "1.0", optional = true }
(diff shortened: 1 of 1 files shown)


CryZe:
Bumping the patch versions generally doesn't do much as cargo update will update them anyway, so this PR may not be doing what you think it does. However if there's security issues in old versions, I guess bumping the patch versions in the Cargo.toml could make sense.

PR #442, opened 2023-07-09
Use the `Equivalent` trait from the `equivalent` crate

Amanieu:
This allows types implementating the `Equivalent` trait to work with other hash table, such as `IndexMap`.

See https://github.com/bluss/indexmap/issues/253.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -117,6 +117,11 @@ pub mod hash_set {
 pub use crate::map::HashMap;
 pub use crate::set::HashSet;
 
+#[cfg(feature = "equivalent")]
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -117,6 +117,11 @@ pub mod hash_set {
 pub use crate::map::HashMap;
 pub use crate::set::HashSet;
 
+#[cfg(feature = "equivalent")]
+use equivalent::Equivalent;
```
cuviper, line 121: Shouldn't this be `pub` to match the API without the feature?
(or else hide them both in a breaking change)

PR #447, opened 2023-07-19
Reserve extra element only if not found

Dandandan:
Trying to use `find_or_find_insert_slot` I saw it always reserves an extra element, even if the element is found.
It seems more close to what's expected and the docs to only reserve when actually returning `InsertSlot`


Dandandan:
Ah I found it only does so for reserving 1 element before returning the insertslot (and doesn't mutate the needed capacity).

So while this potentially allocates unnecessary for 1 element, this doesn't seem to be a problem.

PR #449, opened 2023-07-29
Allow serializing `HashMap`s that use a custom allocator

tomcc-reactor:
Hi,
I'm using `HashMap` with custom allocators (eg. not `Global`) in my project, and I got hit with a "`HashMap` doesn't implement trait `Serialize`" error.

It looks like it has a very trivial fix, eg. exposing the `A` generic parameter on the impl so that it doesn't default to `Global`.

I have a very limited understanding of HashBrown so let me know if this is a change that makes sense!
- Is `crate::raw::Allocator` the right way to import the `Allocator` trait?
- Do you need `Deserialize` to be fixed as well?
- What about `HashSet`?

Anyway, if the change makes sense I can try to clean it up a bit more :)

### DIFF
--- a/src/external_trait_impls/serde.rs
+++ b/src/external_trait_impls/serde.rs
@@ -21,11 +21,12 @@ mod map {
 
     use super::size_hint;
 
-    impl<K, V, H> Serialize for HashMap<K, V, H>
+    impl<K, V, H, A> Serialize for HashMap<K, V, H, A>
     where
         K: Serialize + Eq + Hash,
         V: Serialize,
         H: BuildHasher,
+        A: crate::raw::Allocator + Clone,
     {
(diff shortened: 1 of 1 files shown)


Amanieu:
> * Is `crate::raw::Allocator` the right way to import the `Allocator` trait?

Yes, but put it in a `use` statement.

> * Do you need `Deserialize` to be fixed as well?

Yes, this should use `A: Allocator + Default`.

> * What about `HashSet`?

Please implement this for `HashSet` as well.

tomcc-reactor:
Done, let me know if it looks ok. 
It's just a bit sad that `serde` doesn't really allow a non-default allocator, but for now that works for my use case.
`HashSet::deserialize_in_place` doesn't require default in practice, but the trait bound will prevent it from working...

PR #450, opened 2023-08-03
Add reinsert to RawOccupiedEntryMut

1kjo:
It is sometimes useful to modify the key associated with an entry without changing the value. For example, when building a cache where keys can change.

It is currently possible to implement this using hashbrown::HashMap by removing the entry and adding a new one with the same value. However, to prevent conflicts with the new key, the code has to:
1. Lookup the entry with the new key, and return an error if it occupied.
2. Delete the entry with the old key.
3. Move the value into the entry with the new key.

This is not possible with the raw entry API. A RawOccupiedEntryMut mutably borrows the map so only one can exist at a time. A workaround is to compute the hash of the new key, do the lookup, destroy the raw entry and rebuild one with the hash later, but this is not ideal as it still does two lookups with the hash when one is sufficient.

RawOccupiedEntryMut already has insert_key that allows changing the key associated with an entry, but it does not actually reinserts the entry into the map, so if the new key has a different hash then the entry will no longer be accessible.

Add a reinsert method to reinsert the entry into the map after its key has been changed.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -3942,6 +3942,168 @@ impl<'a, K, V, S, A: Allocator + Clone> RawOccupiedEntryMut<'a, K, V, S, A> {
     }
 }
 
+impl<'a, K: Eq, V, S, A: Allocator + Clone> RawOccupiedEntryMut<'a, K, V, S, A> {
+    /// Reinserts the entry into the map.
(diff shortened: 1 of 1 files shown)


Amanieu:

PR #452, opened 2023-08-17
Add a `insert_unique_hashed_nocheck` method

Zoxc:
This adds a `insert_unique_hashed_nocheck` method which inserts a key-value pair being given a hash for the key and not checking if the key already exists. It's a more powerful variant of `insert_unique_unchecked` since it takes a hash value and returns `RawOccupiedEntryMut`. I found a use for it in rustc's query system.


Amanieu:

PR #457, opened 2023-08-23
Relax MSRV to 1.63.0

cuviper:
No code changes are required, and this allows the current compiler
shipped by Debian stable (bookworm) to work. The only consideration is
that recent versions of `allocator-api2` do require 1.64, but the
minimum dependency 0.2.9 is fine, and this is optional anyway.


cuviper:
This is motivated by https://github.com/hyperium/h2/pull/698#issuecomment-1690663550

cuviper:
Thanks! Could we get a release, so I can propagate this relaxation to `indexmap`?

PR #458, opened 2023-08-24
Simplify `Clone` by removing redundant guards

JustForFun88:
These extra guards only complicate the code where it is not needed anyway. The `Drop` function checks the number of `items` in a table before dropping elements (see below), so that dropping the uninitialize table as well as table with no actual data (but with `FULL` control bytes) is safe . Added tests to check that the current behavior of `Drop` won't change in the future.

```rust
impl<T, A: Allocator + Clone> Drop for RawTable<T, A> {
    fn drop(&mut self) {
        if !self.table.is_empty_singleton() {
            unsafe {
                self.drop_elements();
                self.free_buckets();
            }
        }
    }
}

impl<T, A: Allocator + Clone> RawTable<T, A> {
    unsafe fn drop_elements(&mut self) {
        if Self::DATA_NEEDS_DROP && !self.is_empty() {
            for item in self.iter() {
                item.drop();
            }
        }
    }
}
```


JustForFun88:
By the way, the old implementation with ManuallyDrop seems to be leaking the allocator.  I'll add a test later

PR #459, opened 2023-08-25
Update RawTable with fallible APIs

### LINKED ISSUE #456: Support fallible eq and hasher in raw API
Hi!

I want to use the raw API of `hashbrown` as the underlying table implementation in [Rune](https://github.com/udoprog/rune-rs/rune).

One aspect which has proven to be awkward is that the hasher I need to pass into `RawTable::find_or_find_insert_slot` is infallible, which works well for a statically typed language, but less so in Rune where I want to store and deal with dynamically typed entries directly.

> Note: I've implemented a prototype of this in my [`raw-infallible-context` branch][branch].

What I need to do is something like this:

```rust
let mut table: RawTable<(Value, Value)> = todo!();
let entry: (Value, Value) = todo!();

// NB: might fail if the key can't be hashed.
let hash = key.hash()?;

// Comparison might fail.
let eq = |other: &(Value, Value)| -> Result<bool, MyError> { Value::eq(&entry.0, &other.0) };
// Hashing might fail.
let hasher = |key: &(Value, Value)| -> Result<u64, MyError> { Value::hash(&key.0) }

let result = table.find_or_find_insert_slot(hash, eq, hasher);
// deal with my custom result
```

The change needed would be fairly straightforward ([I think][branch]) but a bit tedious, what I think is needed is to refactor the API into a pair of methods like this (ignore naming):

```rust
use std::convert::Infallible;

#[inline(always)]
fn into_ok<T>(result: Result<T, Infallible>) -> T {
    match result {
        Ok(value) => value,
        Err(error) => match error {},
    }
}

#[inline(always)]
fn infallible_eq<T>(mut f: impl FnMut(&T) -> bool) -> impl FnMut(&T) -> Result<bool, Infallible> {
    move |value| Ok::<_, Infallible>(f(value))
}

#[inline(always)]
fn infallible_hasher<T>(f: impl Fn(&T) -> u64) -> impl Fn(&T) -> Result<u64, Infallible> {
    move |value| Ok::<_, Infallible>(f(value))
}

impl RawTable {
    /// Backwards compatibility
    #[inline(always)]
    pub fn find_or_find_insert_slot(
        &mut self,
        hash: u64,
        eq: impl FnMut(&T) -> bool,
        hasher: impl Fn(&T) ->  u64,
    ) -> Result<Bucket<T>, InsertSlot> {
        into_ok(self.find_or_find_insert_slot2(hash, infallible_eq(eq), infallible_hasher(hasher)))
    }

    /// New method
    pub fn find_or_find_insert_slot2<E>(
        &mut self,
        hash: u64,
        mut eq: impl FnMut(&T) -> Result<bool, E>,
        hasher: impl Fn(&T) -> Result<u64, E>,
    ) -> Result<Result<Bucket<T>, InsertSlot>, E> {
        todo!()
    }
}
```

The signatures leave something to be desired and such a change would have to be propagated through the rest of the API, [the majority if which have proven to be internal][branch]. I would also hope that the compiler can easily prove the infallible variants are no-ops but this is something that should probably be investigated.

[branch]: https://github.com/rust-lang/hashbrown/compare/master...udoprog:hashbrown:raw-infallible-context?expand=1

udoprog:
This introduces additional raw functions which allows for the use of infallible `eq` and `hasher` implementations with a passed around context.

See #456 for a detailed motivation.

Note that `eq` could be `impl Fn` rather than `impl FnMut` after this.

I've also not bothered converting methods that I'm not using yet, such as `get_many_unchecked_mut` since I want to get an idea whether this change would be accepted or not.

Closes #456

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -1,6 +1,7 @@
 use crate::alloc::alloc::{handle_alloc_error, Layout};
 use crate::scopeguard::guard;
 use crate::TryReserveError;
+use core::convert::Infallible;
 use core::iter::FusedIterator;
 use core::marker::PhantomData;
 use core::mem;
@@ -64,6 +65,27 @@ fn invalid_mut<T>(addr: usize) -> *mut T {
 #[inline]
 unsafe fn offset_from<T>(to: *const T, from: *const T) -> usize {
     to.offset_from(from) as usize
+}
+
+/// Helper for coercing an infallible result into `Ok`.
+#[inline(always)]
+fn into_ok<T>(result: Result<T, Infallible>) -> T {
+    match result {
+        Ok(value) => value,
+        Err(error) => match error {},
+    }
+}
+
+#[inline(always)]
+fn infallible_eq<T>(
+    mut f: impl FnMut(&T) -> bool,
+) -> impl FnMut(&mut (), &T) -> Result<bool, Infallible> {
+    move |_, value| Ok::<_, Infallible>(f(value))
+}
+
+#[inline(always)]
+fn infallible_hasher<T>(f: impl Fn(&T) -> u64) -> impl Fn(&mut (), &T) -> Result<u64, Infallible> {
+    move |_, value| Ok::<_, Infallible>(f(value))
 }
 
 /// Whether memory allocation errors should return an error or abort.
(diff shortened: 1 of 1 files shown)


udoprog:
:thinking:
```
   = note: /usr/bin/ld: /rust/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-19964a87984cce8a.rlib(std-19964a87984cce8a.std.d3cd61b6198d5d17-cgu.0.rcgu.o): TLS transition from R_X86_64_TLSLD to R_X86_64_TPOFF32 against `_ZN3std9panicking11panic_count17LOCAL_PANIC_COUNT7__getit3VAL17h3ab482653e57521bE.0' at 0x5af in section `.text._ZN3std2rt19lang_start_internal17h8c00664471eb83c4E' failed
          /rust/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-19964a87984cce8a.rlib: error adding symbols: Bad value
          collect2: error: ld returned 1 exit status
```

JustForFun88:
Issue [rust-lang/rust#115239](https://github.com/rust-lang/rust/issues/115239)

udoprog:
Closing based on feedback in #456

PR #461, opened 2023-08-27
Shared context

udoprog:
Supersedes #459

Based on [this comment](https://github.com/rust-lang/hashbrown/issues/456#issuecomment-1692900055) it seems like there's interest in passing shared contexts around in the raw API, this PR extracts just that aspect from #459 in so that at least the context aspect of #456 will be accepted.

Motivation: Passing a context around is necessary to allow `eq` and `hasher` to share the same mutable environment, Both can't implement `FnMut` and capture the same environment, because those two captures would both need to have exclusive access to it which is disallowed by Rust's aliasing rules.

There's three levels of possibly increasing controversy in this PR, separated by three commits:
* The first one introduces a context which is passed around through `*_with_context` methods, maintaining the same API in the process (`eq` is `impl FnMut`).
* The second commit makes both functions `impl Fn` for the new methods, meaning if you want to use a mutable shared environment there you *must* use the `&mut C` context variable to pass it around.
* The third commit makes the internals use function pointers, completely disallowing any context that is not explicitly passed through `&mut C`.

If there is interest in passing a context around like this, feel free to decide which one you want to keep, and which one to discard.

Thank you!

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -3242,7 +3242,11 @@ impl<'a, K, V, S, A: Allocator + Clone> RawEntryBuilderMut<'a, K, V, S, A> {
     where
         for<'b> F: FnMut(&'b K) -> bool,
     {
-        match self.map.table.find(hash, |(k, _)| is_match(k)) {
+        match self
+            .map
+            .table
+            .find_with_context(&mut is_match, hash, |is_match, (k, _)| is_match(k))
+        {
             Some(elem) => RawEntryMut::Occupied(RawOccupiedEntryMut {
                 elem,
                 table: &mut self.map.table,
@@ -3313,7 +3317,11 @@ impl<'a, K, V, S, A: Allocator + Clone> RawEntryBuilder<'a, K, V, S, A> {
     where
         F: FnMut(&K) -> bool,
     {
-        match self.map.table.get(hash, |(k, _)| is_match(k)) {
+        match self
+            .map
+            .table
+            .get_with_context(&mut is_match, hash, |is_match, (k, _)| is_match(k))
+        {
             Some((key, value)) => Some((key, value)),
             None => None,
         }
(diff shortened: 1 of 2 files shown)


Amanieu:
What I had in mind was actually much simpler. `find_or_find_insert_slot` is the only method which takes both `hasher` and `eq` callbacks. Therefore only that method needs to support contexts which can be shared by both callbacks.

When there is only one callback, the caller can easily borrow the context exclusively in the closure environment. We would only need to change the `hasher` callbacks to be `FnMut` instead of `Fn`, but that shouldn't cause any problems.

udoprog:
It's certainly possible, I opted not to to maintain a more coherent API (and was considering commenting on it). Consider how the caller would otherwise need to build multiple kinds of adapters ([like these](https://github.com/rust-lang/hashbrown/blob/master/src/map.rs#L223) or [these](https://github.com/rune-rs/rune/blob/main/crates/rune/src/hashbrown/table.rs#L265)) when interacting with it.

From your comment I gather that's not an overriding concern for you?

JustForFun88:
It would be nice if the functions were documented (preferably with examples).  After all, it is impossible to understand what `context` is and how to use it either at first or second glance.

We have to study the implementation.

Amanieu:
I would prefer keeping the API simple and only add the context variant for `find_or_find_insert_slot`. Otherwise there might be confusion if there are 2 ways of doing the same thing (context vs capturing in closure).

udoprog:
Ok, which of the three variants listed above do you want `find_or_find_insert_slot_with_context` to use?

Amanieu:
I would say the first variant (keep the API the same, just add a context), except that all `hasher: impl Fn()` are changed to `hasher: impl FnMut()`.

udoprog:
Closing now since things are moving and I don't have the time to do this.

PR #465, opened 2023-08-31
Fix two bugs in `clone_from` one of which is mine 😅 (was introduced in  #458)

JustForFun88:
My bad 😅, I totally forgot that we can modify table from another thread, which can result in an invalid `self` table in case if we have a panic during the cloning of elements (we will have an empty table with invalid control bytes to which we can get accessed via `RawIterRange` during parallel iteration or via the `find` function).
This pull request fixes my bug by partially reverting the old implementation of the `clone_from` function.

Also fixes leaking allocator when `self_.buckets() != source.buckets`. We used to rewrite the table (`&mut **self_ as *mut Self).write()`), forgetting that although we freed the old table memory, we forgot about the old allocator (**this behavior was found during testing**). Now, to allocate a new table, we use the old allocator, via `ptr::read(&self_.table.alloc)`.

The latter is worth thinking about. Maybe instead of using the old allocator, it's better to use the new one via `alloc.clone()` and just drop the old one?

**P.S.** Added tests for the cases described above.

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -2770,7 +2770,7 @@ impl<T: Clone, A: Allocator + Clone> Clone for RawTable<T, A> {
             unsafe {
                 // Make sure that if any panics occurs, we clear the table and
                 // leave it in an empty state.
-                let mut guard = guard(&mut *self, |self_| {
+                let mut self_ = guard(self, |self_| {
                     self_.clear_no_drop();
                 });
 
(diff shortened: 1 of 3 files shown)


src/raw/mod.rs
```diff
@@ -2781,31 +2781,34 @@ impl<T: Clone, A: Allocator + Clone> Clone for RawTable<T, A> {
...
+                    let alloc = ptr::read(&self_.table.alloc);
+                    (&mut **self_ as *mut Self).write(
                         // Avoid `Result::unwrap_or_else` because it bloats LLVM IR.
                         //
                         // SAFETY: This is safe as we are taking the size of an already allocated table
                         // and therefore сapacity overflow cannot occur, `self.table.buckets()` is power
                         // of two and all allocator errors will be caught inside `RawTableInner::new_uninitialized`.
                         match Self::new_uninitialized(
-                            self.table.alloc.clone(),
+                            alloc,
```
JustForFun88, line 2811: Maybe it's better to use `alloc.clone()` here?

src/raw/mod.rs
```diff
@@ -2817,9 +2820,11 @@ impl<T: Clone, A: Allocator + Clone> Clone for RawTable<T, A> {
...
                 // Cloning elements may fail (the clone function may panic), but the `ScopeGuard`
                 // inside the `clone_from_impl` function will take care of that, dropping all
-                // cloned elements if necessary. The `Drop` of `RawTable` takes care of the rest
-                // by freeing up the allocated memory.
-                self.clone_from_spec(source);
+                // cloned elements if necessary. Our `ScopeGuard` will clear the table.
+                self_.clone_from_spec(source);
+
+                // Disarm the scope guard if cloning was successful.
+                mem::forget(self_);
```
Amanieu, line 2827: I prefer `ScopeGuard::into_inner` for disarming a guard, it's more explicit.

src/raw/mod.rs
Amanieu: No, your current code looks fine and is arguably more correct than cloning the allocator.

In the long term, I would like to remove the `Clone` bound on the allocator. It is only required due to an implementation quirk where we have two `RawTable` with an allocator each when resizing a table.

JustForFun88: Interesting idea.  I got an idea how to do it.  We can simply split out the part without the allocator from RawTableInner, and shove all the code that does not depend on the allocator there.  This will allow us to have two RawTableInner_Inner and one allocator at the same time (well, in theory, I haven't thought deeply yet).

Can I try to implement this?

src/raw/mod.rs
JustForFun88: Done

PR #466, opened 2023-08-31
Add low-level `HashTable` API

Amanieu:
The primary use case for this type over `HashMap` or `HashSet` is to support types that do not implement the `Hash` and `Eq` traits, but instead require additional data not contained in the key itself to compute a hash and compare two elements for equality.

`HashTable` has some similarities with `RawTable`, but has a completely safe API. It is intended as a replacement for the existing raw entry API, with the intend of deprecating the latter and eventually removing it.

Examples of when this can be useful include:
- An `IndexMap` implementation where indices into a `Vec` are stored as elements in a `HashTable<usize>`. Hashing and comparing the elements requires indexing the associated `Vec` to get the actual value referred to by the index.
- Avoiding re-computing a hash when it is already known.
- Mutating the key of an element in a way that doesn't affect its hash.

To achieve this, `HashTable` methods that search for an element in the table require a hash value and equality function to be explicitly passed in as arguments. The method will then iterate over the elements with the given hash and call the equality function on each of them, until a match is found.

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -46,7 +46,7 @@ pub struct HashTable<T, A = Global>
 where
     A: Allocator,
 {
-    pub(crate) table: RawTable<T, A>,
+    pub(crate) raw: RawTable<T, A>,
 }
 
 impl<T> HashTable<T, Global> {
@@ -65,7 +65,7 @@ impl<T> HashTable<T, Global> {
     /// ```
     pub const fn new() -> Self {
         Self {
-            table: RawTable::new(),
+            raw: RawTable::new(),
         }
     }
 
@@ -84,7 +84,7 @@ impl<T> HashTable<T, Global> {
     /// ```
     pub fn with_capacity(capacity: usize) -> Self {
         Self {
-            table: RawTable::with_capacity(capacity),
+            raw: RawTable::with_capacity(capacity),
         }
     }
 }
@@ -133,7 +133,7 @@ where
     /// ```
     pub const fn new_in(alloc: A) -> Self {
         Self {
-            table: RawTable::new_in(alloc),
+            raw: RawTable::new_in(alloc),
         }
     }
 
@@ -182,13 +182,13 @@ where
     /// ```
     pub fn with_capacity_in(capacity: usize, alloc: A) -> Self {
         Self {
-            table: RawTable::with_capacity_in(capacity, alloc),
+            raw: RawTable::with_capacity_in(capacity, alloc),
         }
     }
 
     /// Returns a reference to the underlying allocator.
     pub fn allocator(&self) -> &A {
-        self.table.allocator()
+        self.raw.allocator()
     }
 
     /// Returns a reference to an entry in the table with the given hash and
@@ -222,7 +222,7 @@ where
     /// # }
     /// ```
     pub fn find(&self, hash: u64, eq: impl FnMut(&T) -> bool) -> Option<&T> {
-        self.table
+        self.raw
             .find(hash, eq)
             .map(|bucket| unsafe { bucket.as_ref() })
     }
@@ -263,7 +263,7 @@ where
     /// # }
     /// ```
     pub fn find_mut(&mut self, hash: u64, eq: impl FnMut(&T) -> bool) -> Option<&mut T> {
-        self.table
+        self.raw
             .find(hash, eq)
             .map(|bucket| unsafe { bucket.as_mut() })
     }
@@ -292,7 +292,7 @@ where
     /// let hasher = BuildHasherDefault::<AHasher>::default();
     /// let hasher = |val: &_| hasher.hash_one(val);
     /// table.insert_unchecked(hasher(&1), (1, "a"), |val| hasher(&val.0));
-    /// if let Some(entry) = table.find_entry(hasher(&1), |val| val.0 == 1) {
+    /// if let Ok(entry) = table.find_entry(hasher(&1), |val| val.0 == 1) {
     ///     entry.remove();
     /// }
     /// assert_eq!(table.find(hasher(&1), |val| val.0 == 1), None);
@@ -306,12 +306,15 @@ where
         &mut self,
         hash: u64,
         eq: impl FnMut(&T) -> bool,
-    ) -> Option<OccupiedEntry<'_, T, A>> {
-        self.table.find(hash, eq).map(|bucket| OccupiedEntry {
-            hash,
-            bucket,
-            table: &mut self.table,
-        })
+    ) -> Result<OccupiedEntry<'_, T, A>, &mut Self> {
+        match self.raw.find(hash, eq) {
+            Some(bucket) => Ok(OccupiedEntry {
+                hash,
+                bucket,
+                table: self,
+            }),
+            None => Err(self),
+        }
     }
 
     /// Returns an `Entry` for an entry in the table with the given hash
@@ -365,16 +368,16 @@ where
         eq: impl FnMut(&T) -> bool,
         hasher: impl Fn(&T) -> u64,
     ) -> Entry<'_, T, A> {
-        match self.table.find_or_find_insert_slot(hash, eq, hasher) {
+        match self.raw.find_or_find_insert_slot(hash, eq, hasher) {
             Ok(bucket) => Entry::Occupied(OccupiedEntry {
                 hash,
                 bucket,
-                table: &mut self.table,
+                table: self,
             }),
             Err(insert_slot) => Entry::Vacant(VacantEntry {
                 hash,
                 insert_slot,
-                table: &mut self.table,
+                table: self,
             }),
         }
     }
@@ -393,11 +396,11 @@ where
         value: T,
         hasher: impl Fn(&T) -> u64,
     ) -> OccupiedEntry<'_, T, A> {
-        let bucket = self.table.insert(hash, value, hasher);
+        let bucket = self.raw.insert(hash, value, hasher);
         OccupiedEntry {
             hash,
             bucket,
-            table: &mut self.table,
+            table: self,
         }
     }
 
@@ -425,7 +428,7 @@ where
     /// # }
     /// ```
     pub fn clear(&mut self) {
-        self.table.clear();
+        self.raw.clear();
     }
 
     /// Shrinks the capacity of the table as much as possible. It will drop
@@ -459,7 +462,7 @@ where
     /// # }
     /// ```
     pub fn shrink_to_fit(&mut self, hasher: impl Fn(&T) -> u64) {
-        self.table.shrink_to(self.len(), hasher)
+        self.raw.shrink_to(self.len(), hasher)
     }
 
     /// Shrinks the capacity of the table with a lower limit. It will drop
@@ -498,7 +501,7 @@ where
     /// # }
     /// ```
     pub fn shrink_to(&mut self, min_capacity: usize, hasher: impl Fn(&T) -> u64) {
-        self.table.shrink_to(min_capacity, hasher);
+        self.raw.shrink_to(min_capacity, hasher);
     }
 
     /// Reserves capacity for at least `additional` more elements to be inserted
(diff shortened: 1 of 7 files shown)


Zoxc:
I wonder if it would make sense to have `eq` and `hasher` utility methods for when you have `Eq` or `Hash` available. Passing different closures to methods is a bit of a code generation trap. It could help in some cases for that.

JustForFun88:
> I wonder if it would make sense to have `eq` and `hasher` utility methods for when you have `Eq` or `Hash` available. Passing different closures to methods is a bit of a code generation trap. It could help in some cases for that.

I think it's hardly possible to predict what type users will want to store in a table.  It can be `T`, `(T1, T2)`, `(T1, T2, ...Tn)`.

rgwh2q05:
In both `HashMap` and `HashTable`, only one entry "view" (`Entry` or `RawEntry`) structure can exist at a time. If you want to manipulate multiple entries at the same time, it seems like you have to use the raw `Bucket` and `RawTable` API.                                                                                       
                                                                                                                                                                                         
I think that this is not possible to fix though: it is similar to arrays where there can only be one mutable reference to an element at a time (although there is `split_off` in that case).

The use case I was wondering about from https://github.com/rust-lang/hashbrown/pull/450 was if you want modify an entry's key (possibly changing its hash), but also check if the new key is already in the map before actually?
1. If the check is done *before* looking up the entry for the old key, then the bucket for the new key is looked up twice.                                                                    
2. If the check is done *after* looking up the entry for the old key, then the bucket for the old key is looked up twice.

cuviper (commented):
I played with actually converting `indexmap` to this. Apart from `get_many_mut` noted below, I think I would also need something like `Occupied`/`VacantEntry::into_table(self) -> &mut HashTable`, because my `OccupiedEntry::remove` methods need to adjust other indices in the table.

src/table.rs
```diff
@@ -0,0 +1,1792 @@
...
+    /// # }
+    /// # fn main() {
+    /// #     #[cfg(feature = "nightly")]
+    /// #     test()
+    /// # }
+    /// ```
+    pub fn find_mut(&mut self, hash: u64, eq: impl FnMut(&T) -> bool) -> Option<&mut T> {
+        self.table
+            .find(hash, eq)
+            .map(|bucket| unsafe { bucket.as_mut() })
```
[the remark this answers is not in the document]
cuviper: See also https://github.com/bluss/indexmap/pull/275

src/table.rs
```diff
@@ -0,0 +1,1792 @@
...
+/// #     #[cfg(feature = "nightly")]
+/// #     test()
+/// # }
+/// ```
+pub struct OccupiedEntry<'a, T, A = Global>
+where
+    A: Allocator + Clone,
+{
+    hash: u64,
+    bucket: Bucket<T>,
```
cuviper, line 1300: The raw bucket prevents auto `Sync`, but that should be fine for `T: Sync`.

src/table.rs
```diff
@@ -0,0 +1,1792 @@
...
+    /// # fn main() {
+    /// #     #[cfg(feature = "nightly")]
+    /// #     test()
+    /// # }
+    /// ```
+    pub fn find_entry(
+        &mut self,
+        hash: u64,
+        eq: impl FnMut(&T) -> bool,
+    ) -> Option<OccupiedEntry<'_, T, A>> {
```
cuviper, line 309: Limitations in the borrow checker make it hard to use `find_entry` to implement my own non-growing `entry`, because if I return my wrapper around this `OccupiedEntry`, I can't have any _other_ branch that tries to store `&mut self` (to allow vacant insert).

Maybe instead:

```suggestion
    ) -> Result<OccupiedEntry<'_, T, A>, &mut Self> {
```

Amanieu: This is surprising. I'm pretty sure the borrow checker is able to tell that a value is no longer used in a branch. Can you show the cause that causes this?

src/table.rs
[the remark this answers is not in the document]
Amanieu: Sure, I'll add that.

src/table.rs
cuviper: Consider something like this:

```rust
use hashbrown::hash_table::{HashTable, OccupiedEntry};

struct Map(HashTable<(i32, String)>);

enum Entry<'a> {
    Occupied(OccupiedEntry<'a, (i32, String)>, i32),
    Vacant(&'a mut Map, i32)
}

impl Map {
    fn entry(&mut self, key: i32) -> Entry<'_> {
        if let Some(e) = self.0.find_entry(0, |&(x, _)| x == key) {
            Entry::Occupied(e, key)
        } else {
            Entry::Vacant(self, key)
        }
    }
}

fn main() {
    let mut map = Map(HashTable::new());
    let _ = map.entry(42);
}
```

```
error[E0499]: cannot borrow `*self` as mutable more than once at a time
  --> examples/mymap.rs:15:27
   |
11 |     fn entry(&mut self, key: i32) -> Entry<'_> {
   |              - let's call the lifetime of this reference `'1`
12 |         if let Some(e) = self.0.find_entry(0, |&(x, _)| x == key) {
   |                          ---------------------------------------- first mutable borrow occurs here
13 |             Entry::Occupied(e, key)
   |             ----------------------- returning this value requires that `self.0` is borrowed for `'1`
14 |         } else {
15 |             Entry::Vacant(self, key)
   |                           ^^^^ second mutable borrow occurs here
```

This kind of conditional return is a known limitation of NLL, which Polonius may eventually resolve:
https://smallcultfollowing.com/babysteps/blog/2018/06/15/mir-based-borrow-check-nll-status-update/#polonius

cuviper: The known workaround is to use a double-lookup, like `if self.contains(..) { let e = self.0.find_entry(..).unwrap(); .. } else { .. }`, so NLL can split the return lifetime.

JustForFun88: May be you can use `HashTable::entry` directly?

cuviper: I don't want the growth, as `entry` documents:

```rust
    /// This method may grow the table in preparation for an insertion. Call
    /// [`HashTable::find_entry`] if this is undesirable.
```

Zoxc:
It would be useful to have fallible variants of `entry` and `insert_unchecked`. They can have better code generation as they don't need to panic.

Amanieu:
> It would be useful to have fallible variants of `entry` and `insert_unchecked`. They can have better code generation as they don't need to panic.

You mean fallible allocations? Isn't that already addressed by `try_reserve`?

src/table.rs
Amanieu: I made the change, but I'm not too happy about it. In particular, `Self` makes for a very poor error type.

cuviper: Alternatively, `VacantEntry` could hold an `Option<InsertSlot>` instead, possibly even setting that from `find_entry` if it can do so without growing. But the rest of the `Entry` and `VacantEntry` methods would need to add parameters to deal with the possibility of growing later.

Or we could add separate `Entry` and `VacantEntry` types for this case, while `OccupiedEntry` can remain shared.

src/table.rs
```diff
@@ -0,0 +1,1949 @@
...
+                .insert_in_slot(self.hash, self.insert_slot, value)
+        };
+        OccupiedEntry {
+            hash: self.hash,
+            bucket,
+            table: self.table,
+        }
+    }
+
+    /// Converts the OccupiedEntry into a mutable reference to the underlying
```
cuviper, line 1759:
```suggestion
    /// Converts the VacantEntry into a mutable reference to the underlying
```

Zoxc:
> You mean fallible allocations? Isn't that already addressed by `try_reserve`?

I mean variants which won't grow, like `Vec::push_within_capacity`.

src/table.rs
cuviper: Here's a separate `FindEntry` with `AbsentEntry` (not firm on that name):
https://github.com/Amanieu/hashbrown/compare/hashtable...cuviper:hashbrown:hashtable-find_entry

Amanieu:
> I mean variants which won't grow, like `Vec::push_within_capacity`.

I could see an `insert_unchecked_within_capacity`, but this doesn't make sense for `entry` since the first thing is does is `reserve(1)`. I'll leave it for a future PR though.

src/table.rs
Amanieu: I think that the best thing for now is to make this `Result<OccupiedEntry, AbsentEntry>`, with the understanding that once NLL is fixed this will be changed to `Option<OccupiedEntry>`.

In the case of `find_entry`, we specifically want to use `find` instead of `find_or_find_insert_slot`, since the former is slightly faster.

Zoxc:
It does make sense for `entry_within_capacity` too. It would just return `None` if there isn't capacity for 1 insertion.

src/table.rs
```diff
@@ -0,0 +1,2013 @@
...
+                table: self,
+            }),
+        }
+    }
+
+    /// Inserts an element into the `HashTable` with the given hash value, but
+    /// without checking whether an equivalent element already exists within the
+    /// table.
+    ///
+    /// This is
```
Zoxc, line 389: Some extra words here.

src/table.rs
```diff
@@ -0,0 +1,2013 @@
...
+    ///
+    /// `hasher` is called if entries need to be moved or copied to a new table.
+    /// This must return the same hash value that each entry was inserted with.
+    pub fn insert_unchecked(
```
Zoxc, line 393: I do prefer the `insert_unique` naming as the `unchecked` postfix makes me think it's memory unsafe.

src/table.rs
```diff
@@ -0,0 +1,2013 @@
...
+    /// #     test()
+    /// # }
+    /// ```
+    pub fn entry(
+        &mut self,
+        hash: u64,
+        eq: impl FnMut(&T) -> bool,
+        hasher: impl Fn(&T) -> u64,
+    ) -> Entry<'_, T, A> {
+        match self.raw.find_or_find_insert_slot(hash, eq, hasher) {
```
JustForFun88, line 371: I'm playing with HashTable and came across a small problem. Can we directly provide some version of the `find_or_find_insert_slot` (***checked insert***) function?
It's very annoying that every time I need to insert a value I have to use the entry syntax (which is not that cheap) or use `find + insert_unchecked`, which is slow.

It can be something like this one:
```rust
pub fn insert<V>(
    &mut self,
    hash: u64,
    value: T,
    mut eq: impl FnMut(&T, &T) -> bool,
    hasher: impl Fn(&T) -> u64,
    replace: impl FnOnce(&mut T, T) -> V,
) -> Option<V> {
    match self
        .raw
        .find_or_find_insert_slot(hash, |found_val| eq(found_val, &value), hasher)
    {
        Ok(bucket) => Some(replace(unsafe { &mut bucket.as_mut() }, value)),
        Err(slot) => {
            unsafe {
                self.raw.insert_in_slot(hash, slot, value);
            }
            None
        }
    }
}
```
Then it can be used like:
```rust

pub struct NewMap<K, V, S = DefaultHashBuilder, A: Allocator = Global> {
    pub(crate) hash_builder: S,
    pub(crate) table: HashTable<(K, V), A>,
}

impl<K, V, S, A> NewMap<K, V, S, A>
where
    K: Eq + core::hash::Hash,
    S: core::hash::BuildHasher,
    A: Allocator,
{
    pub fn insert(&mut self, k: K, v: V) -> Option<V> {
        let hash = make_hash::<K, S>(&self.hash_builder, &k);
        let hasher = make_hasher::<_, V, S>(&self.hash_builder);

        self.table.insert(
            hash,
            (k, v),
            |found, new| found.0 == new.0,
            hasher,
            |(_, val_ref), (_, val)| core::mem::replace(val_ref, val),
        )
    }
}
```

Amanieu: `entry` already maps directly to `find_or_find_insert_slot`. You can then use `Entry::insert` to unconditionally overwrite an existing value, or `Entry::or_insert` to only insert a new value if an old one doesn't already exist.

PR #472, opened 2023-10-02
Move `RawTableInner` and `TableLayout` to separate module

JustForFun88:
All methods of the RawTableInner structure assume one important detail that is not described in the documentation: the `ctrl` and `bucket_mask` fields must remain unchanged throughout the life of the current `RawTableInner` instance. And this is without taking into account the specific requirements for the `ctrl` and `bucket_mask` fields.

Describing these requirements in each individual method will only lead to needlessly bloated documentation, since rustс has no way to block changing the individual fields of structure, and there is no way to check that they have not been changed.

To provide the minimum possible protection against changing the fields, we can only declare structures in a separate module, making necessary fields private.

Since the functions for getting the `bucket_mask` are marked as `#[inline(always)]` this should not affect compilation speed or performance.

But in general, this pull request just improves safety a little, **fell free to close it if you don't like this idea**.

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -41,6 +41,9 @@ mod bitmask;
 use self::bitmask::BitMaskIter;
 use self::imp::Group;
 
+mod private_table_inner;
+use private_table_inner::{RawTableInner, TableLayout};
+
 // Branch prediction hint. This is currently only available on nightly but it
 // consistently improves performance by 10-15%.
 #[cfg(not(feature = "nightly"))]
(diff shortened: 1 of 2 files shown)


Amanieu:
I'm not really a fan of splitting up code just for the safe of safety, but in this case the `raw/mod.rs` file was already too large to be properly manageable, so it's good to split it up.

I am somewhat concerned about the compile-time impact of so many additional inline functions. Could you do a rustc perf run (using a PR on rust-lang/rust) to check that no compile-time regressions will occur once the standard library is updated to use the latest hashbrown?

Issue #473, opened 2023-10-18
ahash shouldn't be the default Hasher


jedisct1:
See https://github.com/tkaitchuck/aHash/issues/163

It's also currently completely broken on ARM: https://github.com/tkaitchuck/aHash/issues/166

jestarray:
Curious, what is the best hasher if one only cares about perf?

thomcc:
It's an unanswerable question. You can make an arbitrarily fast hasher by reducing the quality of the output. For example, only hash the first 16 bytes of input strings, or even just return a constant value. Both of those are legal Hash implementations.

mati865:
aHash is no longer the default.

PR #479, opened 2023-10-27
bump ahash version 0.8.0 -> 0.8.6

initiativist:
ahash 0.8.0 was yanked.

should also fix the issues mentioned in [hashbrown/issues/473](https://github.com/rust-lang/hashbrown/issues/473)

I think I ran all the necessary tests so let me know if I need to do more.


Amanieu:
Thanks!

@bors r+

PR #480, opened 2023-11-06
Specialize `fold` implementation of iterators

a1phyr:
This provides 5-8% iteration speedups when using internal iteration.
Similar work could probably be done with `try_fold` but this requires `nightly` to implement.

Maybe this is worth running a rustc perf run with this before merge ?

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -50,12 +50,35 @@ use self::imp::Group;
 
 // Branch prediction hint. This is currently only available on nightly but it
 // consistently improves performance by 10-15%.
-#[cfg(not(feature = "nightly"))]
-use core::convert::identity as likely;
-#[cfg(not(feature = "nightly"))]
-use core::convert::identity as unlikely;
+// #[cfg(not(feature = "nightly"))]
+// use core::convert::identity as likely;
+// #[cfg(not(feature = "nightly"))]
+// use core::convert::identity as unlikely;
 #[cfg(feature = "nightly")]
 use core::intrinsics::{likely, unlikely};
+
+#[cfg(not(feature = "nightly"))]
+#[inline]
+#[cold]
+fn cold() {}
+
+#[cfg(not(feature = "nightly"))]
+#[inline]
+fn likely(b: bool) -> bool {
+    if !b {
+        cold()
+    }
+    b
+}
+#[cfg(not(feature = "nightly"))]
+#[cold]
+#[inline]
+fn unlikely(b: bool) -> bool {
+    if b {
+        cold()
+    }
+    b
+}
 
 // Use strict provenance functions if available.
 #[cfg(feature = "nightly")]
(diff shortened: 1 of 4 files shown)


src/raw/mod.rs
```diff
@@ -3846,6 +3869,41 @@ impl<T> RawIterRange<T> {
...
+    where
+        F: FnMut(B, Bucket<T>) -> B,
+    {
+        if n == 0 {
+            return acc;
+        }
+
+        loop {
+            while let Some(index) = self.current_group.next() {
+                debug_assert!(n != 0);
```
JustForFun88, line 3887:
```suggestion
                debug_assert!(n != 0);
                // The returned `index` will always be in the range `0..Group::WIDTH`,
                // so that calling `self.data.next_n(index)` is safe (see detailed explanation below).
```

src/raw/mod.rs
```diff
@@ -3846,6 +3869,41 @@ impl<T> RawIterRange<T> {
...
+
+            if n == 0 {
+                return acc;
+            }
+
+            // We might read past self.end up to the next group boundary,
+            // but this is fine because it only occurs on tables smaller
+            // than the group size where the trailing control bytes are all
+            // EMPTY. On larger tables self.end is guaranteed to be aligned
+            // to the group size (since tables are power-of-two sized).
```
JustForFun88, line 3901:
```suggestion        
            // SAFETY: The caller of this function ensures that:
            //
            // 1. The provided `n` value matches the actual number of items in the table;
            // 2. The table is alive and did not moved.
            //
            // Taking the above into account, we always stay within the bounds, because:
            //
            // 1. For tables smaller than the group width (self.buckets() <= Group::WIDTH),
            //    we will never end up in the given branch, since we should have already
            //    yielded all the elements of the table.
            //
            // 2. For tables larger than the group width. The the number of buckets is a
            //    power of two (2 ^ n), Group::WIDTH is also power of two (2 ^ k). Sinse
            //    `(2 ^ n) > (2 ^ k)`, than `(2 ^ n) % (2 ^ k) = 0`. As we start from the
            //    the start of the array of control bytes, and never try to iterate after
            //    getting all the elements, the last `self.current_group` will read bytes
            //    from the `self.buckets() - Group::WIDTH` index.  We know also that
            //    `self.current_group.next()` will always retun indices within the range
            //    `0..Group::WIDTH`.
            //
            //    Knowing all of the above and taking into account that we are synchronizing
            //    the `self.data` index with the index we used to read the `self.current_group`,
            //    the subsequent `self.data.next_n(index)` will always return a bucket with
            //    an index number less than `self.buckets()`.
            //
            //    The last `self.next_ctrl`, whose index would be `self.buckets()`, will never
            //    actually be read, since we should have already yielded all the elements of
            //    the table.
```

src/table.rs
```diff
@@ -1894,12 +1905,23 @@ impl<'a, T> Iterator for IterMut<'a, T> {
     type Item = &'a mut T;
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.inner.next().map(|bucket| unsafe { bucket.as_mut() })
+        // Avoid `Option::map` because it bloats LLVM IR.
+        let bucket = self.inner.next()?;
+        Some(unsafe { bucket.as_mut() })
```
JustForFun88, line 1910: Maybe it's better to use `mach` directly? This will take the same two lines

src/table.rs
```diff
@@ -1861,12 +1861,23 @@ impl<'a, T> Iterator for Iter<'a, T> {
     type Item = &'a T;
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.inner.next().map(|bucket| unsafe { bucket.as_ref() })
+        // Avoid `Option::map` because it bloats LLVM IR.
+        let bucket = self.inner.next()?;
+        Some(unsafe { bucket.as_ref() })
```
JustForFun88, line 1866: Maybe it's better to use `mach` directly? This will take the same two lines

src/raw/mod.rs
```diff
@@ -50,13 +50,36 @@ use self::imp::Group;
...
+#[cfg(not(feature = "nightly"))]
+#[cold]
+#[inline]
+fn unlikely(b: bool) -> bool {
+    if b {
+        cold()
+    }
+    b
+}
+
```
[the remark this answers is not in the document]
a1phyr: Whoops didn't want to let this through

PR #488, opened 2023-12-06
Add gxhash in bench

ogxd:
Not meant to be merged


ogxd:
#487

clarfonthey:
Going to close this since when updating the benchmarks I think we're going to be incorporating a lot of the various hash function crates, and we have an open issue for gxhash.

PR #491, opened 2023-12-11
Add function to support pre-hashed keys

joaovbsevero:
## Motivation

I'm working on a monad type with expensive method resolution (`hash` being one of them). To speed up the hashmap insertion I pre-hash my inner value when it is known, build my monad and insert into the HashMap using this new method with the pre-generated hash. It gave me a speed up of about 15%.

## Implementation

The implementation is straightforward, using the same code path as `insert` but with a given hash.
The use of `mem::swap` is necessary to ensure that the old value (if any) is dropped and proved to be faster on my benchmarks than `ptr::write`.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1824,6 +1824,53 @@ where
         (k_ref, v_ref)
     }
 
+    /// Inserts a key-value pair into the map using pre-generated key hash value.
+    ///
+    /// This method is intended to be used by monads that overwrite the default
+    /// hash method of the inner type.
+    ///
+    /// No value is returned from this function, this functions intent is to be
+    /// a fast insertion path for container types.
+    ///
+    /// [`None`]: https://doc.rust-lang.org/std/option/enum.Option.html#variant.None
+    /// [`std::collections`]: https://doc.rust-lang.org/std/collections/index.html
+    /// [module-level documentation]: https://doc.rust-lang.org/std/collections/index.html#insert-and-complex-keys
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use hashbrown::HashMap;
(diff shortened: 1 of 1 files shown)


Amanieu:
Have you considered using the `HashTable` type instead? It is specifically designed for this kind of use case.

There is also the raw entry API which provides similar functionality, but that is deprecated in favor of `HashTable`.

jojo0severo:
> Have you considered using the `HashTable` type instead? It is specifically designed for this kind of use case.
> 
> There is also the raw entry API which provides similar functionality, but that is deprecated in favor of `HashTable`.

I found the HashTable to low-level for what I'm doing, this Pull Request is just a suggestion, if considered not actually usefull I will just use my forked version keeping up-to-date with the library.

Amanieu:
This method is unsuitable because we don't want to expose custom hash values in `HashMap`/`HashSet`. This could cause unsoundness because it would allow constructing a `HashMap<usize, usize>` which has inconsistent behavior. This is relevant because no types with custom `Hash` impl are used: you should be able to trust that maps of standard types behave predictably.

This is why all functionality related to custom hashing was moved to `HashTable`. Even though it has a low-level API, you can easily build a wrapper on top of it for your use case.

jojo0severo:
> This method is unsuitable because we don't want to expose custom hash values in `HashMap`/`HashSet`. This could cause unsoundness because it would allow constructing a `HashMap<usize, usize>` which has inconsistent behavior. This is relevant because no types with custom `Hash` impl are used: you should be able to trust that maps of standard types behave predictably.
> 
> This is why all functionality related to custom hashing was moved to `HashTable`. Even though it has a low-level API, you can easily build a wrapper on top of it for your use case.

No problem, thanks for spending time looking into the PR 😄

PR #501, opened 2024-02-10
Use a bit less NonNull::new_unchecked

saethlin:
Using `NonNull::cast` instead of `NonNull::new_unchecked` is a bit less unsafe code, and it may provide a small improvement to compile times since https://github.com/rust-lang/rust/pull/120594


saethlin:
This project's CI is broken due to a dependency on a previous version of itself. Well ain't that something.

Amanieu:
I fixed the CI separately in #502, can you rebase?

saethlin:
Rebased!

PR #507, opened 2024-02-22
Introduce Swiss Cheese benchmark

matthieu-m:
* Changes:

- Introduce new swiss_cheese.rs benchmark.

* Motivation:

The current design of RawTable leaves tombstones behind to indicate deleted elements, going from tightly packed elements to a much more sparse design as more and more elements are deleted -- until more elements are inserted again, at least.

This creates a "swiss cheese" situation, with a decreasing elements density over time. This situation is expected to lead to a higher number of cache misses as this density decreases.

The swiss_cheese.rs benchmark suite aims at measuring the impact of this effect on performance.

### DIFF
--- a/benches/swiss_cheese.rs
+++ b/benches/swiss_cheese.rs
@@ -1,0 +1,785 @@
+//! Compares look-up performance in a "hollowed out" hash-map.
+//!
+//! The layout of a Swiss Map leads to a number of potential cache-misses:
+//!
+//! -   The hash-map itself is assumed to be in cache, in a hot loop.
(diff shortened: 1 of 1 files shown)


Amanieu:
CI should be fixed now, can you rebase? I am curious to see what kind of results you are getting with this benchmark.

Amanieu:
With that said, I don't think it makes sense to merge this benchmark. It is useful as an experiment, but it's less useful for measuring the actual performance of the table.

matthieu-m:
I don't think it makes much sense either.

Making a PR was the easiest to share both code and result, but it may be a bit too specific.

PR #518, opened 2024-04-09
Add an assertion to `HashSet::get_or_insert_with`

cuviper:
Since the Rust libs-api team considers it problematic for `HashSet<T>`
to be unreliable with well-behaved `T` (not user-controlled), we should
not allow mismatches to be inserted through `get_or_insert_with`.

### DIFF
--- a/src/set.rs
+++ b/src/set.rs
@@ -955,6 +955,10 @@ where
     /// Inserts a value computed from `f` into the set if the given `value` is
     /// not present, then returns a reference to the value in the set.
     ///
+    /// # Panics
+    ///
+    /// Will panic if `f`'s return value is not equivalent to the query `value`.
+    ///
     /// # Examples
     ///
     /// ```
@@ -970,6 +974,13 @@ where
     /// }
     /// assert_eq!(set.len(), 4); // a new "fish" was inserted
     /// ```
+    ///
+    /// The following example will panic because the new value doesn't match.
+    ///
+    /// ```should_panic
+    /// let mut set = hashbrown::HashSet::new();
+    /// set.get_or_insert_with("rust", |_| String::new());
+    /// ```
     #[cfg_attr(feature = "inline-more", inline)]
     pub fn get_or_insert_with<Q: ?Sized, F>(&mut self, value: &Q, f: F) -> &T
     where
@@ -981,7 +992,11 @@ where
         self.map
             .raw_entry_mut()
             .from_key(value)
-            .or_insert_with(|| (f(value), ()))
+            .or_insert_with(|| {
+                let new = f(value);
+                assert!(value.equivalent(&new), "new value is not equivalent");
+                (new, ())
+            })
             .0
     }
 


cuviper:
I didn't notice #400 before, but I think that's doing a lot more than is really needed for the immediate problem. For instance, I don't think it's important to compare the hashes, because that calculation is not under user control here.

JustForFun88:
> I didn't notice #400 before, but I think that's doing a lot more than is really needed for the immediate problem. For instance, I don't think it's important to compare the hashes, because that calculation is not under user control here. And value equality already implies hash equality, especially for known-good non-user types.

I can correct the code if necessary.

JustForFun88:
> I didn't notice #400 before, but I think that's doing a lot more than is really needed for the immediate problem. For instance, I don't think it's important to compare the hashes, because that calculation is not under user control here. And value equality already implies hash equality, especially for known-good non-user types.

I removed unnecessary code. Now the comparison is carried out by values. I don’t think this will affect performance. I use the `from_key_hashed_nocheck` method, so the hash is calculated only once, where the old version of the function calculated the hash twice (separately inside `from_key` and inside `or_insert_with` methods)

Amanieu:
For some background, this came up while I was trying to remove `RawEntry` from hashbrown (by moving it under an off-by-default cargo feature). The only internal uses of the raw entry API were from the `HashSet::get_or_insert*` methods. I can rewrite `get_or_insert` and `get_or_insert_owned` to use `Entry` and `EntryRef` (with some minor changes). But for `get_or_insert_with` I would have to fall back to the `RawTable` API, which I would rather avoid: `HashSet` should only expose operations that can also be performed on a normal `HashMap`.

JustForFun88:
> I can rewrite `get_or_insert` and `get_or_insert_owned` to use `Entry` and `EntryRef` (with some minor changes).

In fact, without unsafe code (or adding a small method to `OccupiedEntry`) you won't be able to do this.  When calling `entry.key()` the compiler will complain that it **cannot return a reference to a local variable**

JustForFun88:
> I can rewrite `get_or_insert` and `get_or_insert_owned` to use `Entry` and `EntryRef` (with some minor changes). But for `get_or_insert_with` I would have to fall back to the `RawTable` API, which I would rather avoid: `HashSet` should only expose operations that can also be performed on a normal `HashMap`.

Well, `get_or_insert` can be rewritten to `Entry` by slightly changing `OccupiedEntry::key`, namely by adding the lifetime `'a` to the key `K` (`pub fn key(&self) -> &'a K`).

But `get_or_insert_owned` is not so easy to solve. First, you need to use `EntryRef`, not `Entry`. Secondly, all existing `VacantEntryRef` methods work with `K: From<&Q>`, and not with `Q: ToOwned<Owned = K>`.

You need to either rewrite `get_or_insert_owned`, or change the signatures of `VacantEntryRef` to `ToOwned`, or add a new method for VacantEntryRef (ala from_owned_with).

cuviper:
> Well, `get_or_insert` can be rewritten to `Entry` by slightly changing `OccupiedEntry::key`, namely by adding the lifetime `'a` to the key `K` (`pub fn key(&self) -> &'a K`).

That would be **unsound**, because a user could follow that with a `remove` call, making that reference dangling. That's why the raw entry has `fn into_key(self) -> &'a mut K`, _consuming_ the entry so it can no longer be used.

JustForFun88:
> That would be **unsound**, because a user could follow that with a `remove` call, making that reference dangling. That's why the raw entry has `fn into_key(self) -> &'a mut K`, _consuming_ the entry so it can no longer be used.

That's right, I forgot about the `remove`. Well, in any case, we will have to expand the existing HashMap API, if we want remove `RawEntry`

cuviper:
Subsumed by #555.

PR #519, opened 2024-04-11
Remove unnecessary `add` in `clone_from_impl`

JustForFun88:
The `guard.0 = index + 1` in the implementation of `clone_from _impl` is completely unnecessary, since an invalid index can be set in advance.


Amanieu:
The `+ 1` is implicitly calculated by the iterator anyways, so it doesn't cost anything extra in optimized builds.

I prefer the original form of the code since `guard.0` clearly indicates the point that separates the initialized part of the array vs the uninitialized part. This also avoids the need to special-case `usize::MAX`.

clarfonthey:
Going to close for the combined reasoning from Amanieu + the merge conflicts.

PR #495, opened 2024-02-02
Make HashSet::insert return OccupiedEntry

SUPERCILEX:
See https://github.com/rust-lang/rust/pull/120077#discussion_r1475708665

PR #521, opened 2024-04-28
Release 0.14.5

Amanieu:
This temporarily reverts (#495) since that is a breaking change that will be included in v0.15.0.


Amanieu:
Oops, I just realized I forgot to bump the version in Cargo.toml, let's try this again...

Issue #489, opened 2023-12-07
Feature: increase capacity according to the actual size returned by the allocator


morrisonlevi:
I have some similar allocators:

1. One gives out whole pages only. On Linux this is commonly 4KiB and on Apple silicon this is commonly 16KiB.
2. One which has one big static chunk. That's it, no subdivision (think no-std).

They are similar in that if I call `HashMap::with_capacity_in` or other function with a capacity of `c`:

- Knowing that `HashMap` can and will often round up the capacity to a bigger size.
- Knowing that  allocators in general are allowed to over-allocate and often will.
- Knowing that my allocators in particular are going to significantly over-allocate the number of bytes.

I would _hope_ that `HashMap` would attempt to use this extra allocation space, but it never inspects the returned size.

Are there any reservations about doing this? It will add cost on every allocation, but that's about the only reason I could think of. The benefit is that for some allocators, they will better utilize the allocation, and for some allocators this would be an extreme improvement. With that said, I am not a hash table expert in general, and certainly a noob in `hash_brown`'s internals, so there may be things I am unaware of.

Amanieu:
I'm open to the idea, but of course we'll have to see how much this impacts common use cases (the global allocator). I don't expect much impact, so I'm pretty optimistic.

Do give it a try and see how it goes.

morrisonlevi:
@Amanieu I made a PR for this: #523. It depends on #524. Could you please review it when you have the time? These are my first contributions to rust-lang/* and I welcome feedback, including nits. In particular, I'm not sure how ZSTs are expected to behave. There is a ZST in the test suite for `test_set::collect` which passes, but I'm not sure what I've done is necessarily logical.

morrisonlevi:
Thanks, @Amanieu, for reviewing and merging #615. Now that it has merged, I have updated #523. When you get time, can you review that PR to provide me with the next steps? Thanks.

PR #523, opened 2024-05-04
feat: recognize and use over sized allocations

morrisonlevi:
Allocators are allowed to return a larger memory chunk than was asked for. If the amount extra is large enough, then the hash table can use the extra space. The Global allocator will not hit this path, because it won't over-size enough to matter, but custom allocators may. An example of an allocator which allocates full system pages is included in the test suite (UNIX only because it uses `mmap`).

This implements #489. ~This relies on PR #524 to increase the minimum number of buckets for certain small types, which in turn constrains the domain of `maximum_buckets_in` so that the alignment can be ignored.~ Merged. 

~I haven't done any performance testing. Since this is on the slow path of making a new allocation, the feature should be doable without too much concern about overhead.~ There is a benchmark now you can run yourself, you can see the [numbers on my Apple M1 machine here](https://github.com/rust-lang/hashbrown/pull/523#issuecomment-3315091391). Since then, I added a fast-path shortcut which cuts this down when the system allocator is used (or when any allocator returns what was asked for).

I am definitely not an expert in swiss tables. Feedback is very welcome, even nitpicking.

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -1442,6 +1442,40 @@ impl RawTableInner {
     }
 }
 
+/// Find the previous power of 2. If it's already a power of 2, it's unchanged.
+/// Passing zero is undefined behavior.
+pub(crate) fn prev_pow2(z: usize) -> usize {
+    let shift = mem::size_of::<usize>() * 8 - 1;
+    1 << (shift - (z.leading_zeros() as usize))
+}
+
+fn maximum_buckets_in(
+    allocation_size: usize,
+    table_layout: TableLayout,
+    group_width: usize,
+) -> usize {
+    // Given an equation like:
+    //   z >= x * y + x + g
+    // x can be maximized by doing:
+    //   x = (z - g) / (y + 1)
+    // If you squint:
+    //   x is the number of buckets
+    //   y is the table_layout.size
+    //   z is the size of the allocation
+    //   g is the group width
+    // But this is ignoring the padding needed for ctrl_align.
+    // If we remember these restrictions:
+    //   x is always a power of 2
+    //   Layout size for T must always be a multiple of T
+    // Then the alignment can be ignored if we add the constraint:
+    //   x * y >= table_layout.ctrl_align
+    // This is taken care of by `capacity_to_buckets`.
+    let numerator = allocation_size - group_width;
+    let denominator = table_layout.size + 1; // todo: ZSTs?
+    let quotient = numerator / denominator;
+    prev_pow2(quotient)
+}
+
 impl RawTableInner {
     /// Allocates a new [`RawTableInner`] with the given number of buckets.
     /// The control bytes and buckets are left uninitialized.
@@ -1459,7 +1493,7 @@ impl RawTableInner {
     unsafe fn new_uninitialized<A>(
         alloc: &A,
         table_layout: TableLayout,
-        buckets: usize,
+        mut buckets: usize,
         fallibility: Fallibility,
     ) -> Result<Self, TryReserveError>
     where
@@ -1468,13 +1502,29 @@ impl RawTableInner {
         debug_assert!(buckets.is_power_of_two());
 
         // Avoid `Option::ok_or_else` because it bloats LLVM IR.
-        let (layout, ctrl_offset) = match table_layout.calculate_layout_for(buckets) {
+        let (layout, mut ctrl_offset) = match table_layout.calculate_layout_for(buckets) {
             Some(lco) => lco,
             None => return Err(fallibility.capacity_overflow()),
         };
 
         let ptr: NonNull<u8> = match do_alloc(alloc, layout) {
-            Ok(block) => block.cast(),
+            Ok(block) => {
+                // Utilize over-sized allocations.
+                let x = maximum_buckets_in(block.len(), table_layout, Group::WIDTH);
+                debug_assert!(x >= buckets);
+                // Calculate the new ctrl_offset.
+                let (_oversized_layout, oversized_ctrl_offset) =
+                    match table_layout.calculate_layout_for(x) {
+                        Some(lco) => lco,
+                        None => unsafe { hint::unreachable_unchecked() },
+                    };
+                debug_assert!(_oversized_layout.size() <= block.len());
+                debug_assert!(oversized_ctrl_offset >= ctrl_offset);
+                ctrl_offset = oversized_ctrl_offset;
+                buckets = x;
+
+                block.cast()
+            }
             Err(_) => return Err(fallibility.alloc_err(layout)),
         };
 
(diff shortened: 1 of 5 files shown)


morrisonlevi:
I've figured out the rough equation to do it without a loop:

```rs
fn maximum_buckets_in(allocation_size: usize, table_layout: TableLayout) -> usize {
    // Given an equation like:
    //   z >= x * y + x
    // x can be maximized by doing:
    //   x = z / (y + 1)
    // If you squint:
    //   z is the size of the allocation
    //   y is the table_layout.size
    //   x is the number of buckets
    // But there are details like x needing to be a power of 2,
    // and there are some extra bytes mixed in (a possible
    // rounding up for table_layout.align, and Group::WIDTH).
    /// todo: how do I factor in the ctrl_align?
    let z = allocation_size - Group::WIDTH;
    let y_plus_1 = table_layout.size + 1;
    prev_pow2(z / y_plus_1)
}
```

I'm not quite sure about the `table_layout.ctrl_align`. I need to think about that more.

morrisonlevi:
So far I haven't been able to make it fail on real-world things. If you can make the `ctrl_align` quite a bit larger than the size, then the alignment becomes significant. It usually doesn't matter because:
1. Types are multiples of their align. This rules out things like `size: 128, ctrl_align: 4096`.
2. Bucket sizes are always powers of 2, and we never allocate less than 4 buckets.

Adjacent to this, I realize that on certain platforms, the minimum buckets being `4` is actually a smidge inefficient:

```
buckets: 4, TableLayout { size: 1, ctrl_align: 16 } leads to 36 bytes allocated
buckets: 8, TableLayout { size: 1, ctrl_align: 16 } leads to 40 bytes allocated
^-- using +4 more bytes more than doubles capacity (3 to 7)
```

Note that for the cost of `4` extra bytes, each of these becomes

morrisonlevi:
I found an issue with very `TableLayout` sizes with large `ctrl_align`s. The optimization in #524 also corrects the domain of `maximum_buckets_in` because `buckets * table_layout.size` (or `x * y` above) will now be at least `ctrl_align`, so we can safely ignore it in the equation.

morrisonlevi:
I have rebased on the latest master and got the tests working.

Amanieu:
Could you write a quick benchmark to measure the added cost of allocating a new `HashMap`? Specifically just measuring `HashMap::with_capacity(X)` for different X values to see how much of an impact this has.

morrisonlevi:
@Amanieu, I rebased my work, added a benchmark, and squashed my changes. This way anyone can check out the branch and compare to master by running:

```sh
# test this branch
cargo +nightly bench --bench with_capacity 

# revert the changes
git revert HEAD --no-edit

# test master
cargo +nightly bench --bench with_capacity 

# remove the revert commit to put us back to where we started
git reset --hard HEAD~1
```

This doesn't actually return the oversized check, since it's just using the system allocator. So the overhead being measured here is "the code performs the oversized check but the check doesn't result in a larger hashtable." Here are the formatted side-by-side results on my macOS M1:

Capacity | With feature (ns/iter) | Reverted (master) (ns/iter) | Overhead vs master
-- | -- | -- | --
0 | 3.50 ± 0.23 | 3.54 ± 0.05 | -1.1%
1 | 24.51 ± 0.60 | 21.18 ± 2.01 | +15.7%
3 | 24.62 ± 0.45 | 21.38 ± 1.67 | +15.2%
7 | 24.29 ± 0.71 | 21.83 ± 1.03 | +11.3%
8 | 51.60 ± 3.30 | 48.44 ± 3.45 | +6.5%
16 | 53.69 ± 2.72 | 51.28 ± 2.11 | +4.7%
32 | 20.60 ± 0.96 | 18.62 ± 0.68 | +10.6%
64 | 20.43 ± 0.44 | 19.18 ± 0.35 | +6.5%
128 | 21.88 ± 0.40 | 20.57 ± 0.37 | +6.4%
256 | 24.51 ± 1.71 | 25.11 ± 0.47 | -2.4%
512 | 31.11 ± 0.44 | 31.26 ± 0.47 | -0.5%
1,024 | 44.89 ± 4.70 | 44.21 ± 0.60 | +1.5%
4,096 | 156.25 ± 36.71 | 143.55 ± 40.47 | +8.9%
16,384 | 688.40 ± 30.35 | 685.02 ± 34.83 | +0.5%
65,536 | 14,748.30 ± 1,930.86 | 14,053.07 ± 1,588.73 | +5.0%

You can see results range from -2.4% to +15.7%. The results jive with what I would have suspected: the overhead is a bit higher on the smaller sizes because the overhead of the oversized check is roughly the same no matter the size, and for the larger sizes the other parts will dominate.

Note that for the allocators I personally have in mind, these sizes are fairly realistic. For example:
 1. Use a fixed-size of pre-allocated memory such as static memory (e.g. 64KiB).
 2. Round up to whole memory pages, which range in size from 4KiB to 16KiB depending on OS before including large pages.
 
I can check more sizes if you are interested, particular non-power-of-two sizes.

src/raw/mod.rs
```diff
@@ -1442,6 +1442,40 @@ impl RawTableInner {
...
+    //   g is the group width
+    // But this is ignoring the padding needed for ctrl_align.
+    // If we remember these restrictions:
+    //   x is always a power of 2
+    //   Layout size for T must always be a multiple of T
+    // Then the alignment can be ignored if we add the constraint:
+    //   x * y >= table_layout.ctrl_align
+    // This is taken care of by `capacity_to_buckets`.
+    let numerator = allocation_size - group_width;
+    let denominator = table_layout.size + 1; // todo: ZSTs?
```
Amanieu, line 1474: The ZST case seems to be handled properly here: the allocation will only contain control bytes and a 0-sized table. This comment can be removed.

src/raw/mod.rs
```diff
@@ -1442,6 +1442,40 @@ impl RawTableInner {
...
+    //   x is the number of buckets
+    //   y is the table_layout.size
+    //   z is the size of the allocation
+    //   g is the group width
+    // But this is ignoring the padding needed for ctrl_align.
+    // If we remember these restrictions:
+    //   x is always a power of 2
+    //   Layout size for T must always be a multiple of T
+    // Then the alignment can be ignored if we add the constraint:
+    //   x * y >= table_layout.ctrl_align
```
Amanieu, line 1471: I had to spend some time to work out why this was correct. The key insight here is that `ctrl_offset = align(x * y, ctrl_align)` in the layout calculation. It might be worth adding a line to the comment with this.

src/raw/mod.rs
```diff
@@ -1442,6 +1442,40 @@ impl RawTableInner {
...
+    // This is taken care of by `capacity_to_buckets`.
```
Amanieu, line 1472: This is brittle: currently `capacity_to_buckets` doesn't actually guarantee this. If we are going to rely on this guarantee then `capacity_to_buckets` needs a `debug_assert` and a comment indicating that `maximum_buckets_in` relies on this behavior. This will remind anyone who changes `capacity_to_buckets` in the future to also adjust `maximum_buckets_in`.

src/raw/mod.rs
morrisonlevi: I left a comment, take a look and suggest more if you can think of something more to note.

src/raw/mod.rs
morrisonlevi: Since #615, it _does_ guarantee this. Well, assuming I didn't make a mistake anyway ^_^. I tried finding issues with it but I don't see any, let me know if you see something.

I think adding assertions can be helpful to make this more obvious and be more resilient to future changes, so I've added `ensure_bucket_bytes_at_least_ctrl_align`. Take a look.

morrisonlevi:
@Amanieu Thanks for your reviews. I've responded, please let me know if you have more questions or suggestions.

PR #524, opened 2024-05-07
perf: increase min buckets on very small types

morrisonlevi:
Consider `HashSet<u8>` on x86_64 with SSE with various bucket sizes and how many bytes the allocation ends up being:

| buckets | capacity | allocated bytes |
| ------- | -------- | --------------- |
|       4 |        3 |              36 |
|       8 |        7 |              40 |
|      16 |       14 |              48 |
|       28 |      32 |              80 |

In general, doubling the number of buckets should roughly double the number of bytes used. However, for small bucket sizes for these small `TableLayout`s (4 -> 8, 8 -> 16), it doesn't happen. This is an edge case which happens because of padding of the control bytes and adding the `Group::WIDTH`. Taking the buckets from 4 to 16 (4x) only takes the allocated bytes from 36 to 48 (~1.3x).

This platform isn't the only one with edges. Here's aarch64 on an M1 for the same `HashSet<u8>`:

| buckets | capacity | allocated bytes |
| ------- | -------- | --------------- |
|       4 |        3 |              20 |
|       8 |        7 |              24 |
|      16 |       14 |              40 |

Notice doubling from 4 to 8 buckets only lead to 4 more bytes (20 -> 24) instead of roughly doubling.

Generalized, `buckets * table_layout.size` needs to be at least as big as `table_layout.ctrl_align`. For the cases I listed above, we'd get these new minimum bucket sizes:

 - x86_64 with SSE: 16
 - aarch64: 8

This is a niche optimization. However, it also removes possible undefined behavior edge case in resize operations. In addition, it would be a useful property when utilizing over-sized allocations (see rust-lang/hashbrown#523).

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -4131,7 +4131,8 @@ impl<'a, 'b, K, Q: ?Sized, V, S, A: Allocator> EntryRef<'a, 'b, K, Q, V, S, A> {
     #[cfg_attr(feature = "inline-more", inline)]
     pub fn insert(self, value: V) -> OccupiedEntry<'a, K, V, S, A>
     where
-        K: Hash + From<&'b Q>,
+        K: Hash,
+        &'b Q: Into<K>,
         S: BuildHasher,
     {
         match self {
@@ -4164,7 +4165,8 @@ impl<'a, 'b, K, Q: ?Sized, V, S, A: Allocator> EntryRef<'a, 'b, K, Q, V, S, A> {
     #[cfg_attr(feature = "inline-more", inline)]
     pub fn or_insert(self, default: V) -> &'a mut V
     where
-        K: Hash + From<&'b Q>,
+        K: Hash,
+        &'b Q: Into<K>,
         S: BuildHasher,
     {
         match self {
@@ -4194,7 +4196,8 @@ impl<'a, 'b, K, Q: ?Sized, V, S, A: Allocator> EntryRef<'a, 'b, K, Q, V, S, A> {
     #[cfg_attr(feature = "inline-more", inline)]
     pub fn or_insert_with<F: FnOnce() -> V>(self, default: F) -> &'a mut V
     where
-        K: Hash + From<&'b Q>,
+        K: Hash,
+        &'b Q: Into<K>,
         S: BuildHasher,
     {
         match self {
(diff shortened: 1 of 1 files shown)


src/raw/mod.rs
```diff
@@ -192,14 +192,40 @@ impl ProbeSeq {
...
+    let cap = {
+        let q = table_layout.ctrl_align / table_layout.size.max(1);
+        match q.checked_mul(7) {
+            None => cap,
+            Some(x) => {
+                let adjusted = x / 8;
+                adjusted.max(cap)
+            }
+        }
+    };
```
Amanieu, line 222: Rather than all this complicated logic, this could all be moved as hard-coded logic into the `cap < 8` case below. Specifically, when `cap <= 7` then we need to choose a bucket count of 4, 8 or 16 depending on the values of `size_of::<T>()` and `cap`.

Amanieu:
Also see #47 for previous discussion of this problem.

src/raw/mod.rs
morrisonlevi: `cap < 8` isn't sufficient for `size_of::<T>() == 1` on platforms with a group size of 16 as `buckets * table_layout.size >= table.ctrl_align` isn't true e.g. `8 * 1 >= 16`. That's why there's more complication.

With that said, with fresh eyes I don't like how the PR is currently written 😄. I'll take a look at improving it next week.

morrisonlevi: For now I at least adjusted the code comments and names to be a bit more explanatory.

morrisonlevi: I found something that seems better to me, at least.

src/raw/mod.rs
```diff
@@ -192,14 +192,39 @@ impl ProbeSeq {
...
+    // minimum capacity upwards for small items. This code only needs to handle
+    // ctrl_align which are less than or equal to Group::WIDTH, because valid
+    // layout sizes are always a multiple of the alignment, so anything with
+    // alignment over the Group::WIDTH won't hit this edge case.
+    let cap = cap.max(match (Group::WIDTH, table_layout.size) {
+        (16, 0..=1) => 14,
+        (16, 2..=3) => 7,
+        (8, 0..=1) => 7,
+        _ => 3,
+    });
```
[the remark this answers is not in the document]
Amanieu: What I had in mind was to move the `max` check inside the `if cap < 8` block entirely (and changing the bound to 15). Then just have hard-coded selection depending on the requested capacity and the group width/item size.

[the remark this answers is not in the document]
morrisonlevi: It feels weird to me. It'll be brittle if we ever add 256 width types, for instance. But I can try doing what you suggest.

PR #525, opened 2024-05-15
feat: borsh serde

dzmitry-lahoda:


dzmitry-lahoda:
not clear how next relates to my PR

```
Compiling rkyv_derive v0.7.44
error: bound is defined in more than one place
    --> src/map.rs:1267:30
     |
[126](https://github.com/rust-lang/hashbrown/actions/runs/9102104127/job/25020927498?pr=525#step:3:127)7 |     pub fn entry_ref<'a, 'b, Q: ?Sized>(&'a mut self, key: &'b Q) -> EntryRef<'a, 'b, K, Q, V, S, A>
     |                              ^
1268 |     where
1269 |         Q: Hash + Equivalent<K>,
     |         ^
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#multiple_bound_locations
     = note: `-D clippy::multiple-bound-locations` implied by `-D clippy::all`
     = help: to override `-D clippy::all` add `#[allow(clippy::multiple_bound_locations)]`

error: bound is defined in more than one place
    --> src/map.rs:[130](https://github.com/rust-lang/hashbrown/actions/runs/9102104127/job/25020927498?pr=525#step:3:131)8:16
     |
1308 |     pub fn get<Q: ?Sized>(&self, k: &Q) -> Option<&V>
     |                ^
1309 |     where
[131](https://github.com/rust-lang/hashbrown/actions/runs/9102104127/job/25020927498?pr=525#step:3:132)0 |         Q: Hash + Equivalent<K>,
```

Amanieu:
The CI failure should be fixed now, can you rebase your branch?

kayabaNerve:
As a policy question, why was borsh added as a dependency to hashbrown instead of borsh adding a dependency of hashbrown (and implementing on that end)?

kayabaNerve:
... borsh already implements support for hashbrown under the hashbrown feature flag??? This prevents borsh from supporting any hashbrown >= 0.15 due to the circular dependency it'd cause. borsh did future-proof by only supporting the range they know they can support (`< 0.15` is one of the bounds), so I don't immediately believe this will cause issues for anyone who updated to hashbrown 0.15 (as borsh should cause the dependency to be duplicated and maintain the <0.15 instance), yet I truly don't understand why this PR exists in the first place.

#554 demonstrates moving such functionality *out of this lib* and yet this PR *explicitly moves it into this lib*.

Amanieu:
There's no particular policy for where trait impls should go, but generally it should be in the crate that most frequently releases major versions with breaking changes.

There are some exceptions:
- If one crate depends on the other, make sure we don't end up with circular dependencies (this happened with rkyv).
- rayon support needs deep integration into the implementation details of `RawTable`. This cannot be done in `rayon` itself.

kayabaNerve:
ACK re: policy and rkyv. I'd argue that by necessity is primary, yet I'd argue for going by niche rather than by release cycle. Smaller crates should expand their functionality rather than expecting larger crates to do so to prevent overwhelming the larger crates (preventing a theoretical future where there's 100 different codecs in hashbrown instead of 1 hashbrown feature in 100 different codecs).

PR #528, opened 2024-05-30
Fix clippy multiple_bound_location warnings

ToMe25:
This PR moves the `?Sized` generics bound used in the map and set source files from the generics predicate to the where clause, if a where clause already exists.

According to what I could find online that `?Sized` bound had to be there until rustc 1.15, but since `?Sized` is already specified in where clauses in these files I don't think thats an issue.

I always put the `?Sized` bound last, because that is where it already appears in the docs, but if `?Sized` first is preferred, I'll change that.

This PR also changes the `Debug` impls for *EntryRef to use a where clause, because they currently have a linebreak in "impl X for Y" line.  
If this last change isn't wanted or considered something separate, please let me know :)

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -224,7 +224,7 @@ where
 #[cfg_attr(feature = "inline-more", inline)]
 fn equivalent_key<Q, K, V>(k: &Q) -> impl Fn(&(K, V)) -> bool + '_
 where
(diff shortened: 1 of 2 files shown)


Amanieu:
Thanks!

@bors r+

PR #531, opened 2024-06-07
Optimize Set is_disjoint

ToMe25:
By using the `Intersection` iterator in `HashSet::is_Disjoint` its performance can be significantly improved in some cases.  
This is because `intersection()` always uses the shorter set as its iterator.

It would also be possible to replicate this "Iterate over the smaller set and check in the larger set" logic in the is_disjoint method.  
However in my benchmarks the approach I chose is faster than that.

This change only causes a significant improvement when called on the larger of two disjoint sets.

My benchmark results:

Name | Before | After | Diff (%)
-- | -- | -- | --
disjoint_is_disjoint_large_small | 1,147.43 | 535.25 | -53,35 %
disjoint_is_disjoint_small_large | 535.66 | 527.59 | -1,51 %
subset_is_disjoint | 9.90 | 10.44 | 5,45 %
superset_is_disjoint | 9.80 | 10.43 | 6,43 %

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -4439,11 +4439,13 @@ impl<T, A: Allocator> FusedIterator for RawDrain<'_, T, A> {}
 ///   created will be yielded by that iterator.
 /// - The order in which the iterator yields buckets is unspecified and may
 ///   change in the future.
+#[cfg(feature = "raw")]
 pub struct RawIterHash<T> {
     inner: RawIterHashInner,
     _marker: PhantomData<T>,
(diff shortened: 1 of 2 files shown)


ToMe25:
That error seems to be a Dead Code error which I'm pretty sure this PR couldn't have caused, is there anything I can/should do about this?

Amanieu:
Just add `#[cfg(feature = "raw")]` to these types. The CI failure is probably due to lint changes in nightly rustc.

PR #535, opened 2024-06-22
Rework the Entry API

Amanieu:
The main change in this PR is that `OccupiedEntry` no longer holds the
key used in the initial `entry` call. As a result, `OccupiedEntryRef` is
no longer required since `OccupiedEntry` can be used in `EntryRef`
directly.

The following methods have been removed:
```rust
// hash_map

impl OccupiedEntry {
    fn replace_entry(self, value: V) -> (K, V);
    fn replace_key(self) -> K;
}

impl EntryRef {
    fn and_replace_entry_with<F>(self, f: F) -> Self;
}

impl VacantEntryRef {
    fn into_key(self) -> K;
}

// hash_set

impl Entry {
    fn replace(self) -> T;
}
```

The following methods have been added:
```rust
impl VacantEntry {
    fn insert_entry(self, value: V) -> OccupiedEntry<'a, K, V, S, A>;
}
```

The following methods have their signatures changed:
```rust
impl EntryRef {
    // Previously returned OccupiedEntryRef
    fn insert(self, value: V) -> OccupiedEntry<'a, K, V, S, A>;
}

impl VacantEntryRef {
    // Previously returned &Q
    fn key(&self) -> &'b Q;

    // Previously returned OccupiedEntryRef
    fn insert(self, value: V) -> OccupiedEntry<'a, K, V, S, A>;
}
```

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1229,7 +1229,6 @@ where
         if let Some(elem) = self.table.find(hash, equivalent_key(&key)) {
             Entry::Occupied(OccupiedEntry {
                 hash,
-                key: Some(key),
                 elem,
                 table: self,
             })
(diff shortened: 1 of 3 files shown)


Amanieu:
Have you tried using the `entry_ref` API?

jwiesler:
Ah, interesting, didn't know about this one thanks. Is there anything I can read up on why it was done this way? The two APIs seem very similar to me.

Amanieu:
It's a separate API for cases where you want to move the key into the map or make a copy of an existing key without moving it.

clarfonthey:
Currently working on updating `hashbrown` for the standard library, and does the removal of the `replace_entry` API mean that there was the decision to remove it on the Rust side as well? Or was it a mistake to remove it from the `RustcOccupiedEntry`?

PR #538, opened 2024-07-14
Shorten code

overlookmotel:
Condense code a little.

I don't know if little PRs like this have more downside in the noise they create for maintainers than the benefit of the changes themselves. If so, please say so! That'd be a completely legitimate response. I'm exploring hashbrown's code, and just submitting small bits where I find them.


Amanieu:
Thanks!

@bors r+

PR #539, opened 2024-07-14
Fix comment

overlookmotel:
From looking at code, I believe this comment was incorrect (or outdated). It seems that `T`s are stored in reverse order.

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -797,7 +797,7 @@ struct RawTableInner {
     // number of buckets in the table.
     bucket_mask: usize,
 
-    // [Padding], T_n, ..., T1, T0, CT0, CT1, ...
+    // [Padding], T_n, ..., T1, T0, C0, C1, ...
     //                              ^ points here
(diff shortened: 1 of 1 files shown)


src/raw/mod.rs
```diff
@@ -797,8 +797,8 @@ struct RawTableInner {
     // number of buckets in the table.
     bucket_mask: usize,
 
-    // [Padding], T1, T2, ..., Tlast, C1, C2, ...
-    //                                ^ points here
+    // [Padding], T_n, ..., T1, T0, CT0, CT1, ...
```
Amanieu, line 800:
```suggestion
    // [Padding], T_n, ..., T1, T0, C0, C1, ...
```

overlookmotel: Have changed as you suggest. I used to `CT` to be consistent with comments further down. Probably I misunderstood.

https://github.com/rust-lang/hashbrown/blob/894b56ffa517efeb14990b2104271c594bc9b9ca/src/raw/mod.rs#L921-L923

PR #541, opened 2024-07-27
Implement Clone and Debug for HashTable's Iter struct

geeknoid:
Found those implementations missing, which is preventing me from implementing the same in encapsulating iterators.

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -1969,6 +1969,23 @@ impl<T> ExactSizeIterator for Iter<'_, T> {
 
 impl<T> FusedIterator for Iter<'_, T> {}
 
+// FIXME(#26925) Remove in favor of `#[derive(Clone)]`
+impl<'a, T> Clone for Iter<'a, T> {
+    #[cfg_attr(feature = "inline-more", inline)]
+    fn clone(&self) -> Iter<'a, T> {
+        Iter {
+            inner: self.inner.clone(),
+            marker: PhantomData,
+        }
+    }
+}
+
+impl<T: fmt::Debug> fmt::Debug for Iter<'_, T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_list().entries(self.clone()).finish()
+    }
+}
+
 /// A mutable iterator over the entries of a `HashTable` in arbitrary order.
 /// The iterator element type is `&'a mut T`.
 ///


src/table.rs
```diff
@@ -1914,6 +1931,23 @@ impl<T> ExactSizeIterator for IterMut<'_, T> {
 
 impl<T> FusedIterator for IterMut<'_, T> {}
 
+// FIXME(#26925) Remove in favor of `#[derive(Clone)]`
+impl<'a, T> Clone for IterMut<'a, T> {
```
Amanieu, line 1935: Cloning a mutable iterator is unsound since it could be used to create multiple aliasing mutable references. All the other impls seem fine though.

geeknoid: Sorry, I'm still new to Rust. I understand your point and it makes sense, so I removed the implementation of Clone here. And in fact, looking at my own code, my wrapping iterators haven't been implementing this either.

However, if this leads to unsoundness, how come I was able to implement the logic in safe code and the compiler didn't complain? Is something in HashTable incorrectly breaking the rules somehow which allowed my transgression?

Amanieu: This isn't entirely safe: if you look above, the `Iterator` trait implementation for `IterMut` uses unsafe code. That unsafe code is only valid if it is the only instance of `IterMut` for a table.

.idea/.gitignore
Amanieu, line 1: Don't include these auto-generated files.

src/table.rs
Amanieu: This isn't resolved, it seem you have re-added the incorrect `Clone` impl.

geeknoid: OK, redeleted. Sorry about that.

cuviper: On the other hand, `Debug for IterMut` is fine.

geeknoid: To implement Debug, I needed Clone, so I just punted on IterMut alltogether.

cuviper: The map `IterMut` does it through a private `iter()` that only clones the `inner: RawIter`.

https://github.com/rust-lang/hashbrown/blob/a69af93fe1aca2b2eedf1e3e01ef7eaef10be927/src/map.rs#L3175-L3183

https://github.com/rust-lang/hashbrown/blob/a69af93fe1aca2b2eedf1e3e01ef7eaef10be927/src/map.rs#L2137-L2146

PR #542, opened 2024-07-30
Implement Default for iterators

clarfonthey:
See rust-lang/rust#128261 for a similar PR for some libstd iterators. In order to do a similar PR for the `HashSet` and `HashMap` types, we need to make these changes to `hashbrown` first to actually implement them.

One small caveat is that I chose to only implement `Default` for `RawIter` and not `RawIterRange`, since it really exists as an implementation detail anyway, and the best implementation would involve just using the `RawIter` implementation anyway.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -2447,6 +2447,14 @@ pub struct IntoKeys<K, V, A: Allocator = Global> {
     inner: IntoIter<K, V, A>,
 }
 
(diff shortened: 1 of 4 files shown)


JustForFun88:
As far as I remember, parallel iterator (rayon) uses RawIterRange directly, so maybe it would be nice to implement Default for RawIterRange too.

clarfonthey:
That would be a bit difficult for a `pub(crate)` iterator.

PR #543, opened 2024-08-05
Implement `Clone`, `Debug` for iterators

clarfonthey:
Follow-up to #542 after I noticed these were missing too. A more complete implementation of #541.

Adds:

* explicit `Clone` for `raw::Iter` and `raw::IntoIter`
* derived `Clone` for iterators that use `raw::IntoIter` internally
* explicit `Debug` implementations for `raw::Iter`, `raw::IterMut`, and `raw::IntoIter`
* public `iter` and `iter_mut` methods to `raw::IntoIter` (only `iter` is needed for `Debug`, but I decided to add these as public since map and set `IntoIter` seem to have public `iter` methods)

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -4280,6 +4280,7 @@ impl ExactSizeIterator for FullBucketsIndices {}
 impl FusedIterator for FullBucketsIndices {}
 
 /// Iterator which consumes a table and returns elements.
+#[derive(Clone)]
 pub struct RawIntoIter<T, A: Allocator = Global> {
     iter: RawIter<T>,
     allocation: Option<(NonNull<u8>, Layout, A)>,
--- a/src/map.rs
+++ b/src/map.rs
@@ -2399,6 +2399,7 @@ impl<K, V> IterMut<'_, K, V> {
 /// assert_eq!(iter.next(), None);
 /// assert_eq!(iter.next(), None);
 /// ```
+#[derive(Clone)]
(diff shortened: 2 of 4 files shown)


src/raw/mod.rs
```diff
@@ -4280,6 +4280,7 @@ impl ExactSizeIterator for FullBucketsIndices {}
 impl FusedIterator for FullBucketsIndices {}
 
 /// Iterator which consumes a table and returns elements.
+#[derive(Clone)]
 pub struct RawIntoIter<T, A: Allocator = Global> {
     iter: RawIter<T>,
     allocation: Option<(NonNull<u8>, Layout, A)>,
```
Amanieu, line 4286: This is not going to work correctly since it will end up double-freeing the underlying allocation.

clarfonthey: Right, that makes sense.

I'll look into actually cloning the allocation instead.

clarfonthey:
Going to close this per: https://github.com/rust-lang/rust/pull/128711#issuecomment-2295040473

I have some ideas on a refactor that would make implementing `Clone` a bit more comfortable, and I'll draft up my general plans in an issue first before implementing so folks can comment on them.

I posted some [less constructive feedback](https://toot.cat/@clarfonthey/112978842060026337) earlier today, but the long story short is that I think that there need to be more safe wrappers on the internals in the code, rather than every single usage of the raw table being unsafe. I need to do more reading into the code before I make a proper writeup though, so, stay tuned.

Issue #545, opened 2024-08-20
Refactoring proposal: cleaning up the internal APIs


clarfonthey:
This is my writeup re: https://github.com/rust-lang/hashbrown/pull/543#issuecomment-2295041421

cc #290 as this should also strive to fix that as well. Ideally, by the end of the refactor, we should be able to add the following to the crate root:

```rust
#![deny(
    clippy::missing_safety_doc,
    clippy::missing_errors_doc,
    clippy::undocumented_unsafe_blocks
)]
```

## Documenting the actual layout

Right now, the documentation for the crate on the actual layout of the hash table is *incredibly* lacking. It mostly just points to the C++ SwissTable implementation and says "we did that!" while ignoring the fact that, *obviously*, it is not quite that. There are a lot of subtleties to the Rust code that are completely omitted from the basic description; both the downsides *and* the upsides are completely undocumented, IMHO, since even if they do exist they're difficult to find in the code.

We should start out by describing the layout from its broadest points first and use this to design the API. I'm also going to be *renaming* some terms so we can use the more specific names for individual pieces inside the API.

So, here's how I've decided to describe it:

The main benefit of this hash table implementation is that unlike other implementations, instead of a unified *bucket* array, we have two separate *data* and *control* arrays. Having a small, dedicated *control* array which does not depend on the type inside the table makes it extremely easy to traverse the table using SIMD instructions. It also improves cache locality by keeping the control data as small as possible, so that any operation on the *table* can happen entirely in cache, even if the data itself is a cache miss.

While the control and data arrays are separate, we can conceptually treat them together as an array of *buckets*, which forms our hash table.

### Control Array

The control array contains a series of *tags* which indicate the status of any given bucket. To make SIMD operations efficient, we split the tags into *groups* whose size depends on the specific SIMD implementation. This imposes a restriction on the control array that its size is at least a multiple of the group size, and its alignment is at least a multiple of the group alignment. We additionally require that the number of buckets, and hence the number of tags, be a power of two.

There are two special tags, `EMPTY` and `DELETED`, which correspond to buckets which have either always been empty, or who have recently become empty. All other tags are considered `FULL`.

Regardless of implementation, the tags are:

* `EMPTY`: `0b1_1111111`
* `DELETED`: `0b1_0000000`
* `FULL`: `0b0_xxxxxxx`, where `xxxxxxx` are seven bits from the hash

We can use the term `NONFULL` to refer to buckets which are either `EMPTY` or `DELETED`.

[...]

PR #546, opened 2024-08-22
Remove the `raw` feature and make `RawTable` private

Amanieu:
This will give more freedom for the internal implementation details of hashbrown to evolve without the need for regular releases with breaking changes.

All existing users of `RawTable` should migrate to the `HashTable` API which is entirely safe while providing the same flexibility as `RawTable`.

This also removes the following features which were only exposed under `RawTable`:
- `RawTable::iter_hash`
- `RawIter::reflect_insert` and `RawIter::reflect_remove`
- `RawTable::clone_from_with_hasher`
- `RawTable::insert_no_grow` and `RawTable::try_insert_no_grow`
- `RawTable::allocation_info`
- `RawTable::try_with_capacity(_in)`
- `HashMap::raw_table(_mut)` and `HashSet::raw_table(_mut)`

If anyone was previously relying on this functionaly, please raise a comment. It may be possible to re-introduce it as a safe API in `HashTable` and/or `HashMap`.


Amanieu:
cc @clarfonthey

Dandandan:
I think https://github.com/Apache/arrow-datafusion 

Uses:
* RawTable::try_insert_no_grow
* RawTable::allocation_info

clarfonthey:
Since you're already doing this, would you mind pushing a release after too so rust-lang/rust#128711 can go? Please and thank you. <3

IMHO the sooner we do this, the better, since it gives people time to migrate away from the API before we start following up on the threat to improve the code. And, since it's a published crate, it's not a big deal, since the old (non-breaking) versions will continue to work correctly.

Dandandan:
FYI @alamb @avantgardnerio

alamb:
We have also been discussing implementing our own custom hash table in https://github.com/apache/datafusion/issues/7095, so perhaps this would be another potential reason to pursue that idea

I agree that using the low level details of how the hash table in hashbrown is implemented is not ideal (e.g. it constrains how hashbrown  can version releases)

@Amanieu  I wonder if you would consider a feature flag on hashbrown like `hashbrown_unstable` (following [the model of `tokio_unstable`](https://docs.rs/tokio/latest/tokio/#unstable-features)). That would DataFusion keep using hashbrown internals (and it would be on us to deal with any changes during upgrade)

avantgardnerio:
This will require work in DataFusion to fix TopK aggregations, and then after it is fixed it will cause significant (40% IIRC) performance regressions.

Amanieu:
I don't think that's a good idea because every minor release of hashbrown could break users of your crate.

However you can just keep using the 0.14 version of hashbrown which will still have the `raw` feature. It's only being removed in 0.15.

alamb:
> However you can just keep using the 0.14 version of hashbrown which will still have the raw feature. It's only being removed in 0.15.

Another option is for us to to fork the crate which might be reasonable if we can get more performance by doing so

Dandandan:
I think we should create some tickets in DataFusion for moving our usage towards `HashTable` / own hashtable (e.g. better vectorized) / forked code.

the-mikedavis:
I was using `RawTable::iter_hash` for a [multi-map](https://en.cppreference.com/w/cpp/container/unordered_multimap)-like type (kind of like `HashMap<K, Vec<T>>`) and I don't believe there's a way to look up multiple entries with the same key from `HashTable<(K, V)>`: `HashTable::find` stops after the first matching element. Would it be possible to introduce something safe in `HashTable` to take its place? I can share more code if you'd like more context and maybe work on a PR if you think it's a reasonable addition.

Amanieu:
I think `iter_hash` makes sense to expose on `HashTable`. Feel free to send a PR, you can do a partial revert of this one to get `iter_hash` back in `RawTable`.

xacrimon:
@Amanieu [dashmap](https://crates.io/crates/dashmap) relies on RawTable functionality. With this merged we'd be effectively locked to 0.14 forever.

Amanieu:
What specific functionality do you need that isn't available though `HashTable` which is the replacement for `RawTable`?

cuviper:
I took a brief look, and the main roadblock that I see will be iterators and entry structs that currently contain `RwLock` guards with corresponding `RawIter` / `Bucket` / `InsertSlot`. With `HashTable`, those second parts all have lifetimes that would need to borrow through the guard, so they become self-referential.

clarfonthey:
Added a comment to #545 mentioning to look into dashmap's internals when I go about ripping out the raw table API. It may be the case that we need to offer some kind of lifetime-erased version of the various types to get it to work, but I'm hoping we can get around that.

Amanieu:
> I took a brief look, and the main roadblock that I see will be iterators and entry structs that currently contain `RwLock` guards with corresponding `RawIter` / `Bucket` / `InsertSlot`. With `HashTable`, those second parts all have lifetimes that would need to borrow through the guard, so they become self-referential.

I thought about this a bit and fundamentally it's not a problem with `HashTable` but rather a limitation of the `MutexGuard` API. I think you can work around this by using a `Mutex<()>` and an `UnsafeCell<HashTable>` where you manage the locking manually.

kyren:
Hi, two projects I'm affiliated with are stuck on hashbrown 0.14 right now due to using the raw API in a way that can't be replicated in 0.15. They both use `RawTable` for *nearly* the same purpose, which I will try to explain.

https://github.com/ruffle-rs/ruffle uses the `RawTable` API to implement AVM2's iteration protocol for object properties. I'm really unfamiliar with this, @moulins other people on the Ruffle team can probably explain in better detail, but the bytecode assumes that object properties are stored in a bucket based hash table, and the bytecode can get entries from an object based on a bucket index and also has an instruction like `HasNext(object, index) -> next_index` to find the next valid bucket.

https://github.com/kyren/piccolo (which is something I'm much better equipped to explain) uses the `RawTable` API for something similar. Lua has a function called `next` that can be used to find "next" key in a Lua table, which is normally done by its "raw" order in a hash table, in hashbrown terms the bucket index (PUC-Rio Lua does not implement tables this way, or at least not all of them do, so this could equivalently be some order from a red-black tree).

The core issue is that both projects implement VMs for languages whose implementation assumes that a hash table has some kind of *observable order* that is stable ONLY when the hash table is not modified in certain ways. For piccolo (and this is the same in the internals of PUC-Rio Lua), that requirement is actually that the hash table order only be stable under *no inserts or removals whatsoever*, only mutating values and *transparent* modifications of keys. How this is accomplished is with something the PUC-Rio Lua and piccolo internals refer to as "dead keys", having a version of a key that compares identically to live keys but doesn't own a value for the purposes of garbage collection. When removing an entry in a hash table, you set the key to its "dead key" equivalent and set the value to `nil` (an entry with a `nil` value in Lua is indistinguishable to an entry not being present). Ruffle's AVM2 does not use this trick right now and instead relies on the order being guaranteed not to change without a rehash, but it is possible to efficiently implement in the AVM2 VM in a similar way to piccolo with tombstone entries, loosening the requirements.

We (some of the Ruffle devs and I) believe we have a minimal API that satisfies both our use cases and comes with hopefully minimal guarantees.

If it were only `piccolo`, the required API would be to get the *next* live entry after some given entry (and some way to get the *first* entry), in whatever the bucket order is, with no guarantees about that order being the same after any sort of mutation of the `HashTable` itself at all. AVM2 is slightly different due to needing to get an entry specifically by bucket index (which the AVM2 VM also caches internally in certain cases), but it never needs to get the bucket index by looking up an existing key (which piccolo must have to implement `next`). Therefore, a union of both requirements could look like this:

1) Access an entry by bucket index
2) Ask for the first valid bucket index greater than or equal to a given one (`0` for the first valid bucket)
3) Ask for the bucket index corresponding to an occupied entry
4) No guarantees whatsoever for iteration order after kind of "table mutation", both insertion and removal

There are probably a lot of room for equivalent APIs that would work here also. The core issue is that we need to observe the bucket order for a hash table that is *not being modified*. It is possible ofc that both projects could create a less efficient custom data structure with auxiliary arrays for the entry order, but in both cases these are pretty much *the* core data structures for the implemented language, so this is valuable enough that our best alternative other than some kind of API change here is a custom fork hashbrown (or stay on 0.14 indefinitely).

clarfonthey:
So, I just want to clarify here that the reason why `RawTable` is being deprecated is because it's old, not written with modern Rust features (like `MaybeUninit`) in mind, and is already soft-replaced by the `HashTable` API. So, all of this functionality you want to use seems extremely reasonable to add onto the `HashTable` API, since it all feels reasonable to me: the goal isn't to get rid of the lower-level APIs, but rather to offer them in a better place.

I think that the ability to keep track of bucket indices beyond just hashes seems reasonable, although I think that at least linking a few demonstrations of existing code you cannot rewrite under the `HashTable` API would be useful, so we know specifically what methods you're using and how they could best be replaced. I get the vague idea based upon what you're describing, but code examples would be the most helpful, since that way we can think of a suitable replacement API.

I think it would be most helpful to open a separate issue for these features (or multiple, depending on how different they are) so they can be tracked and followed better. There are a bunch of different features that people want from `RawTable` that haven't made their way onto `HashTable` yet, and it's best to separate out the discussions so we make sure nothing gets missed.

kyren:
> So, I just want to clarify here that the reason why RawTable is being deprecated is because it's old, not written with modern Rust features (like MaybeUninit) in mind, and is already soft-replaced by the HashTable API. So, all of this functionality you want to use seems extremely reasonable to add onto the HashTable API, since it all feels reasonable to me: the goal isn't to get rid of the lower-level APIs, but rather to offer them in a better place.

Ah I see, I may have gotten the wrong impression from reading the PR comments and from discussion with Ruffle devs, I don't think I correctly understood the primary motivation for getting rid of RawTable (implementation freedom vs just being old and needing to be rewritten). I understand better now, thank you!

> I think that the ability to keep track of bucket indices beyond just hashes seems reasonable, although I think that at least linking a few demonstrations of existing code you cannot rewrite under the HashTable API would be useful, so we know specifically what methods you're using and how they could best be replaced. I get the vague idea based upon what you're describing, but code examples would be the most helpful, since that way we can think of a suitable replacement API.

Sure thing, the code I'm the most familiar with is is https://github.com/kyren/piccolo/blob/master/src/table/raw.rs, specifically the `RawTable::next` method. The way that Lua tables work and Lua table iteration works is a bit strange especially if you're not familiar with Lua, but hopefully it should be possible to follow.

Berrysoft:
https://github.com/Berrysoft/wepoll2 I used the `raw` feature to make a panic-free `HashMap`. The `insert` method of `HashMap` calls `reserve` which will panic if allocation fails. I have to reimplement a `try_insert` method to call `try_reserve` to remove the panic code path. Do you have some advice on porting to v0.15?

clarfonthey:
I think that falliable allocation APIs are a very reasonable addition to all the types in the crate, so, it's worth filing an issue for.

comphead:
DataFusion crosslink https://github.com/apache/datafusion/issues/13433

PR #547, opened 2024-08-23
Add `get_or_insert_equivalent` method to HashSet

Soveu:
I'm probably the only person that might need this API, however it also simplifies a little other `get_or_insert` methods, so maybe it will be worth going this direction.


Soveu:
Yeah, seems it does the same thing.

PR #549, opened 2024-08-28
Add `HashTable::iter_hash`, `HashTable::iter_hash_mut`

the-mikedavis:
This is a follow-up to https://github.com/rust-lang/hashbrown/pull/546 ([comment](https://github.com/rust-lang/hashbrown/pull/546#issuecomment-2316251486)). `iter_hash` from the old raw API can be useful for reading from a "bag" / "multi map" type which allows duplicate key-value pairs. Exposing it safely in `HashTable` takes a fairly small wrapper around `RawIterHash`. This PR partially reverts #546 to restore `RawTable::iter_hash` and its associated types.

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -1372,6 +1372,20 @@ impl<T, A: Allocator> RawTable<T, A> {
         self.table.iter()
     }
 
+    /// Returns an iterator over occupied buckets that could match a given hash.
+    ///
+    /// `RawTable` only stores 7 bits of the hash value, so this iterator may
+    /// return items that have a hash value different than the one provided. You
+    /// should always validate the returned values before using them.
+    ///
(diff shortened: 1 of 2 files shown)


Amanieu:
Should there be a `mut` version of this iterator? What about removing one item out of multiple with the same key? Do we also need to add APIs for those use cases?

the-mikedavis:
For my use-case I will only need to read entries as-is after insertion. I could see mutation being useful hypothetically though. Maybe that could benefit from some kind of iterator that returns `Entry`-like items?

Amanieu:
Returning `Entry` doesn't work because you then end up with multiples `Entry`s each of which has a mutable reference to the table, which violates aliasing rules. I think just adding `iter_hash_mut` for now would be enough.

the-mikedavis:
Thanks @Amanieu! My multi map type is much cleaner as a HashTable now :)

PR #550, opened 2024-08-29
Fix `--no-default-features` without `--features raw-entry`

cuviper:
`HashSet` still uses `raw_entry_mut()` internally, but now we can build
that without exposing it in the public API.


cuviper:
Subsumed by #555.

PR #556, opened 2024-09-12
Make `insert_unique_unchecked` unsafe

Amanieu:
This is in line with the standard library guarantees that it should be impossible to create an inconsistent `HashMap` with well-defined key types.

### DIFF
--- a/src/set.rs
+++ b/src/set.rs
@@ -1118,25 +1118,29 @@ where
 
     /// Insert a value the set without checking if the value already exists in the set.
     ///
-    /// Returns a reference to the value just inserted.
-    ///
-    /// This operation is safe if a value does not exist in the set.
-    ///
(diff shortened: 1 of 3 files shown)


src/set.rs
```diff
@@ -1118,25 +1118,29 @@ where
...
+    /// For example, when constructing a set from another set, we know
+    /// that values are unique.
+    ///
+    /// # Safety
     ///
-    /// This operation is safe if a value does not exist in the set.
+    /// This operation is safe if a key does not exist in the set.
     ///
-    /// However, if a value exists in the set already, the behavior is unspecified:
+    /// However, if a key exists in the set already, the behavior is unspecified:
```
cuviper, line 1132: These should probably be left as "value", not "key".

cuviper (commented):
Makes sense to me to be `unsafe`.

PR #558, opened 2024-09-16
Reduce typo count.

waywardmonkeys:
This renames 2 identifiers, one in benches, one in Rayon support. Neither are public.

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -135,7 +135,7 @@ fn h1(hash: u64) -> usize {
     hash as usize
 }
 
-// Constant for h2 function that grabing the top 7 bits of the hash.
+// Constant for h2 function that grabs the top 7 bits of the hash.
 const MIN_HASH_LEN: usize = if mem::size_of::<usize>() < mem::size_of::<u64>() {
     mem::size_of::<usize>()
 } else {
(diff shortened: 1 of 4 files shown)


src/raw/mod.rs
```diff
@@ -1788,7 +1788,7 @@ impl RawTableInner {
     ///
     /// This function does not check if the given element exists in the table. Also,
     /// this function does not check if there is enough space in the table to insert
-    /// a new element. Caller of the funtion must make ensure that the table has at
+    /// a new element. The caller of the function must make ensure that the table has at
```
cuviper, line 1791: This should either be "make sure" or "ensure".
```suggestion
    /// a new element. The caller of the function must make sure that the table has at
```

src/raw/mod.rs
```diff
@@ -3171,7 +3171,7 @@ impl<T: Clone, A: Allocator + Clone> Clone for RawTable<T, A> {
                 // Avoid `Result::ok_or_else` because it bloats LLVM IR.
                 //
                 // SAFETY: This is safe as we are taking the size of an already allocated table
-                // and therefore сapacity overflow cannot occur, `self.table.buckets()` is power
+                // and therefore capacity overflow cannot occur, `self.table.buckets()` is power
```
cuviper, line 3174: How strange, that first "c" was a U+0441 Cyrillic Es before!

cuviper (approved):
LGTM, just one additional nit if you don't mind.

src/raw/mod.rs
waywardmonkeys: Indeed!

PR #561, opened 2024-09-20
Implement `Debug`, `FusedIterator` and `Iterator::fold` for all `HashTable` iterators

Amanieu:


src/table.rs
```diff
@@ -2074,8 +2088,19 @@ impl<'a, T> Iterator for IterHash<'a, T> {
...
         }
     }
+
+    fn fold<B, F>(self, init: B, mut f: F) -> B
+    where
+        Self: Sized,
+        F: FnMut(B, Self::Item) -> B,
+    {
+        self.inner
+            .fold(init, |acc, bucket| unsafe { f(acc, bucket.as_ref()) })
```
cuviper, line 2098: This is *okay*, but the inner `RawIterHash` isn't doing anything special for `fold`.
(Ditto for `IterHashMut`, and `Drain`'s `RawDrain`.)

cuviper (commented):
Can `IterHash` and `IterHashMut` also get `Debug`?
And perhaps `Clone for IterHash` too?

src/raw/mod.rs
```diff
@@ -4099,6 +4101,27 @@ impl<T> RawIterHash<T> {
...
+    #[cfg_attr(feature = "inline-more", inline)]
+    fn clone(&self) -> Self {
+        Self {
+            inner: self.inner.clone(),
+            _marker: PhantomData,
+        }
+    }
+}
+
+impl<T> Default for RawIterHash<T> {
```
cuviper, line 4114: Great - did you mean to propagate that up too?

PR #562, opened 2024-09-25
Change signature of `get_many_mut` APIs

Urgau:
This PR changes the signature and contract of the `get_many_mut` APIs by

1. panicking on overlapping keys
2. returning an array of Option rather than an Option of array.

This was asked by T-libs-api in https://github.com/rust-lang/rust/issues/97601#issuecomment-2372109119 regarding the corresponding std `HashMap` functions.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1467,8 +1467,11 @@ where
     /// Attempts to get mutable references to `N` values in the map at once.
     ///
     /// Returns an array of length `N` with the results of each query. For soundness, at most one
-    /// mutable reference will be returned to any value. `None` will be returned if any of the
-    /// keys are duplicates or missing.
+    /// mutable reference will be returned to any value. `None` will be used if the key is missing.
+    ///
+    /// # Panics
+    ///
+    /// Panics if any keys are overlapping.
     ///
     /// # Examples
     ///
@@ -1481,16 +1484,23 @@ where
     /// libraries.insert("Herzogin-Anna-Amalia-Bibliothek".to_string(), 1691);
     /// libraries.insert("Library of Congress".to_string(), 1800);
     ///
+    /// // Get Athenæum and Bodleian Library
+    /// let [Some(a), Some(b)] = libraries.get_many_mut([
+    ///     "Athenæum",
+    ///     "Bodleian Library",
+    /// ]) else { panic!() };
+    ///
+    /// // Assert values of Athenæum and Library of Congress
     /// let got = libraries.get_many_mut([
     ///     "Athenæum",
     ///     "Library of Congress",
     /// ]);
     /// assert_eq!(
     ///     got,
-    ///     Some([
-    ///         &mut 1807,
-    ///         &mut 1800,
-    ///     ]),
+    ///     [
+    ///         Some(&mut 1807),
+    ///         Some(&mut 1800),
+    ///     ],
     /// );
     ///
     /// // Missing keys result in None
@@ -1498,16 +1508,28 @@ where
     ///     "Athenæum",
     ///     "New York Public Library",
     /// ]);
-    /// assert_eq!(got, None);
+    /// assert_eq!(
+    ///     got,
+    ///     [
+    ///         Some(&mut 1807),
+    ///         None
+    ///     ]
+    /// );
+    /// ```
     ///
-    /// // Duplicate keys result in None
+    /// ```should_panic
+    /// use hashbrown::HashMap;
+    ///
+    /// let mut libraries = HashMap::new();
+    /// libraries.insert("Athenæum".to_string(), 1807);
+    ///
+    /// // Duplicate keys panic!
     /// let got = libraries.get_many_mut([
     ///     "Athenæum",
     ///     "Athenæum",
     /// ]);
-    /// assert_eq!(got, None);
     /// ```
-    pub fn get_many_mut<Q, const N: usize>(&mut self, ks: [&Q; N]) -> Option<[&'_ mut V; N]>
+    pub fn get_many_mut<Q, const N: usize>(&mut self, ks: [&Q; N]) -> [Option<&'_ mut V>; N]
     where
         Q: Hash + Equivalent<K> + ?Sized,
     {
@@ -1517,8 +1539,8 @@ where
     /// Attempts to get mutable references to `N` values in the map at once, without validating that
     /// the values are unique.
     ///
-    /// Returns an array of length `N` with the results of each query. `None` will be returned if
-    /// any of the keys are missing.
+    /// Returns an array of length `N` with the results of each query. `None` will be used if
+    /// the key is missing.
     ///
     /// For a safe alternative see [`get_many_mut`](`HashMap::get_many_mut`).
     ///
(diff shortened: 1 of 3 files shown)


src/table.rs
```diff
@@ -969,6 +969,10 @@ where
     /// mutable reference will be returned to any value. `None` will be returned if any of the
     /// keys are duplicates or missing.
```
cuviper, line 970: Another doc update needed here.

src/raw/mod.rs
```diff
@@ -1291,48 +1291,52 @@ impl<T, A: Allocator> RawTable<T, A> {
...
-                    return None;
+                if let Some(cur) = cur {
+                    if ptrs[..i]
+                        .iter()
+                        .flatten()
+                        .any(|&prev| ptr::eq::<T>(prev, cur))
+                    {
+                        panic!("duplicate keys found");
+                    }
                 }
```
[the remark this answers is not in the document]
cuviper: Actually, I think you could further change to not destructure `&cur` in the pattern before, but I'm not going to bother writing a larger "suggested change" here.

src/raw/mod.rs
```diff
@@ -1291,48 +1291,52 @@ impl<T, A: Allocator> RawTable<T, A> {
...
+        let ptrs = self.get_many_mut_pointers(hashes, eq);
+        ptrs.map(|ptr| ptr.map(|ptr| &mut *ptr))
     }
 
     unsafe fn get_many_mut_pointers<const N: usize>(
         &mut self,
         hashes: [u64; N],
         mut eq: impl FnMut(usize, &T) -> bool,
-    ) -> Option<[*mut T; N]> {
+    ) -> [Option<*mut T>; N] {
```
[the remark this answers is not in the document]
Urgau: I've changed it to use a `NonNull<T>`, but I had a to aa convenience method `Bucket::as_non_null`, as `Bucket::as_ptr` does some non-trivial operations.

src/raw/mod.rs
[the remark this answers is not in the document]
Urgau: I applied your suggested change. Regarding your second point, I'm not sure I'm following, did you mean `(i, Some(cur))`?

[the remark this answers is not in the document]
cuviper: I meant changing `for (i, &cur) in` to `(i, cur)`, because `is_some` and `contains` both need that reference anyway. It shouldn't matter much though, LLVM can surely optimize it either way.

src/raw/mod.rs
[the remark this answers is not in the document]
cuviper: Oh, good call, I forgot that the `Bucket` wasn't a *direct* pointer.

src/raw/mod.rs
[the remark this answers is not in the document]
Urgau: Done anyway.

cuviper (approved):
LGTM, but I'll leave it to @Amanieu to make sure of the API change.

src/raw/mod.rs
```diff
@@ -1291,48 +1298,46 @@ impl<T, A: Allocator> RawTable<T, A> {
...
         // TODO use `MaybeUninit::uninit_array` here instead once that's stable.
-        let mut outs: MaybeUninit<[*mut T; N]> = MaybeUninit::uninit();
+        let mut outs: MaybeUninit<[Option<NonNull<T>>; N]> = MaybeUninit::uninit();
         let outs_ptr = outs.as_mut_ptr();
 
         for (i, &hash) in hashes.iter().enumerate() {
-            let cur = self.find(hash, |k| eq(i, k))?;
-            *(*outs_ptr).get_unchecked_mut(i) = cur.as_mut();
+            let cur = self.find(hash, |k| eq(i, k)).map(|cur| cur.as_non_null());
+            *(*outs_ptr).get_unchecked_mut(i) = cur;
```
Urgau, line 1337: This is pre-existing but this line creates a `&mut uninit` which the UB-ness of that is still undecided https://github.com/rust-lang/unsafe-code-guidelines/issues/346. Let me change it to pure raw pointers instead to be on the safe side.

Urgau: Done in [`b582436` (#562)](https://github.com/rust-lang/hashbrown/pull/562/commits/b582436f06c080198f388ce2e2eb4030acc64afa)

cuviper: Couldn't you still do that with the `NonNull`?

Urgau: Of course, fixed in [`6b288355`](https://github.com/rust-lang/hashbrown/pull/562/commits/6b2883557ff39d6f63c050b495dd277718efc77b).

*Sorry for that, I'm working with two different computers and forgot to pull on the first one. My bad.*

src/raw/mod.rs
```diff
@@ -1291,48 +1298,47 @@ impl<T, A: Allocator> RawTable<T, A> {
...
     unsafe fn get_many_mut_pointers<const N: usize>(
         &mut self,
         hashes: [u64; N],
         mut eq: impl FnMut(usize, &T) -> bool,
-    ) -> Option<[*mut T; N]> {
+    ) -> [Option<NonNull<T>>; N] {
         // TODO use `MaybeUninit::uninit_array` here instead once that's stable.
-        let mut outs: MaybeUninit<[*mut T; N]> = MaybeUninit::uninit();
-        let outs_ptr = outs.as_mut_ptr();
+        let mut outs: MaybeUninit<[Option<NonNull<T>>; N]> = MaybeUninit::uninit();
```
Amanieu, line 1332: This can be simplified to just be a `[Option<NonNull<T>>; N]` without the `MaybeUninit`. The array can just be initialized to `None`. This also addresses the TODOs.

src/map.rs
```diff
@@ -1467,8 +1467,11 @@ where
     /// Attempts to get mutable references to `N` values in the map at once.
     ///
     /// Returns an array of length `N` with the results of each query. For soundness, at most one
-    /// mutable reference will be returned to any value. `None` will be returned if any of the
-    /// keys are duplicates or missing.
+    /// mutable reference will be returned to any value. `None` will be used if the key is missing.
+    ///
+    /// # Panics
+    ///
+    /// Panics if some keys are overlapping.
```
Amanieu, line 1474:
```suggestion
    /// Panics if any keys are overlapping.
```

src/map.rs
```diff
@@ -1487,27 +1490,33 @@ where
     /// ]);
     /// assert_eq!(
     ///     got,
-    ///     Some([
-    ///         &mut 1807,
-    ///         &mut 1800,
-    ///     ]),
+    ///     [
```
Amanieu, line 1493: I think it would help if the docs showed an example using the let..else syntax to extract the results:
```rust
let [Some(a), Some(b)] = map.get_many_mut(..) else { panic!() };
```

src/raw/mod.rs
Urgau: Indeed, done.

src/map.rs
Urgau: Added let..else examples.

src/raw/mod.rs
```diff
@@ -1291,48 +1297,45 @@ impl<T, A: Allocator> RawTable<T, A> {
...
-        let mut outs: MaybeUninit<[*mut T; N]> = MaybeUninit::uninit();
+    ) -> [Option<NonNull<T>>; N] {
+        let mut outs: [Option<NonNull<T>>; N] = [None; N];
         let outs_ptr = outs.as_mut_ptr();
 
         for (i, &hash) in hashes.iter().enumerate() {
-            let cur = self.find(hash, |k| eq(i, k))?;
-            *(*outs_ptr).get_unchecked_mut(i) = cur.as_mut();
+            let cur = self.find(hash, |k| eq(i, k)).map(|cur| cur.as_non_null());
+            outs_ptr.add(i).write(cur);
```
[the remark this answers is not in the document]
Urgau: Done, and changed the function to use `array::from_fn` at the same time.

src/map.rs
```diff
@@ -1481,33 +1484,46 @@ where
...
+    ///         Some(&mut 1807),
+    ///         Some(&mut 1800),
+    ///     ],
     /// );
     ///
     /// // Missing keys result in None
     /// let got = libraries.get_many_mut([
-    ///     "Athenæum",
+    ///     "Athenæum1",
```
Amanieu, line 1508: Just noticed this, could you change it back to just `Athenæum` and have the result be `[Some, None]`?

PR #565, opened 2024-10-03
Add Tag(u8) newtype in an attempt to stop using byte-pointers for everything

clarfonthey:
The longer-term goal is to make it so that all the pointers passed around inside the are either `Tag` pointers or `T` pointers, so that we know whether we're using them for indexing into the control or the buckets. Then, `u8` pointers mean that we're referring to a raw allocation, rather than the control bytes.

However, the current code isn't really built for this, and the result is a lot of pointer casts everywhere. I didn't want to just replace `u8` with `Tag` everywhere, since there are some cases where we use `u8` to really mean bytes, and that would be counter to the original purpose.

One short-term gain, however, is that the constant tags and the various methods on them can now be real associated constants and methods, instead of just standalone functions and constants that have to be imported separately.

----

This change also bumps MSRV to 1.65.0. I could bump it higher, but only 1.65 was needed, so, I decided to go with the smallest amount.

### DIFF
--- a/src/raw/mod.rs
+++ b/src/raw/mod.rs
@@ -102,29 +102,54 @@ trait SizedTypeProperties: Sized {
 
 impl<T> SizedTypeProperties for T {}
 
-/// Control byte value for an empty bucket.
-const EMPTY: u8 = 0b1111_1111;
+/// Single tag in a control group.
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+#[repr(transparent)]
+struct Tag(u8);
+impl Tag {
+    /// Control tag value for an empty bucket.
+    const EMPTY: Tag = Tag(0b1111_1111);
 
-/// Control byte value for a deleted bucket.
-const DELETED: u8 = 0b1000_0000;
+    /// Control tag value for a deleted bucket.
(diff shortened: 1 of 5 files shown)


src/raw/mod.rs
```diff
@@ -1492,7 +1498,9 @@ impl RawTableInner {
     const fn new() -> Self {
         Self {
             // Be careful to cast the entire slice to a raw pointer.
-            ctrl: unsafe { NonNull::new_unchecked(Group::static_empty() as *const _ as *mut u8) },
+            ctrl: unsafe {
+                NonNull::new_unchecked(Group::static_empty().as_ptr() as *const u8 as *mut u8)
+            },
```
clarfonthey, line 1503: I wanted to make this `cast_mut().cast()` but `cast_mut` would require bumping the MSRV, and not really sure how folks felt about that.

Amanieu: Rust 1.63 is over 2 years old. Feel free to bump MSRV all the way up to 1.80 (oldstable).

clarfonthey: I'm gonna do that, then.

PR #568, opened 2024-10-06
Further sequester `Group`/`Tag` code

clarfonthey:
Was originally going to make this part of a larger change, but decided this bit is enough for its own PR.

Essentially, this completely moves the `Group` and `Tag` code outside of the `raw` module into a separate `control` module. This module will also eventually be the future home for stuff like `RawIterHash` and `ProbeSeq`, but I decided that that would be better as a separate PR, since moving *and* modifying the files makes it harder to review.

### DIFF
--- a/src/control/bitmask.rs
+++ b/src/control/bitmask.rs
@@ -1,4 +1,4 @@
-use super::imp::{
+use super::group::{
     BitMaskWord, NonZeroBitMaskWord, BITMASK_ITER_MASK, BITMASK_MASK, BITMASK_STRIDE,
 };
 
@@ -102,7 +102,7 @@ impl IntoIterator for BitMask {
 
 /// Iterator over the contents of a `BitMask`, returning the indices of set
 /// bits.
-#[derive(Copy, Clone)]
+#[derive(Clone)]
 pub(crate) struct BitMaskIter(pub(crate) BitMask);
 
 impl Iterator for BitMaskIter {
--- a/src/control/group/mod.rs
+++ b/src/control/group/mod.rs
@@ -1,0 +1,35 @@
+cfg_if! {
+    // Use the SSE2 implementation if possible: it allows us to scan 16 buckets
+    // at once instead of 8. We don't bother with AVX since it would require
+    // runtime dispatch and wouldn't gain us much anyways: the probability of
+    // finding a match drops off drastically after the first few buckets.
+    //
+    // I attempted an implementation on ARM using NEON instructions, but it
+    // turns out that most NEON instructions have multi-cycle latency, which in
+    // the end outweighs any gains over the generic implementation.
+    if #[cfg(all(
+        target_feature = "sse2",
+        any(target_arch = "x86", target_arch = "x86_64"),
+        not(miri),
+    ))] {
+        mod sse2;
+        use sse2 as imp;
+    } else if #[cfg(all(
+        target_arch = "aarch64",
+        target_feature = "neon",
+        // NEON intrinsics are currently broken on big-endian targets.
+        // See https://github.com/rust-lang/stdarch/issues/1484.
+        target_endian = "little",
+        not(miri),
+    ))] {
+        mod neon;
+        use neon as imp;
+    } else {
+        mod generic;
+        use generic as imp;
+    }
+}
+pub(crate) use self::imp::Group;
+pub(super) use self::imp::{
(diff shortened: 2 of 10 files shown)


src/control/group/mod.rs
```diff
@@ -0,0 +1,35 @@
+cfg_if! {
```
clarfonthey, line 1: This is just copied directly from the `raw` module, without any additional changes.

src/control/tag.rs
```diff
@@ -0,0 +1,81 @@
...
+    fn fill_empty(&mut self) {
+        self.fill_tag(Tag::EMPTY)
+    }
+}
+impl TagSliceExt for [Tag] {
+    fn fill_tag(&mut self, tag: Tag) {
+        // SAFETY: We have access to the entire slice, so, we can write to the entire slice.
+        unsafe { self.as_mut_ptr().write_bytes(tag.0, self.len()) }
+    }
+}
```
clarfonthey, line 81: This is new code.

The `TagSliceExt` was a mostly impulse decision to be able to do `slice.fill_empty()` instead of `Tag::EMPTY.fill(slice)`, and since it's solely an internal API anyway it isn't a big deal. The most import part is that this lets us make the inside of a tag private relative to the `raw` module.

src/control/tag.rs
```diff
@@ -0,0 +1,81 @@
...
+            if self.special_is_empty() {
+                f.pad("EMPTY")
+            } else {
+                f.pad("DELETED")
+            }
+        } else {
+            f.debug_tuple("full").field(&(self.0 & 0x7F)).finish()
+        }
+    }
+}
```
clarfonthey, line 64: I added this to replace the built-in debug to help me debug the code I'm going to add in a future PR, so it's a bit easier to read. I figure that if we care about the additional compile time added by this one debug impl so much we can `cfg(debug_assertions)` gate it later.

src/util.rs
```diff
@@ -0,0 +1,14 @@
+// FIXME: Branch prediction hint. This is currently only available on nightly
```
clarfonthey, line 1: I wasn't feeling particularly creative and will need this in the `control` module too in the future, so, I decided to separate them into their own module.

Maybe `nightly` would be a better name, but I don't think it matters that much.

src/control/bitmask.rs
```diff
@@ -102,7 +102,7 @@ impl IntoIterator for BitMask {
 
 /// Iterator over the contents of a `BitMask`, returning the indices of set
 /// bits.
-#[derive(Copy, Clone)]
+#[derive(Clone)]
```
clarfonthey, line 105: Iterators in general shouldn't implement `Copy`, for better or worse.

clarfonthey:
So, I figured out why my probing logic wasn't working: it wasn't because it wasn't working, but because my debug-assert checking that it was working wasn't working, and giving me false positives. So, #578 exists to demonstrate the fuller version of the control module so you can get an idea where I'm coming from with this change, even though this should be merged first.

src/control/tag.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// Fills the control with the given tag.
+    fn fill_tag(&mut self, tag: Tag);
+
+    /// Clears out the control.
+    fn fill_empty(&mut self) {
+        self.fill_tag(Tag::EMPTY)
+    }
+}
+impl TagSliceExt for [Tag] {
+    fn fill_tag(&mut self, tag: Tag) {
```
Amanieu, line 77: This needs `#[inline]`.

src/control/tag.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+/// Extension trait for slices of tags.
+pub(crate) trait TagSliceExt {
+    /// Fills the control with the given tag.
+    fn fill_tag(&mut self, tag: Tag);
+
+    /// Clears out the control.
+    fn fill_empty(&mut self) {
```
Amanieu, line 72: This needs `#[inline]`.

src/control/tag.rs
Amanieu: This isn't a problem for compilation time since it's not a generic function.

src/util.rs
Amanieu: It's actually broken at the moment on nightly until https://github.com/rust-lang/rust/pull/120370 is merged, so it doesn't even do anything. But it's probably fine to keep for now as a reminder to fix this later.

PR #570, opened 2024-10-07
Revert "feat: borsh serde"

dzmitry-lahoda:
Reverts rust-lang/hashbrown#525


kayabaNerve:
This doesn't update the Cargo.toml to remove the borsh feature/dependency from there.

dzmitry-lahoda:
Sorry, will revert. So weird how it happend. I pressed revert in GH on commit which did added both rust and toml changes.

Amanieu:
Can you update this PR to also revert the changes in Cargo.toml?

We will likely release 0.16 after this is merged since this is a breaking change.

PR #572, opened 2024-10-09
ci: test the MSRV with minimal dependency versions

cuviper:
This includes a slight bump of rayon for use of `ParallelIterator::copied`, and `pub(crate)` on `raw::Tag` because older compilers complain about the `pub(crate)` methods that expose it.


Amanieu:
I do think we should at least bump the MSRV to 1.65 since that's what we're actually testing in CI.

cuviper:
Yes, MSRV was already bumped to 1.65 in #565.

PR #575, opened 2024-10-11
Switch from using Bors to merge queue in CI

Kobzol:
Part of ongoing effort of reducing the amount of repos managed by homu (https://github.com/rust-lang/infra-team/issues/169).

I left the conclusion jobs at the end, because they run a large job matrix, so it would be annoying to enumerate all these jobs in the CI config. Instead, I added `conclusion` as a required CI check for the `master` branch in https://github.com/rust-lang/team/pull/1577.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,11 +1,8 @@
 name: Rust
 
 on:
-  push:
-    branches-ignore:
-      - trying.tmp
-      - staging.tmp
   pull_request:
+  merge_group:
 
 jobs:
   miri:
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -90,28 +87,14 @@ jobs:
...
-      - name: Mark the job as successful
-        run: exit 0
-
-  end-failure:
-    name: bors build finished
-    if: github.event.pusher.name == 'bors' && (failure() || cancelled())
+  conclusion:
+    needs: [test, msrv]
+    # !cancelled() executes the job regardless of whether the previous jobs passed or failed
+    if: ${{ !cancelled() }}
```
MarcoIeni, line 93: Can you comment why we need this?
Maybe because it's not executed if previous jobs were skipped?

Kobzol: I copied it from another repo that uses merge queues :D But yes, the issue is that if a job in `needs` would get skipped, `conclusion` would get skipped too, which we do not want. That is essentially what the comment already states, I think.

MarcoIeni: I see. The comment doesn't mention skipped! :)

Kobzol: Good point, expanded the comment.

Kobzol:
@Amanieu This should be ready now. Once it's merged, we will enable merge queues for the repo and remove the `push` trigger.

.github/workflows/rust.yml
```diff
@@ -90,28 +88,15 @@ jobs:
...
-      - name: Mark the job as successful
-        run: exit 0
-
-  end-failure:
-    name: bors build finished
-    if: github.event.pusher.name == 'bors' && (failure() || cancelled())
+  conclusion:
+    needs: [test, msrv]
+    # !cancelled() executes the job regardless of whether the previous jobs passed or failed.
+    # Otherwise, this job would be skipped by default.
```
MarcoIeni, line 94:
```suggestion
    # !cancelled() executes the job regardless of whether the previous jobs passed, failed or get skipped.
```
I think this is clearer, what do you think?

.github/workflows/rust.yml
```diff
@@ -1,11 +1,9 @@
 name: Rust
 
 on:
-  push:
-    branches-ignore:
-      - trying.tmp
-      - staging.tmp
   pull_request:
+  merge_group:
+  push:
```
MarcoIeni, line 6:
```suggestion
```
not needed if we merge manually (without bors)

.github/workflows/rust.yml
Kobzol: It's a bit subtle in that if a job in `needs` fails, **this job** gets skipped, which we want to avoid. So the issue is not so much the previous jobs being skipped, but them failing. But your description is indeed more general. I changed it to that.

MarcoIeni (approved):

.github/workflows/rust.yml
```diff
@@ -90,28 +87,14 @@ jobs:
...
-    needs: [miri, rustfmt_clippy, test, msrv]
-
     steps:
-      - name: Mark the job as a failure
-        run: exit 1
+      - name: Conclusion
+        run: |
+          jq -C <<< '${{ toJson(needs) }}'
+          # Check if all needs were successful.
+          jq --exit-status 'all(.result == "success")' <<< '${{ toJson(needs) }}'
```
MarcoIeni, line 100: copy pasted the same blob we added in other repos
https://github.com/rust-lang/rust-clippy/blob/8bc1a9e49a22b0e43c942102b13ded8704851228/.github/workflows/clippy_pr.yml#L62

PR #578, opened 2024-10-15
Move probing logic into control module

clarfonthey:
This is the follow-up to #568, which should be merged first. (That's why this is marked as a draft.)

Essentially, this pulls the bulk of the probing logic out of `RawTable` and into iterators which can more easily be reasoned with in isolation. The code isn't identical to that in `RawTable`, but it seems to pass all the tests, so, I'm assuming that it all works correctly.


src/control/probe.rs
```diff
@@ -0,0 +1,463 @@
...
+    ///
+    /// This is equivalent to `empty().is_some()`.
+    #[inline]
+    pub(crate) fn has_empty(&self) -> bool {
+        self.group.match_empty().any_bit_set()
+    }
+
+    /// Finds an empty slot in the group to insert a new item.
+    #[inline]
+    pub(crate) fn empty(&self) -> Option<usize> {
```
clarfonthey, line 35: This is a combination of `RawTableInner::find_insert_slot_in_group` and `RawTableInner::fix_insert_slot`.

src/control/probe.rs
```diff
@@ -0,0 +1,463 @@
...
+
+    /// Number of buckets, minus one.
+    bucket_mask: usize,
+}
+impl GroupProbe {
+    /// Checks whether the group contains any empty slots.
+    ///
+    /// This is equivalent to `empty().is_some()`.
+    #[inline]
+    pub(crate) fn has_empty(&self) -> bool {
```
clarfonthey, line 29: This was used in `RawTableInner::find_or_find_insert_slot_inner`, `RawTableInner::find_inner`, and `RawIterHashInner::next`.

src/control/probe.rs
```diff
@@ -0,0 +1,463 @@
...
+
+    /// Number of buckets, minus one.
+    bucket_mask: usize,
+
+    /// Marker indicating that we have items of type T.
+    marker: PhantomData<*const T>,
+}
+impl<T> GroupProbeItems<T> {
+    /// Return the index of the first item that satisfies the predicate.
+    pub(crate) fn find(self, mut f: impl FnMut(&T) -> bool) -> Option<usize> {
```
clarfonthey, line 151: This is taken from `RawTableInner::find_or_find_insert_slot_inner`.

src/control/probe.rs
```diff
@@ -0,0 +1,463 @@
...
+            bucket_mask,
+            #[cfg(debug_assertions)]
+            hash,
+            #[cfg(debug_assertions)]
+            iter: 0,
+        }
+    }
+
+    /// Gets the inner slice of groups.
+    fn control_slice(&self) -> &[Tag] {
```
clarfonthey, line 321: This method basically indicated a bug in the existing group code: we assume that there are always this many initialized tags in the control data, but that's not what `Group::static_empty` gave us: it only gave us a single group, rather than two like this would imply.

So, to help ensure that methods like this always work, and so we don't have to waltz around the concept of "we probably won't read that many groups," the static group allocations were doubled.

src/control/probe.rs
```diff
@@ -0,0 +1,463 @@
...
+        // SAFETY: We always allocate this many control tags.
+        unsafe {
+            slice::from_raw_parts_mut(self.ctrl.as_ptr(), self.bucket_mask + 1 + Group::WIDTH)
+        }
+    }
+
+    /// Verifies that the control slice is valid.
+    fn assert_at_least_one_empty(&self) {
```
clarfonthey, line 329: I added this while I was debugging; ultimately not needed, but it feels nice to have anyway. If you want, I can be more aggressive making sure this code is optimized out.

src/control/probe.rs
```diff
@@ -0,0 +1,463 @@
...
+                // We should have found an empty bucket by now and ended the probe.
+                panic!(
+                    "Went past end of probe sequence searching for hash {:x?} (tag {:x?}, index {:?}) in {:x?}",
+                    self.hash,
+                    self.hash_tag,
+                    (self.hash as usize & self.bucket_mask),
+                    self.control_slice()
+                );
+            }
+            self.iter += 1;
```
clarfonthey, line 443: This is the bug that made me think my code wasn't working at first: I was checking if `self.stride > self.bucket_mask`, when really, I should be checking if `self.stride > self.bucket_mask * Group::WIDTH`.

Except… that can overflow. So, since this is all for a debug assertion anyway, I decided to just add an extra field when debug assertions are enabled that tracks how many iterations have passed. That way, we can debug "infinite" loops, but we don't actually take the performance hit in release mode.

src/raw/mod.rs
```diff
@@ -181,11 +143,6 @@ impl TableLayout {
     }
 }
 
-/// A reference to an empty bucket into which an can be inserted.
```
clarfonthey, line 283: I decided to delete `InsertSlot` because I think it's a bad API: there's not much stopping you from creating an invalid insertion slot, and it doesn't really gain you anything by making the type different from an ordinary index. We can revisit this type of API for `HashTable` if it seems worthwhile in the future.

src/raw/mod.rs
```diff
@@ -4161,45 +3642,17 @@ impl<T> Iterator for RawIterHash<T> {
...
-                self.bitmask = self.group.match_tag(self.tag_hash).into_iter();
+        match self.inner.next() {
+            Some((_, ptr)) => {
+                // FIXME: this is the worst API ever oh no oh dear I hate it
+                // FIXME: use NonNull::add if MSRV > 1.80.0. hopefully this is
+                //   deleted before then…
+                // SAFETY: Bucket wants our pointer to be offset by one, so,
+                //   make it so…
+                let ptr = unsafe { NonNull::new_unchecked(ptr.as_ptr().add(1)) };
+                Some(Bucket { ptr })
```
clarfonthey, line 3653: Shout outs to miri for helping me find this issue.

So, okay. I didn't want to fix *all* usages of `Bucket` in this PR and make them all actually put the pointer in the *correct* place since that's a big change. But I'm also not changing `ProbeItems` to return a pointer to somewhere *other* than where the item is.

If it's actually a big enough deal for performance, I'll try and fix the bucket API to be less bad. But that API is on the chopping block and hopefully I can get to it before we want to do a release anyway.

src/control/probe.rs
```diff
@@ -0,0 +1,469 @@
...
+    iter: usize,
+}
+impl Probe {
+    /// Creates a new probe for a table for a given hash.
+    ///
+    /// # Safety
+    ///
+    /// The given bucket mask must be equal to the power-of-two number of buckets minus one,
+    /// and the given control pointer must be a valid hash table.
+    pub(crate) unsafe fn new(ctrl: NonNull<Tag>, bucket_mask: usize, hash: u64) -> Probe {
```
Amanieu, line 301: All non-generic functions need to have `#[inline]`. And hot generic functions should as well, since it causes them to be replicated in all codegen units they are referenced by instead of relying on LTO to inline them.

src/control/probe.rs
Amanieu: Actually a single group is sufficient. We only need `GROUP_SIZE - 1` trailing bytes, and the static empty table has 1 bucket. Hence a total of `GROUP_SIZE` control bytes.

Also since `control_slice` is only used by debugging code, it only needs to include the first `bucket_mask + 1` control bytes. The trailing bytes are a mirror of the first bytes, or EMPTY in the case of tables smaller than the group size.

src/control/probe.rs
Amanieu: This assert will incorrectly pass on tables smaller than the group size since the tail bytes are EMPTY. Hence the recommendation for `control_bytes` to ignore the trailing bytes.

src/control/probe.rs
```diff
@@ -0,0 +1,469 @@
...
+
+    /// Finds an empty slot to insert something.
+    ///
+    /// Although not a safety violation, this method *will* loop infinitely if
+    /// the table does not contain an empty slot. You have been warned.
+    pub(crate) fn empty(mut self) -> usize {
+        self.assert_at_least_one_empty();
+        loop {
+            // SAFETY: We always return an item from the iterator.
+            let group_probe = unsafe { self.next().unwrap_unchecked() };
```
Amanieu, line 346: Rather than using `unwrap_unchecked` everywhere, it would be better to not use the `Iterator` trait here and just have an inherent `next` method.

src/control/probe.rs
Amanieu, line 1: Some overall thoughts on this file:

This is a *lot* of code that is generic over `T` which really doesn't need to be. This can severely hurt compilation times. It would be better to keep all of the probing logic only using `usize` and have a separate `get`/`get_mut` on the `RawTable` to access elements by index.

Basically, I would expect this entire file to only deal with `usize` indices and not contain any mention of `T`. Perhaps it would be easier if you first ripped out the `Bucket` API and replaced it with indices before cleaning up the internals.

src/control/probe.rs
clarfonthey: So, I agree that we technically only need this many bytes, but all the existing code relies on `num_ctrl_bytes`, which is defined as `bucket_mask + 1 + Group::WIDTH`, not `bucket_mask + Group::WIDTH`. So, in all cases, you need at least two groups.

I'm open to changing this, but I figured that the empty table case wasn't really worth optimising for as long as it didn't allocate anything.

src/control/probe.rs
clarfonthey: Wouldn't the tail bytes being empty satisfy the condition that at least one of the bytes is empty?

src/control/tag.rs
```diff
@@ -30,6 +30,12 @@ impl Tag {
         self.0 & 0x01 != 0
     }
 
+    /// Checks whether a control value is EMPTY.
+    #[inline]
+    pub(crate) const fn is_empty(self) -> bool {
```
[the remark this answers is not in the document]
clarfonthey: Is there a reason why `special_is_empty` doesn't do that, then?

src/control/probe.rs
clarfonthey: I'm not entirely sure if I'm misunderstanding this, but I only made things generic over T when I absolutely had to: the only things that involve `T` either directly return pointers to `T` or require calling functions that take `&T`, both of which can't be made non-generic. The majority of the probing logic isn't generic at all, and operates just on `usize`.

src/control/probe.rs
Amanieu: So I went back and double-checked the implementation. We only need `Group::WIDTH - 1` tail bytes for probing, but we actually need `Group::WIDTH` padding bytes for `set_ctrl` (the last byte is only written but never read), hence the current definition of `num_ctrl_bytes`.

So the correct length for the static allocation should be `1 + Group::WIDTH`.

src/control/probe.rs
Amanieu: The problem is that it would satisfy this condition *even if the table was actually full*.

src/control/tag.rs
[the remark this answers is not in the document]
Amanieu: Yes, there is a good reason for having it implemented this way. In the insertion code we use this code:
```rust
self.table.growth_left -= old_ctrl.special_is_empty() as usize;
```
Because `special_is_empty` only looks at the low bit, this becomes branchless and compiles down to 2 instructions: `and` & `sub`.

src/control/probe.rs
Amanieu: My main concern is that `GroupProbeItems` and `ProbeItems` are generic over `T` and have methods that contain a lot of the searching logic which is going to be replicated for each `T`. I think it would be better if all the code in this file only dealt with `usize` indices and instead let the caller deal with mapping these indices to values in the data array. Similarly methods like `find_full` can take a `FnMut(usize) -> bool` closure where the closure deals with getting the element for a given index.

clarfonthey:
So, I've gotten back into this, and I am realising that `cargo llvm-lines` is returning double what it did for the commit before it, but it's kind of strange, because it's not due to monomorphisation. I'm going to try and investigate a bit.

However, I am going to see what tinier bits of this I can merge in that will be less intrusive, like the double-width static group allocation since that's technically removing some potential UB.

clarfonthey:
Have plans to split this into smaller changes later. But, since this won't be merged as-is, closing.

PR #579, opened 2024-10-17
wip: allow providing the key at insertion time for EntryRef

djugei:
Either explicitly through insert_kv or implicitly through insert_clone.

I have added an assert in insert_kv just to be sure, but if it is only a logic error to insert a different key at an entry or is caught deeper in the stack it should be removed/changed to debug_assert.

If this is something the maintainers are interested in i can implement the EntryRef and OccupiedEntryRef variants and document them.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -2984,10 +2984,10 @@ where
 /// }
 /// assert!(map["b"] == 20 && map.len() == 2);
 /// ```
-pub struct VacantEntryRef<'a, 'b, K, Q: ?Sized, V, S, A: Allocator = Global> {
+pub struct VacantEntryRef<'map, 'key, K, Q: ?Sized, V, S, A: Allocator = Global> {
     hash: u64,
-    key: &'b Q,
-    table: &'a mut HashMap<K, V, S, A>,
+    key: &'key Q,
+    table: &'map mut HashMap<K, V, S, A>,
 }
 
 impl<K, Q, V, S, A> Debug for VacantEntryRef<'_, '_, K, Q, V, S, A>
@@ -4328,7 +4328,25 @@ impl<'a, 'b, K, Q: ?Sized, V: Default, S, A: Allocator> EntryRef<'a, 'b, K, Q, V
     }
 }
 
-impl<'a, 'b, K, Q: ?Sized, V, S, A: Allocator> VacantEntryRef<'a, 'b, K, Q, V, S, A> {
+impl<'map, 'key, K, V, S, A: Allocator> VacantEntryRef<'map, 'key, K, K, V, S, A> {
+    /// insert, cloing the key
+    #[cfg_attr(feature = "inline-more", inline)]
+    pub fn insert_clone(self, value: V) -> &'map mut V
+    where
+        K: Hash + Clone,
+        S: BuildHasher,
+    {
+        let table = &mut self.table.table;
+        let entry = table.insert_entry(
+            self.hash,
+            (self.key.clone(), value),
+            make_hasher::<_, V, S>(&self.table.hash_builder),
+        );
+        &mut entry.1
+    }
+}
+
+impl<'map, 'key, K, Q: ?Sized, V, S, A: Allocator> VacantEntryRef<'map, 'key, K, Q, V, S, A> {
     /// Gets a reference to the key that would be used when inserting a value
     /// through the `VacantEntryRef`.
     ///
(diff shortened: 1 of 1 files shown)


cuviper:
> I have added an assert in insert_kv just to be sure, but if it is only a logic error to insert a different key at an entry or is caught deeper in the stack it should be removed/changed to debug_assert.

I expect we do want a full assert, just like we did in `HashSet::get_or_insert_with`:
https://github.com/rust-lang/hashbrown/blob/eea980411a75c5229a562f9543fe1b035ad384c9/src/set.rs#L957

src/map.rs
```diff
@@ -4378,6 +4396,24 @@ impl<'a, 'b, K, Q: ?Sized, V, S, A: Allocator> VacantEntryRef<'a, 'b, K, Q, V, S
         &mut entry.1
     }
 
+    /// provide explicit key at insert-time instead of relying on there being effectively a from &K to K implementation and not working with cloneable values
+    #[cfg_attr(feature = "inline-more", inline)]
+    pub fn insert_kv(self, key: K, value: V) -> &'map mut V
+    where
+        K: Hash,
+        for<'k> &'k K: PartialEq<&'key Q>,
```
cuviper, line 4404: I think this should use `Q: Equivalent<K>` instead, just like `entry_ref` did when it was created.

src/map.rs
```diff
@@ -4395,9 +4431,9 @@ impl<'a, 'b, K, Q: ?Sized, V, S, A: Allocator> VacantEntryRef<'a, 'b, K, Q, V, S
     /// }
     /// ```
     #[cfg_attr(feature = "inline-more", inline)]
-    pub fn insert_entry(self, value: V) -> OccupiedEntry<'a, K, V, S, A>
+    pub fn insert_entry(self, value: V) -> OccupiedEntry<'map, K, V, S, A>
     where
-        K: Hash + From<&'b Q>,
+        K: Hash + From<&'key Q>,
```
cuviper, line 4436: Not _your_ problem, but this is an issue too, since downstream `Q` might implement a bad `From<&Q> for K`.

djugei: yeah i noticed that too, that was half the reason i asked about the assert

src/map.rs
djugei: ah yes that covers exactly what is needed there. had not seen that trait so far and tbh i have not read a lot of hashbrowns code, i just needed one of those methods to avoid either copying my key more than needed or doing a double lookup.

djugei:
i think i will drop the insert_clone as insert_kv already covers the usecase.
if passing the wrong item is just a logic error i do think that using debug_assert is the correct move, though it is somewhat likely for the compiler to be smart even about the full assert.

should i make the full pr?

cuviper:
I'll defer to @Amanieu to decide on the API design.

djugei:
i disagree with clippy on this one.

src/map.rs
```diff
@@ -4330,7 +4330,25 @@ impl<'a, 'b, K, Q: ?Sized, V: Default, S, A: Allocator> EntryRef<'a, 'b, K, Q, V
     }
 }
 
-impl<'a, 'b, K, Q: ?Sized, V, S, A: Allocator> VacantEntryRef<'a, 'b, K, Q, V, S, A> {
+impl<'map, 'key, K, V, S, A: Allocator> VacantEntryRef<'map, 'key, K, K, V, S, A> {
+    /// insert, cloing the key
```
evmar, line 4334: typo: "cloning"

src/map.rs
```diff
@@ -4380,6 +4398,24 @@ impl<'a, 'b, K, Q: ?Sized, V, S, A: Allocator> VacantEntryRef<'a, 'b, K, Q, V, S
         &mut entry.1
     }
 
+    /// provide explicit key at insert-time instead of relying on there being effectively a from &K to K implementation and not working with cloneable values
+    #[cfg_attr(feature = "inline-more", inline)]
+    pub fn insert_kv(self, key: K, value: V) -> &'map mut V
```
Amanieu, line 4403: I feel that a better name would be `insert_with_key`. This indicates that the key is being explicitly provided.

Amanieu (commented):
Overall, I think the API additions make sense. I've added some comments on the specific.

And finally, the docs will need to be improved.

src/map.rs
djugei: done

djugei:
@Amanieu i have integrated the suggestions. Unsure if the docs are enough, i mainly point to the regular insert docs instead of duplicating.

This should be ready to merge now i think, i have rebased it on top of master.

Amanieu (commented):
I think this can be merged with some updates to the comments and panic message.

Amanieu:
I've cleaned up the doc comment a bit. Thanks for working on this!

PR #586, opened 2024-11-13
mark const fn constructors as rustc_const_stable_indirect

RalfJung:
This is a promise that they can be implemented entirely using stable const features (and the compiler checks this when `-Zforce-unstable-if-unmarked` is set).

Part of https://github.com/rust-lang/rust/issues/102575.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -52,6 +52,7 @@ pub enum DefaultHashBuilder {}
 extern crate std;
 
 #[cfg_attr(test, macro_use)]
+#[cfg_attr(feature = "rustc-dep-of-std", allow(unused_extern_crates))]
 extern crate alloc;
 
 #[cfg(feature = "nightly")]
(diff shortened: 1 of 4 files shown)


Amanieu (approved):

Amanieu:
I assume you need a new hashbrown release with this?

RalfJung:
Yeah to use this in std this will need a release.

cuviper:
I think you won't be able to bootstrap that until the new feature reaches beta for stage0.

RalfJung:
Ah, true, good point.

There's a beta branching in about a week so that's not too long of a wait. :) But I should probably double-check that this will actually work...

RalfJung:
I checked that if I do make these `not(bootstrap)`, and add my local hashbrown as a path dependency to `std`, then everything builds in stage 1. So it should therefore also work with stage 0 after a bootstrap bump.

RalfJung:
The bootstrap bump should happen any day now. @Amanieu would be nice to get a release. :)

Amanieu:
Done

PR #588, opened 2024-11-14
feat: Add filter to HashSet/Table/Map

tugtugtug:
With the removal of the raw table, it is hard to implement an efficient loop to conditionally remove/keep certain fields up to a limit. i.e. a loop that can be aborted and does not require rehash the key for removal of the entry.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -929,48 +929,64 @@ impl<K, V, S, A: Allocator> HashMap<K, V, S, A> {
         }
     }
 
-    /// Retains only the elements specified by the predicate and breaks the iteration when
-    /// the predicate fails. Keeps the allocated memory for reuse.
+    /// Iterates over elements, applying the specified `ControlFlow` predicate to each
     ///
-    /// In other words, remove all pairs `(k, v)` such that `f(&k, &mut v)` returns `Some(false)` until
-    /// `f(&k, &mut v)` returns `None`
+    /// ### Element Fate
+    /// - Kept if `f(&k, &mut v)` returns `ControlFlow::<Any>(true)`
+    /// - Removed if `f(&k, &mut v)` returns `ControlFlow::<Any>(false)`
+    ///
+    /// ### Iteration Control
+    /// - Continue iterating if `f(&k, &mut v)` returns `ControlFlow::Continue`
+    /// - Abort iteration immediately (after applying the element fate) if `f(&k, &mut v)`
+    ///   returns `ControlFlow::Break`
+    ///
     /// The elements are visited in unsorted (and unspecified) order.
     ///
     /// # Examples
     ///
     /// ```
     /// use hashbrown::HashMap;
+    /// use core::ops::ControlFlow;
     ///
     /// let mut map: HashMap<i32, i32> = (0..8).map(|x|(x, x*10)).collect();
     /// assert_eq!(map.len(), 8);
     /// let mut removed = 0;
-    /// map.retain_with_break(|&k, _| if removed < 3 {
+    /// map.filter(|&k, _| if removed < 3 {
     ///     if k % 2 == 0 {
-    ///         Some(true)
+    ///         ControlFlow::Continue(true)
     ///     } else {
     ///         removed += 1;
-    ///         Some(false)
+    ///         ControlFlow::Continue(false)
     ///     }
     /// } else {
-    ///     None
+    ///     // keep this item and break
+    ///     ControlFlow::Break(true)
     /// });
     ///
     /// // We can see, that the number of elements inside map is changed and the
-    /// // length matches when we have aborted the retain with the return of `None`
+    /// // length matches when we have aborted the iteration with the return of `ControlFlow::Break`
     /// assert_eq!(map.len(), 5);
     /// ```
-    pub fn retain_with_break<F>(&mut self, mut f: F)
+    pub fn filter<F>(&mut self, mut f: F)
     where
-        F: FnMut(&K, &mut V) -> Option<bool>,
+        F: FnMut(&K, &mut V) -> core::ops::ControlFlow<bool, bool>,
     {
         // Here we only use `iter` as a temporary, preventing use-after-free
         unsafe {
             for item in self.table.iter() {
                 let &mut (ref key, ref mut value) = item.as_mut();
                 match f(key, value) {
-                    Some(false) => self.table.erase(item),
-                    Some(true) => continue,
-                    None => break,
+                    core::ops::ControlFlow::Continue(kept) => {
+                        if !kept {
+                            self.table.erase(item);
+                        }
+                    }
+                    core::ops::ControlFlow::Break(kept) => {
+                        if !kept {
+                            self.table.erase(item);
+                        }
+                        break;
+                    }
                 }
             }
         }
@@ -5957,20 +5973,21 @@ mod test_map {
     }
 
     #[test]
-    fn test_retain_with_break() {
+    fn test_filter() {
         let mut map: HashMap<i32, i32> = (0..100).map(|x| (x, x * 10)).collect();
-        // looping and removing any key > 50, but stop after 40 iterations
+        // looping and removing any key > 50, but stop after 40 removed
         let mut removed = 0;
-        map.retain_with_break(|&k, _| {
-            if removed < 40 {
-                if k > 50 {
-                    removed += 1;
-                    Some(false)
+        map.filter(|&k, _| {
+            if k > 50 {
+                removed += 1;
+                if removed < 40 {
+                    core::ops::ControlFlow::Continue(false)
                 } else {
-                    Some(true)
+                    // remove this item and break
+                    core::ops::ControlFlow::Break(false)
                 }
             } else {
-                None
+                core::ops::ControlFlow::Continue(true)
             }
         });
         assert_eq!(map.len(), 60);
--- a/src/map.rs
+++ b/src/map.rs
@@ -929,6 +929,53 @@ impl<K, V, S, A: Allocator> HashMap<K, V, S, A> {
         }
     }
 
+    /// Retains only the elements specified by the predicate and breaks the iteration when
+    /// the predicate fails. Keeps the allocated memory for reuse.
+    ///
+    /// In other words, remove all pairs `(k, v)` such that `f(&k, &mut v)` returns `Some(false)` until
+    /// `f(&k, &mut v)` returns `None`
+    /// The elements are visited in unsorted (and unspecified) order.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use hashbrown::HashMap;
+    ///
+    /// let mut map: HashMap<i32, i32> = (0..8).map(|x|(x, x*10)).collect();
+    /// assert_eq!(map.len(), 8);
+    /// let mut removed = 0;
+    /// map.retain_with_break(|&k, _| if removed < 3 {
+    ///     if k % 2 == 0 {
+    ///         Some(true)
+    ///     } else {
+    ///         removed += 1;
+    ///         Some(false)
+    ///     }
+    /// } else {
+    ///     None
+    /// });
+    ///
+    /// // We can see, that the number of elements inside map is changed and the
+    /// // length matches when we have aborted the retain with the return of `None`
+    /// assert_eq!(map.len(), 5);
+    /// ```
+    pub fn retain_with_break<F>(&mut self, mut f: F)
+    where
+        F: FnMut(&K, &mut V) -> Option<bool>,
+    {
+        // Here we only use `iter` as a temporary, preventing use-after-free
+        unsafe {
+            for item in self.table.iter() {
+                let &mut (ref key, ref mut value) = item.as_mut();
+                match f(key, value) {
+                    Some(false) => self.table.erase(item),
+                    Some(true) => continue,
+                    None => break,
+                }
+            }
+        }
+    }
+
     /// Drains elements which are true under the given predicate,
     /// and returns an iterator over the removed items.
     ///
@@ -5907,6 +5954,30 @@ mod test_map {
         assert_eq!(map[&2], 20);
         assert_eq!(map[&4], 40);
         assert_eq!(map[&6], 60);
+    }
+
+    #[test]
+    fn test_retain_with_break() {
+        let mut map: HashMap<i32, i32> = (0..100).map(|x| (x, x * 10)).collect();
+        // looping and removing any key > 50, but stop after 40 iterations
+        let mut removed = 0;
+        map.retain_with_break(|&k, _| {
+            if removed < 40 {
+                if k > 50 {
+                    removed += 1;
+                    Some(false)
+                } else {
+                    Some(true)
+                }
+            } else {
+                None
+            }
+        });
+        assert_eq!(map.len(), 60);
+        // check nothing up to 50 is removed
+        for k in 0..=50 {
+            assert_eq!(map[&k], k * 10);
+        }
     }
 
     #[test]
--- a/src/set.rs
+++ b/src/set.rs
@@ -372,35 +372,45 @@ impl<T, S, A: Allocator> HashSet<T, S, A> {
         self.map.retain(|k, _| f(k));
     }
 
-    /// Retains only the elements specified by the predicate until the predicate returns `None`.
+    /// Iterates over elements, applying the specified `ControlFlow` predicate to each
     ///
-    /// In other words, remove all elements `e` such that `f(&e)` returns `Some(false)` until
-    /// `f(&e)` returns `None`.
+    /// ### Element Fate
+    /// - Kept if `f(&e)` returns `ControlFlow::<Any>(true)`
+    /// - Removed if `f(&e)` returns `ControlFlow::<Any>(false)`
+    ///
+    /// ### Iteration Control
+    /// - Continue iterating if `f(&e)` returns `ControlFlow::Continue`
+    /// - Abort iteration immediately (after applying the element fate) if `f(&e)`
+    ///   returns `ControlFlow::Break`
+    ///
+    /// The elements are visited in unsorted (and unspecified) order.
     ///
     /// # Examples
     ///
     /// ```
     /// use hashbrown::HashSet;
+    /// use core::ops::ControlFlow;
     ///
     /// let xs = [1,2,3,4,5,6];
     /// let mut set: HashSet<i32> = xs.into_iter().collect();
     /// let mut count = 0;
-    /// set.retain_with_break(|&k| if count < 2 {
+    /// set.filter(|&k| if count < 2 {
     ///     if k % 2 == 0 {
-    ///         Some(true)
+    ///         ControlFlow::Continue(true)
     ///     } else {
-    ///         Some(false)
+    ///         ControlFlow::Continue(false)
     ///     }
     /// } else {
-    ///     None
+    ///     // keep this item and break
+    ///     ControlFlow::Break(true)
     /// });
     /// assert_eq!(set.len(), 3);
     /// ```
-    pub fn retain_with_break<F>(&mut self, mut f: F)
+    pub fn filter<F>(&mut self, mut f: F)
     where
-        F: FnMut(&T) -> Option<bool>,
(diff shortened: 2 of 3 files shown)


tugtugtug:
not sure if the name makes sense... maybe `prune` is better?

cuviper:
You can already do this with `extract_if`, because dropping the iterator will keep the remainder.

```rust
    let mut removed = 0;
    for _ in map.extract_if(|&k, _| k % 2 != 0) {
        removed += 1;
        if removed == 3 {
            break;
        }
    }
```

tugtugtug:
> ```rust
> extract_if
> ```

Thanks @cuviper , does this still loop over all hash entries tho? If the intention is to loop over all entries, then the exisiting retain works as well, as one's predicate can just stop returning false, once it returned enough trues.

tugtugtug:
> > ```rust
> > extract_if
> > ```
> 
> Thanks @cuviper , does this still loop over all hash entries tho? If the intention is to loop over all entries, then the exisiting retain works as well, as one's predicate can just stop returning false, once it returned enough trues.

Nvm, I see how it works now, it certainly did not catch my attention how it works :), thought it works as retain, maybe I'll add an example to the documentaion of extract_if to capture this use case.

cuviper:
I suppose a limitation compared to your proposal is that you can only stop `extract_if` on the boundary of an extracted item, after each `Iterator::next()`. So `extract_if` can't shortcut something like "on this condition, keep this item and everything after."

tugtugtug:
> I suppose a limitation compared to your proposal is that you can only stop `extract_if` on the boundary of an extracted item, after each `Iterator::next()`. So `extract_if` can't shortcut something like "on this condition, keep this item and everything after."

Yeah, I was trying to write a test with my use case, and found it awkward if I would like to break out early before the next matched entry.

tugtugtug:
@cuviper do you think then the addition of the API still serves a purpose, the main use case for the new api is to limit the # of iterations per access to the hashmap, and extract_if certainly does not serve that purpose well as discussed.

cuviper:
As a workaround, you could extract that final item to terminate the loop, and then re-insert it. Even with rehashing, that may be faster than continuing to iterate the rest of the map.

Otherwise, yes I do think this API _can_ serve a unique purpose, but `Option<bool>` is kind of awkward to use, and I'm not sure the use case (in that gap of `extract_if`) is common enough to warrant the addition.

If we do add something -- it would be slightly more powerful returning `ControlFlow<bool, bool>`, so you have explicit `Continue` and `Break` signals, and the ability to keep/erase in either case.

tugtugtug:
Now renamed the function to `filter` (by lack of a better name) with `ControlFlow<bool, bool>` as the return value, also beefed up documentation to explain the behaviors of the return combos.

tugtugtug:
@cuviper any feedback with the current PR? As mentioned, I'll work towards any suggestion that would improve this use case.

cuviper:
I don't love the name `filter`, but I don't have a good suggestion right now. Otherwise, the shape of this seems pretty good. Ultimately @Amanieu should decide on new `hashbrown` APIs though.

tugtugtug:
> I don't love the name `filter`, but I don't have a good suggestion right now. Otherwise, the shape of this seems pretty good. Ultimately @Amanieu should decide on new `hashbrown` APIs though.

Yeah, same, I'd be open to any suggestions with the naming. I couldn't come up with anything better, my candidates are, `prune`, `filter`, `retain_with_control`.

Amanieu:
Can't you implement the same functionality by just using a `skip_rest` flag in your closure? Just set that flag to true when you would return `Break` and then cause the closure to return `true` for all remaining elements? The compiler should optimize that into a `break` (do check this though).

tugtugtug:
Thank you @Amanieu!  I just tried it again, and yes, if I just have a simple variable that controls the flow, then it does indeed get optimized by the compiler. However, if I have any sort of indirection, say if the condition is managed by another lambda or another function, then the compiler would not be able to optimize it out.
Imagine the maintainance of the hashmap is based on the stress level of the resources. So the loop is not really based on a constant but rather some calculation. I've made a simple example showing that the compiler optimization is not reliable. The following example can be compiled with "inline-more" and release mode, but would still print "3" multiple times.

```
use hashbrown::HashMap;
fn main() {
    let mut map: HashMap<i32, i32> = (0..8).map(|x|(x, x*10)).collect();
    let mut counter = 0;
    let stop = |counter| {
        if counter == 3 {
            println!("{}", counter);
            return true;
        }
        return false;
    };
    map.retain(|&k, _| {
        if stop(counter) {
            return true;
        }
        if k % 2 == 0 {
            counter += 1;
            return true;
        }
        return false;
    });
    assert!(map.len() >= 3);
}

```

cuviper:
That print is an observable side effect -- the compiler **cannot** skip that by optimization.

tugtugtug:
> That print is an observable side effect -- the compiler **cannot** skip that by optimization.

@cuviper After rerun I found even with the simplest test, it is still failing (so my previous comment wasn't correct):
```
use hashbrown::HashMap;
fn main() {
    let mut map: HashMap<i32, i32> = (0..8).map(|x|(x, x*10)).collect();
    let mut counter = 0;
    let mut counter_true = 0;
    map.retain(|&k, _| {
        if counter == 3 {
            counter_true += 1;
            return true;
        }
        if k % 2 == 0 {
            counter += 1;
            return true;
        }
        return false;
    });
    assert_eq!(counter_true, 1);
}
```

tugtugtug:
> > That print is an observable side effect -- the compiler **cannot** skip that by optimization.
> 
> @cuviper After rerun I found even with the simplest test, it is still failing (so my previous comment wasn't correct):
> 
> ```
> use hashbrown::HashMap;
> fn main() {
>     let mut map: HashMap<i32, i32> = (0..8).map(|x|(x, x*10)).collect();
>     let mut counter = 0;
>     let mut counter_true = 0;
>     map.retain(|&k, _| {
>         if counter == 3 {
>             counter_true += 1;
>             return true;
>         }
>         if k % 2 == 0 {
>             counter += 1;
>             return true;
>         }
>         return false;
>     });
>     assert_eq!(counter_true, 1);
> }
> ```

I guess my point is, if one has to craft their code so careful to make sure the loop ends when it should, it signals the need for this more explicit API, no?

cuviper:
The values of `counter` and `counter_true` can't be changed by optimization either. The way to think about this is to see what the behavior would be without any optimization, and then it should behave the exact same way when optimized -- only faster. So if you're looking at those values after the loop ends, the compiler has to give you the same answer either way.

(That doesn't include explicit debug/release differences like debug-assertions and overflow-checks, which change the code that is produced even before optimization starts.)

If you want the optimizer to see that it can skip the rest of the loop, you should "break" in a way that doesn't modify anything that would be visible outside that loop.

> I guess my point is, if one has to craft their code so carefully to make sure the loop ends when it should, it signals the need for this more explicit API, no?

Maybe, **if** that use case is deemed significant enough to warrant a larger API. I'm sure there are a lot of things we can _imagine_ that would be easier to accomplish with some specialized API, but it doesn't mean we should add them all.

tugtugtug:
@cuviper thank you for your thoughtful comments. I understand your perspective, but I respectfully disagree that users of this library need to have in-depth knowledge of compiler internals to optimize their code. In my experience as shown above, achieving optimal performance can be quite challenging.

> here are a lot of things we can _imagine_ that would be easier to accomplish with some specialized API, but it doesn't mean we should add them all.

I feel this was misunderstood. My intention was not to propose an API that's completely new or unreasonable, but rather to share a specific use case that might not have been considered seriously previously, but was achievable through the raw table apis that are now taken away. If removing the raw table API is in favor of the use of the hashbrown APIs, then I felt there is a gap that should be filled by this extra API.

I think I tried to provide realistic and relevant examples, and I'm open to reasonable workarounds that address the concerns.

cuviper:
I'm also not trying to dismiss your case, just offering a reason why we _might_ not address it. Sometimes the answer can be "no" for other reasons too, like maintenance burden. Either way, this PR is not decided yet!

tugtugtug:
Just to finish the topic about the compiler optimization:
This is the latest I tried, compiled with release mode:
```
use hashbrown::HashMap;
fn main() {
    let mut map: HashMap<i32, i32> = (0..8).map(|x|(x, x*10)).collect();
    let mut counter = 0;
    map.retain(|&k, _| {
        counter += 1;
        if counter >= 3 {
            return true;
        }
        if k % 2 == 0 {
            return true;
        }
        return false;
    });
}
```
This is the assembly from the `counter +=1;`:
```
inc    %edx                                                               
cmp    $0x2,%edx                                                          
jg     0x55555555cd68 <_ZN14test_hashbrown4main17h595336c6b17be939E+584>  
tzcnt  %r9d,%r10d                                                         
shl    $0x3,%r10d                                                         
mov    %r8,%r11                                                           
sub    %r10,%r11                                                          
mov    -0x8(%r11),%r10d                                                   
and    $0x1,%r10d                                                         
je     0x55555555cd68 <_ZN14test_hashbrown4main17h595336c6b17be939E+584>  
mov    %rdi,%r10                                                          
sub    %r11,%r10                                                          
sar    $0x3,%r10                                                          
lea    -0x10(%r10),%r11                                                   
and    %rsi,%r11                                                          
movdqu (%rdi,%r11,1),%xmm1                                                
pcmpeqb %xmm0,%xmm1                                                       
psllw  $0x7,%xmm1                                                         
pmovmskb %xmm1,%ebx                                                       
test   %ebx,%ebx                                                          
jne    0x55555555cd30 <_ZN14test_hashbrown4main17h595336c6b17be939E+528>  
```
I'm not great at assembly code, but afaik, this does not early exit the loop afaik. Use GDB to log the count of the `return true;`, I also saw it hits more than 3 times.

@cuviper do you have more things to try? I could give it a couple of more iterations just to close off this topic.

tugtugtug:
> I'm also not trying to dismiss your case, just offering a reason why we _might_ not address it. Sometimes the answer can be "no" for other reasons too, like maintenance burden. Either way, this PR is not decided yet!

I did feel a bit discouraged as you could see. But again, I would be convinced if any reasonable workarounds are available to normal users like myself, or any suggestions to make this API more acceptable to be included. Having lost the access to the raw table already incurred extra maintenance to the users of the library afaik, but I support that as it would reduce the maintainence cost to this great library.
And I understand the maintainers of the library has the final saying for whatever reasons. I'll accept either way, as I have tried my best.

cuviper:
There's also the fun fact that `counter += 1` can in theory wrap around, so in the optimizer's eyes that may not always remain `>=3`. I tried `counter = counter.saturating_add(1)` though, and that didn't help.

tugtugtug:
> There's also the fun fact that `counter += 1` can in theory wrap around, so in the optimizer's eyes that may not always remain `>=3`. I tried `counter = counter.saturating_add(1)` though, and that didn't help.

True, I mean, we could keep trying, but I feel like this is probably beyond any reasonable effort from a normal user :)

Amanieu:
Overall, I think this ties into the discussion we had a few days ago in the libs-api meeting about `extract_if`, specifically around desires for a more general version of this operation as shown in https://github.com/rust-lang/rust/issues/43244#issuecomment-2495135299. I think the best course of action here would be to design some form of cursor API that would work for both `Vec` and `HashMap`.

tugtugtug:
> Overall, I think this ties into the discussion we had a few days ago in the libs-api meeting about `extract_if`, specifically around desires for a more general version of this operation as shown in [rust-lang/rust#43244 (comment)](https://github.com/rust-lang/rust/issues/43244#issuecomment-2495135299). I think the best course of action here would be to design some form of cursor API that would work for both `Vec` and `HashMap`.

The cursor API looks interesting and promising for the general use cases, however implementation seems rather complex for what this api tries to fix, that is a simple flow control when looping over the elements. I'd say that's overkill, and performance wise I'm not sire that's going to be great comparing a simple loop with a break.

tugtugtug:
> looks interesting and promisin

@Amanieu if you decide against this api after all, could you please give some directions to ppl like me who is basically stuck without the raw table access? I'm against forking unless absolutely necessary, yet I don't see a way forward as the use case is simply removed from the current offerings without much lead time. Even with the cursor API (which could be a long way to go), I'm not sure we will get the same performance. IMO, simplicity is sometimes better than a more general API that seemingly does everything.

Amanieu:
In the absence of a complete cursor API, I think your best bet would be to use `extract_if` to achieve what you want.

Amanieu:
You can use a `Cell<bool>` instead of an `AtomicBool`.

tugtugtug:
> You can use a `Cell<bool>` instead of an `AtomicBool`.

Yes, true, I can profile it to see... but my guess is it is going to suffer, I'll post the results.

tugtugtug:
> map.insert(item.0, item.1);
> 			break;

For everyone's interest:
cargo.toml
```
[package]
name = "test-hashbrown"
version = "0.1.0"
edition = "2021"

[dependencies]
criterion = "0.5.1"
hashbrown = {version="0.15.1"}

[profile.release-with-debug]
inherits = "release"
debug = true

[[bench]]
name = "extract_if_vs_retain_with_control"
harness = false

[patch.crates-io]
hashbrown = { git = "https://github.com/tugtugtug/hashbrown.git", rev = "35b4479" }
```

extract_if_vs_retain_with_control.rs:
```
use criterion::{criterion_group, criterion_main, Criterion};

use std::{cell::Cell, ops::ControlFlow};

use hashbrown::HashMap;

fn a(bench: &mut Criterion) {
	let mut map: HashMap<i32, i32> = (0..1000000).map(|x|(x, x*10)).collect();
	let mut counter = 0;
	let len = map.len();
	let exit : Cell<bool> = Cell::from(false);
	bench.bench_function("extract_if", |b| b.iter(|| {
		for item in map.extract_if(|&k, _| {
				counter += 1;
				if counter >= len/2 {
					exit.set(true);
					return true;
				}
				if k % 2 == 0 {
					return true;
				}
				return false;
			}) {
			if exit.get() {
				map.insert(item.0, item.1);
				break;
			}
		}
	}));
}

fn b(bench: &mut Criterion) {
	let mut map: HashMap<i32, i32> = (0..1000000).map(|x|(x, x*10)).collect();
	let mut counter = 0;
	let len = map.len();
	bench.bench_function("filter", |b| b.iter(|| {
		map.filter(|&k, _| {
			counter += 1;
			if counter >= len/2 {
				return ControlFlow::Break(true);
			}
			if k % 2 == 0 {
				return ControlFlow::Continue(true);
			}
			return ControlFlow::Continue(false);
		});
	}));
}

criterion_group!(benches, a, b);
criterion_main!(benches);
```

results:
```
Running benches/extract_if_vs_retain_with_control.rs (target/release/deps/extract_if_vs_retain_with_control-f8113af52880ae3d)
Gnuplot not found, using plotters backend
extract_if              time:   [21.256 ns 21.334 ns 21.436 ns]
                        change: [-8.0738% -7.7875% -7.4780%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 20 outliers among 100 measurements (20.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  18 (18.00%) high severe

filter                  time:   [970.39 ps 976.80 ps 984.13 ps]
                        change: [-0.0616% +0.5815% +1.2798%] (p = 0.09 > 0.05)
                        No change in performance detected.
Found 21 outliers among 100 measurements (21.00%)
  4 (4.00%) low mild
  5 (5.00%) high mild
  12 (12.00%) high severe
```

I'm not sure if things are optimized out in the filter case, but it does seem to be significantly faster according to this test.

Amanieu:
You can also avoid re-inserting the last item if you return `false` after setting the flag. That should make the performance of `extract_if` match the performance of `filter`.

tugtugtug:
> You can also avoid re-inserting the last item if you return `false` after setting the flag. That should make the performance of `extract_if` match the performance of `filter`.
I'm not sure that's technically the matching use case anymore. Returning false with `exit` would make the loop continue until the next return true. In this example, it may be the next item, in reality it can be many more items iterated, which defeats the entire purpose of the flow control. Did i misunderstand how `extract_if` works?

tugtugtug:
lol, I am really bad at profiling, now I removed the re-insert, and also removed the need to recreate the flow control variables, I think this is the real deal.
```
use criterion::{black_box, criterion_group, criterion_main, BatchSize, Criterion};

use std::{cell::Cell, ops::ControlFlow};

use hashbrown::HashMap;

fn a(bench: &mut Criterion) {
	let map: HashMap<i32, i32> = (0..1000000).map(|x|(x, x*10)).collect();

	bench.bench_function("extract_if", |b|
		b.iter_batched(|| map.clone(), |mut map| {
			let mut counter = 0;
			let len = map.len();
			let exit : Cell<bool> = Cell::from(false);
			for item in map.extract_if(|&k, _| {
					counter += 1;
					if counter >= len/2 {
						exit.set(true);
						return true;
					}
					if k % 2 == black_box(0) {
						// opposite to retain/filter
						return false;
					}
					// opposite to retain/filter
					return true;
				}) {
				if exit.get() {
					//map.insert(item.0, item.1);
					break;
				}
			}
			assert!(map.len() > 0);
		}, BatchSize::NumBatches(10000))
	);
}

fn b(bench: &mut Criterion) {
	let map: HashMap<i32, i32> = (0..1000000).map(|x|(x, x*10)).collect();
	let break_true = ControlFlow::Break(true);
	let continue_true = ControlFlow::Continue(true);
	let continue_false = ControlFlow::Break(false);
	bench.bench_function("filter", |b|
		b.iter_batched(|| map.clone(), |mut map| {
			let mut counter = 0;
			let len = map.len();
			map.filter(|&k, _| {
				counter += 1;
				if counter >= len/2 {
					return break_true;
				}
				if k % 2 == black_box(0) {
					return continue_true;
				}
				return continue_false;
			});
			assert!(map.len() > 0);
		}, BatchSize::NumBatches(10000))
	);
}

criterion_group!(benches, a, b);
criterion_main!(benches);
```
Result:
```

     Running benches/extract_if_vs_retain_with_control.rs (target/release/deps/extract_if_vs_retain_with_control-f8113af52880ae3d)
Gnuplot not found, using plotters backend
extract_if              time:   [3.9801 ms 3.9931 ms 4.0083 ms]
                        change: [+0.1292% +0.5393% +0.9372%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 12 outliers among 100 measurements (12.00%)
  2 (2.00%) high mild
  10 (10.00%) high severe

filter                  time:   [448.85 ns 469.59 ns 492.91 ns]
                        change: [-1.6250% +3.5311% +9.6972%] (p = 0.22 > 0.05)
                        No change in performance detected.
Found 10 outliers among 100 measurements (10.00%)
  7 (7.00%) high mild
  3 (3.00%) high severe
```
@Amanieu it seems filter performance is still significantly faster.

tugtugtug:
Ironically, I ended up writing a trait for `HashMap` to wrap `extract_if` as none of my team likes the awkardness of using it. There you go, we don't have the proper API support from `HashMap`, but you can use this trait to patch that yourself for now.

```
pub(crate) trait RetainWithFlowControl<K, V, S>
where
	S: BuildHasher,
{
	fn retain_with_flow_control<F: FnMut(&K, &mut V) -> ControlFlow<bool, bool>>(&mut self, f: F);
}

impl<K, V, S> RetainWithFlowControl<K, V, S> for HashMap<K, V, S>
where
	K: Eq + Hash,
	S: BuildHasher,
{
	#[inline]
	fn retain_with_flow_control<F: FnMut(&K, &mut V) -> ControlFlow<bool, bool>>(&mut self, mut f: F) {
		let done = Cell::new(ControlFlow::Continue(false));
		for item in self.extract_if(|k, v| -> bool {
			let ret = f(k, v);
			match ret {
				ControlFlow::Break(true)
				| ControlFlow::Break(false) => {
					done.set(ret);
					// we must return true to abort the loop
					true
				},
				// the logic is reversed
				ControlFlow::Continue(ret) => !ret,
			}
		}) {
			if let ControlFlow::Break(ret) = done.get() {
				if !ret {
					// the last item is inserted back when the retain is matched
					self.insert(item.0, item.1);
				}
				break;
			}
		}
	}
}
```

PR #590, opened 2024-11-19
ci: add support for loongarch64-unknown-linux-gnu

heiher:


cuviper:
It should already be *supported* even if it's not explicitly tested, just through the generic implementations. The other arches that have particular CI testing are because they have specialized implementations, like for neon or sse, but plenty of other arches work fine as well.

Amanieu:
I agree with @cuviper, it doesn't make sense to add CI for loongarch64 unless you're also planning on implementing SIMD support for it like we do for ARM & x86.

Amanieu (approved):

PR #593, opened 2024-11-29
Change OccupiedEntry/RawOccupiedEntryMut lifetimes

ChosenName:
Changes the lifetimes returned by `key(_mut)`, `get(_mut)` and `entry(_mut)` to `'a`. This makes the returned value valid for the lifetime of the hash map which makes returning them valid.

## Example
```rust
use hashbrown::HashMap;

fn get_and_insert<'a>(map: &'a mut HashMap<String, String>, key: String, value: String) -> &'a String {
    map.raw_entry_mut()
        .from_key(&key)
        .insert(key, value)
        get()
}
```

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -3816,7 +3816,7 @@ impl<'a, K, V, S, A: Allocator> OccupiedEntry<'a, K, V, S, A> {
     /// }
     /// ```
     #[cfg_attr(feature = "inline-more", inline)]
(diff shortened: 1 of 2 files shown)


Amanieu:
This is incorrect because it would cause the returned lifetime to still be valid even after the entry is removed.

clarfonthey:
Yeah, this is why you need `into_key_mut` to get the original lifetimes back.

PR #596, opened 2024-12-07
Double the width of static allocations for empty tables

clarfonthey:
See this discussion here: https://github.com/rust-lang/hashbrown/pull/578#discussion_r1800275708

Effectively, the current code will actually need `Group::WIDTH + 1` bytes for an empty table, but we only allocate `Group::WIDTH` bytes for it. This could potentially invoke UB based upon the invariants we set in the table, but because we happen to short-circuit empty tables in just the right ways in the actual code, we never actually trigger it.

Since the static allocations are aligned to the group width, we double the size returned, since it would otherwise just add that much extra padding anyway. And since the size of the static allocations added by this is effectively insignificant (8-16 bytes depending on the implementation chosen), it's fair to just do this anyway to make the code more robust.

### DIFF
--- a/src/control/group/generic.rs
+++ b/src/control/group/generic.rs
@@ -50,20 +50,23 @@ impl Group {
     /// Number of bytes in the group.
     pub(crate) const WIDTH: usize = mem::size_of::<Self>();
 
+    /// Double the group width; size of [`Group::static_empty`].
+    pub(crate) const DOUBLE_WIDTH: usize = Group::WIDTH * 2;
+
     /// Returns a full group of empty tags, suitable for use as the initial
     /// value for an empty hash table.
     ///
     /// This is guaranteed to be aligned to the group size.
     #[inline]
-    pub(crate) const fn static_empty() -> &'static [Tag; Group::WIDTH] {
+    pub(crate) const fn static_empty() -> &'static [Tag; Group::DOUBLE_WIDTH] {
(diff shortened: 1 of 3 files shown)


Amanieu:
I still don't think this is needed: the only place that needs `WIDTH` instead of `WIDTH - 1` trailing control bytes is `set_ctrl`, and that requires write access which doesn't work with the static empty table anyways. The only other places which access the trailing control bytes are where `Group::load` is called and that only accesses `WIDTH - 1` trailing bytes since the input index is always less than `num_buckets`.

clarfonthey:
I think perhaps the biggest issue here is that `bucket_mask` can't differentiate between zero and one element, and so maybe I'm trying to jump the gun a little too soon to make `bucket_mask` dictate the size properly when we really should be storing `len` instead. But also, I'm not sure that's right either.

I think that we should be consistent in making sure what guarantees the table actually stores, but you're right that this is probably the wrong approach. I'll close this for now.

Amanieu:
I do want to keep `bucket_mask` as it is because it is part of the critical path for hash table lookups. Every CPU cycle of latency between calculating a hash value and loading the control bytes matters.

clarfonthey:
I mean, that's fair, and it's been long enough that I've looked into the code that I really should go back to figuring out why exactly I wanted this.

PR #601, opened 2024-12-17
HashMap: #or_default_entry and #or_insert_entry

braddunbar:
I often find myself in a situation where I would like to insert a key into a map via `#entry_ref` and then use the key from the resulting entry.

```rust
// Get a byte slice from a buffer
let key = client.request.get(index);

// Insert the value (default)
let mut entry = match queues.entry_ref(&key) {
    EntryRef::Occupied(entry) => entry,
    EntryRef::Vacant(entry) => entry.insert_entry(Default::default()),
};

// Use the value
entry.get_mut().insert_back(client.id);

// Reuse the key instead of copying the bytes again
keys.insert(client.id, entry.key());
```

This is common enough that I'd love to have functions for it, similar to `insert_entry`.

```rust
// Get a byte slice from a buffer
let key = client.request.get(index);

// Insert the value (default)
let mut entry = queues.entry_ref(&key).or_default_entry();

// Use the value
entry.get_mut().insert_back(client.id);

// Reuse the key instead of copying the bytes again
keys.insert(client.id, entry.key());
```

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -6229,7 +6229,7 @@ mod test_map {
             }
 
             for (k, v) in map {
-                println!("{}, {}", k, v);
+                println!("{k}, {v}");
             }
         }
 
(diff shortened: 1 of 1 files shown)


djugei:
do you mean queues.insert instead of keys.insert in the last line of your examples?

braddunbar:
Hey @djugei! I actually meant them as two separate hash maps. The idea is that I'd like to access a key without copying the lookup value via `entry_ref`, and then reuse that key to insert into another hash map.

It's really just a shortcut to insert if necessary and return me the entry to use.

PR #605, opened 2025-01-05
Added Allocator template argument for rustc_iter

brunogouveia:
This change will allow `rustc_iter` to be called on `IntoIter` and `Drain` that are parameterised with custom allocator types.

It seems these two implementations were missed on the initial work to support custom allocators #133.


Amanieu:
Oops I forgot we're not using bors any more :)

Amanieu (approved):

Issue #564, opened 2024-10-03
Hashbrown fails to compile if 0.14.x and 0.15.x are in the same project, only one with the nightly feature


orlp:
If a project (most likely indirectly through dependencies) depends on both version 0.14.x and 0.15.x of hashbrown, and only one of the two dependencies has the `nightly` feature flag set, the project will fail to compile due to using the `allocator_api` without enabling the unstable feature for it:

```rust
error[E0658]: use of unstable library feature 'allocator_api'
    --> /Users/orlp/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hashbrown-0.14.5/src/raw/mod.rs:4329:23
     |
4329 |                 alloc.deallocate(ptr, layout);
     |                       ^^^^^^^^^^
     |
     = note: see issue #32838 <https://github.com/rust-lang/rust/issues/32838> for more information
     = help: add `#![feature(allocator_api)]` to the crate attributes to enable
     = note: this compiler was built on 2024-10-02; consider upgrading it if it is out of date
```

This can be reproduced by making a clean new project and adding the following two lines to `Cargo.toml`:

```toml
hashbrown_14 = { package = "hashbrown", version = "=0.14.5", features = [] }
hashbrown_15 = { package = "hashbrown", version = "=0.15.0", features = ["nightly"] }
```

If it's just one of the two the project compiles fine, regardless of the `nightly` feature flag. If neither or both have the `nightly` feature flag everything compiles fine as well. However if one has the `nightly` feature flag and the other does not, the project doesn't compile.

orlp:
The problem appears to be that `allocator-api2`'s recommended usage is inherently flawed:

> Your library MAY provide a feature that will enable "allocator-api2/nightly". When this feature is enabled, your library MUST enable unstable `#![feature(allocator_api)]` or it may not compile. If feature is not provided, your library may not be compatible with the rest of the users and cause compilation errors on nightly channel when some other crate enables "allocator-api2/nightly" feature.

Feature unification makes the requirement "When this feature is enabled your library MUST enable unstable `#![feature(allocator_api)]` or it may not compile" fundamentally impossible to fulfill.

Amanieu:
I'm not sure there's anything actionable we can do on hashbrown's side...

cuviper:
[...]

Issue #483, opened 2023-11-10
`hashbrown` fails to compile as a transitive dependency


gootorov:
Hi

When using certain crates (repro below) it is possible for `hashbrown` to fail to compile, even though it is a transitive dependency.

To reproduce, create a new crate: `cargo new --lib hashbrown_issue`
With the following `Cargo.toml`:
```Toml
[package]
name = "hashbrown_issue"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
sqlx = { version = "0.7.2", features = ["runtime-tokio-rustls"], default-features = false }
blink-alloc = { version = "0.3.1", features = ["nightly"] }
```
And the following `src/lib.rs`:
```Rust
#![feature(allocator_api)]
```

Then, attempt to compile: `cargo +nightly build`. This produces a lot of similar errors:
```Rust
error[E0658]: use of unstable library feature 'allocator_api'
    --> /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hashbrown-0.14.2/src/raw/mod.rs:4248:23
     |
4248 |                 alloc.deallocate(ptr, layout);
     |                       ^^^^^^^^^^
     |
     = note: see issue #32838 <https://github.com/rust-lang/rust/issues/32838> for more information
     = help: add `#![feature(allocator_api)]` to the crate attributes to enable

For more information about this error, try `rustc --explain E0658`.
error: could not compile `hashbrown` (lib) due to 292 previous errors
```

The toolchain used is `cargo 1.75.0-nightly (7046d992f 2023-11-08)`.

This might be related to #417. To be honest, I'm not too sure whether the issue should be opened in `hashbrown`, or `sqlx`/`blink-alloc`/`allocator-api2`.

#### Possible workarounds
* Adding `hashbrown = { version = "0.14.2", features = ["nightly"] }` to `Cargo.toml` resolves the issue. Even if `hashbrown` is not used as a direct dependency.
However, this is undesirable as it enables a lot of unstable features in `hashbrown`. Limiting to `#![feature(allocator_api)]` is preferable.
* Replacing `blink-alloc` with `bumpalo = { version = "3.14.0", features = ["allocator_api"] }` also works.

gootorov:
cc @zakarumych 

Your help may be needed here

zakarumych:
[...]

PR #606, opened 2025-01-27
Don't enable `allocator-api2/nightly` feature

NeuralModder:
See #564 for rationale. Tentatively fixes said issue and #483, though the root cause is up to the `allocator-api2` crate to fix.

Direct use of `allocator-api2` in tests had to be replaced with the indirection used elsewhere in the crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ default = ["default-hasher", "inline-more", "allocator-api2", "equivalent", "raw
 
 # Enables use of nightly features. This is only guaranteed to work on the latest
 # version of nightly Rust.
-nightly = ["allocator-api2?/nightly", "bumpalo/allocator_api"]
+nightly = ["bumpalo/allocator_api"]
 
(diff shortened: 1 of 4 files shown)


Amanieu:
I'm not sure what's up with miri but it seems related to the allocator API. I'm away this week so I can't look into it myself.

NeuralModder:
@Amanieu I fixed the miri (actually a cargo test) thing, the tests were using `allocator-api2` directly but there must've been some kind of commingling when nightly was enabled, replacing the calls to `allocator-api2` with calls to `crate::raw` fixed it.

What's failing now is `tests/set.rs` due to a linker warning, but only on `aarch64-apple-darwin` (or at least, not on linux), possibly only on nightly. My best guess is that this a platform-specific linker oddity, but of course, I have no idea why this PR would cause it. I unfortunately do not have a cross-compilation environment set up to debug this, either.

Amanieu (approved):

PR #609, opened 2025-03-02
Use `criterion` for benchmark

ldm0:
Supercedes #243 
Related to #4

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ equivalent = { version = "1.0", optional = true, default-features = false }
 lazy_static = "1.4"
 rand = { version = "0.8.3", features = ["small_rng"] }
 rayon = "1.2"
+criterion = "0.5"
 fnv = "1.0.7"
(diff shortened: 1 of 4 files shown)


ldm0:
@Amanieu done

cuviper:
Personally, I've come to prefer [divan](https://crates.io/crates/divan), but it's not terribly important either way.

.github/workflows/rust.yml
```diff
@@ -82,7 +82,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - uses: dtolnay/rust-toolchain@nightly
-    - uses: dtolnay/rust-toolchain@1.65.0 # MSRV
+    - uses: dtolnay/rust-toolchain@1.74.0 # MSRV
```
Amanieu, line 85: One of our dev-dependencies needs 1.81. We can just raise the MSRV: since the cargo resolver now takes this into account it will no longer break people's builds.

PR #619, opened 2025-04-29
Add release-plz for automated releases

Amanieu:


.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,60 @@
+name: Release-plz
+
+permissions:
+  pull-requests: write
+  contents: write
+
```
marcoieni, line 6:
```suggestion
```
this is already specified for every single job

PR #622, opened 2025-05-08
Add missing test for set

AbeZbm:
Hi,

Thanks for your time to review this PR.

We are researchers focusing on generating Rust tests by LLM to statisfy specific execution paths in functions. By examing the existing code, we found that one test can be added to improve the repo's overall test coverage.
The code region we covered is:
https://github.com/rust-lang/hashbrown/blob/62634581de016626f31e1d993b504fe7db1b39f1/src/set.rs#L1630-L1633

The current doc test only covers rhs.len() >= self.len(). To keep documentation example concise, we add a unit test for the rhs.len() < self.len() case.

Thanks again for reviewing.

### DIFF
--- a/src/set.rs
+++ b/src/set.rs
@@ -3102,4 +3102,20 @@ mod test_set {
             set.get_or_insert_with(&key, |_| value);
         }
     }
+
+    #[test]
+    fn test_sub_assign() {
+        let mut a: HashSet<_> = vec![1, 2, 3, 4].into_iter().collect();
+        let b: HashSet<_> = vec![3, 5].into_iter().collect();
+
+        a -= &b;
+
+        let mut i = 0;
+        let expected = [1, 2, 4];
+        for x in &a {
+            assert!(expected.contains(x));
+            i += 1;
+        }
+        assert_eq!(i, expected.len());
+    }
 }


clarfonthey:
While I'll admit the bar for quality is substantially lower for tests, by default I'm not happy about the idea of using LLMs at all for code.

The code itself *works* and is okay. It is true that `SubAssign` is technically untested and that should probably be rectified, but I would rather focus on the proper development of a test suite than haphazard generation of tests using lying machines.

Of course, I'm not a maintainer, so, this only counts as guidance here. But I would much rather not see time wasted reviewing tests by machines so prone to failure. Tests are only good if the person writing them knows what they're doing, and simply passing isn't enough; the goal is to codify correctness, not simply the current behaviour.

(This test in particular checks all the boxes as "okay." I just would not like to see more like it which potentially don't.)

AbeZbm:
@clarfonthey Thank you for taking the time to review and share your valuable feedback!

We fully understand your concerns about LLM-generated test quality. As a research team, our core work focuses precisely on generating effective tests by LLM, the submitted test case represents a manually vetted valid case that survived our review process, not raw generated output. We wholeheartedly agree with the principle thad "tests should codify correctness rather than simply pass".

Thank you again for highlighting these important considerations.

PR #625, opened 2025-06-04
Remove optional dependency on compiler-builtins

tgross35:
Since `hashbrown` always has `extern crate alloc`, it will get `compiler-builtins` via `alloc`'s dependency. There is no longer any need to depend on the crates.io version.


Amanieu:
Do you need a release for this now?

tgross35:
It's not needed until https://github.com/rust-lang/rust/pull/141993 actually merges, but sometime within the next day or two would be ideal so the dep can be dropped not too long after.

PR #633, opened 2025-08-04
Change the default hasher to rapidhash

hoxxep:
Hello,

I've recently released [rapidhash](https://github.com/hoxxep/rapidhash) v3.0.0. It uses the same integer hashing as foldhash, but comprehensively beats the bytes/string hashing performance on all platforms we've benchmarked (common server chips and M1 Max). It offers the same minimal DoS resistance and portability too.

Aside from marginally increased code size and a higher MSRV (rapidhash=1.77, foldhash=1.60) I believe there are no other downsides to changing the default to rapidhash.

If the MSRV is an issue, I can take a look at reducing it? It's only this high because the hashing algorithm is fully `const`, and efficiently reading/splitting slices in const is difficult with the older versions.

Full benchmarks are available in the README and [docs](https://github.com/hoxxep/rapidhash/tree/master/docs) folder. Very happy to answer any questions or hear feedback on rapidhash. Thank you!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Changed the default hasher to rapidhash. (#563)
+
 ## [0.15.4](https://github.com/rust-lang/hashbrown/compare/v0.15.3...v0.15.4) - 2025-06-05
 
 ### Changed
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ rust-version = "1.65.0"
 
 [dependencies]
 # For the default hasher
-foldhash = { version = "0.1.2", default-features = false, optional = true }
+rapidhash = { version = "3.0.0", default-features = false, optional = true }
 
 # For external trait impls
 rayon = { version = "1.2", optional = true }
@@ -62,10 +62,10 @@ rustc-dep-of-std = [
 # Enables the deprecated RawEntry API.
 raw-entry = []
 
-# Provides a default hasher. Currently this is foldhash but this is subject to
+# Provides a default hasher. Currently this is rapidhash but this is subject to
 # change in the future. Note that the default hasher does *not* provide HashDoS
 # resistance, unlike the one in the standard library.
-default-hasher = ["dep:foldhash"]
+default-hasher = ["dep:rapidhash"]
 
 # Enables usage of `#[inline]` on far more functions than by default in this
 # crate. This may lead to a performance increase but often comes at a compile
--- a/README.md
+++ b/README.md
@@ -26,8 +26,8 @@ in environments without `std`, such as embedded systems and kernels.
 ## Features
 
 - Drop-in replacement for the standard library `HashMap` and `HashSet` types.
-- Uses [foldhash](https://github.com/orlp/foldhash) as the default hasher, which is much faster than SipHash.
-  However, foldhash does *not provide the same level of HashDoS resistance* as SipHash, so if that is important to you, you might want to consider using a different hasher.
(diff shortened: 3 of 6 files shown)


Amanieu:
I'm always happy to switch to a better hash function. However I would like to give @orlp (author of foldhash) an opportunity to review this first.

hoxxep:
Absolutely, I'm very grateful to @orlp for the work he's done! The rapidhash v3.0 release is effectively foldhash's integer hashing approach, but replaces the `Hasher::write` with a carefully tuned rapidhash-derived alternative that appears to beat foldhash on almost all string hashing benchmarks.

It took me months to match foldhash's performance. There were many small subtleties fighting against the inliner, reducing the hashmap size, faster global seeding, the integer buffer, avoiding register clobbering, and keeping the avalanching logic in `Hasher::finish` as small and simple as possible. A lot of credit goes to the novel ideas in foldhash and how well it works with the rust hashing traits.

@orlp Any feedback on rapidhash itself, or how to roll it out so that people can benefit from it would be great too. I think it's strictly an upgrade on foldhash, but would happily take any feedback or critique. Also happy to chat offline.

orlp:
A quick status update: with some changes to inlining the largest gaps between rapidhash and foldhash are already closed. I'm still tweaking the string hash a bit further but expect a release probably somewhere next weekend.

hoxxep:
Agreed foldhash v0.2.0 mostly closes the gap. Comparing to rapidhash v4.0.0 there's a marginal improvement on [16,48] length inputs, but it's only around 15% faster.

image

On the foldhash benchmark suite, the main differences are uuid and tenkilobyte inputs. Everything else is fairly equal.
```
realworld/StrUuid/hashonly-struuid-rapidhash-f                                                                             
                        time:   [2.4554 ns 2.4633 ns 2.4727 ns]
realworld/StrUuid/hashonly-struuid-foldhash-fast                                                                             
                        time:   [3.0297 ns 3.0361 ns 3.0439 ns]
realworld/StrUuid/lookupmiss-struuid-rapidhash-f                                                                             
                        time:   [5.5069 ns 5.5526 ns 5.5984 ns]
realworld/StrUuid/lookupmiss-struuid-foldhash-fast                                                                             
                        time:   [5.9742 ns 6.0190 ns 6.0639 ns]
realworld/StrUuid/lookuphit-struuid-rapidhash-f                                                                             
                        time:   [8.5505 ns 8.5899 ns 8.6297 ns]
realworld/StrUuid/lookuphit-struuid-foldhash-fast                                                                             
                        time:   [9.3978 ns 9.4323 ns 9.4674 ns]
realworld/StrUuid/setbuild-struuid-rapidhash-f                                                                            
                        time:   [126.76 µs 127.10 µs 127.48 µs]
realworld/StrUuid/setbuild-struuid-foldhash-fast                                                                            
                        time:   [135.98 µs 136.28 µs 136.60 µs]
```

hoxxep:
I've added a PR #641 to bump foldhash to v0.2.0 since the performance is so similar.

The thumbv6m test fails here because `RandomState::default()` is not implemented for types where it wouldn't give a random result. If required, I can use compile-time randomness instead and add this back in via v4.0.1.

As for MSRV, rapidhash reduced its MSRV to 1.71.0 with v4.0.0, while hashbrown is 1.65.0.

clarfonthey:
Worth adding on here that the MSRV was bumped to 1.85 recently, so, that shouldn't be an issue any more if you want to continue pursuing this.

clarfonthey:
Going to close for now since we need to update the benchmarks, and as part of that, we're going to want to try out a bunch of hasher crates, including rapidhash. Feel free to post in #713 to continue the discussion; I think we're going to hold off until the benchmarks are updated before we consider updating the default hasher.

PR #641, opened 2025-08-26
perf: Bump foldhash to 0.2.0

hoxxep:
Following up from the rapidhash PR #633, this bumps foldhash to `0.2.0` for a large improvement in foldhash's string hashing.

Both rapidhash and foldhash are equal in the hashbrown benchmark suite as it only tests the hashing of `usize` objects.

The reason foldhash 0.2.0 required a major version bump is because `FoldHasher` now has a lifetime parameter. I don't know how users are expected to use `DefaultHashBuilder`, but noting that it likely makes this a breaking change for hashbrown?

cc. @orlp @Amanieu from #633

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Bump foldhash, the default hasher, to 0.2.0.
+
 ## [0.15.5](https://github.com/rust-lang/hashbrown/compare/v0.15.4...v0.15.5) - 2025-08-07
 
 ### Added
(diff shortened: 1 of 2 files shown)


jqnatividad:
FYI... I also got this compilation error when I attempted to use this PR on [patch.crates-io] of my project's Cargo.toml

```
error[E0204]: the trait `core::marker::Copy` cannot be implemented for this type
  --> /Users/jdoe/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hashlink-0.10.0/src/lib.rs:19:17
   |
19 | #[derive(Clone, Copy, Default, Debug)]
   |                 ^^^^
20 | pub struct DefaultHashBuilder(hashbrown::DefaultHashBuilder);
   |                               ----------------------------- this field does not implement `core::marker::Copy`
```

PR #643, opened 2025-08-28
Change `DefaultHashBuilder` to an opaque newtype

cuviper:
This hides the underlying hasher as an implementation detail, so it's no
longer a public dependency. A newtype `DefaultHasher` is also added here
to wrap the underlying `<_ as BuildHasher>::Hasher` type.

This is a breaking change now, requiring a semver bump, but it will
allow later changes to the hasher without breaking.

### DIFF
--- a/src/hasher.rs
+++ b/src/hasher.rs
@@ -63,6 +63,13 @@ impl Hasher for DefaultHasher {
         write_isize(isize),
     }
 
+    // feature(hasher_prefixfree_extras)
+    #[cfg(feature = "nightly")]
+    forward_writes! {
+        write_length_prefix(usize),
+        write_str(&str),
+    }
+
     #[inline(always)]
     fn finish(&self) -> u64 {
         self.inner.finish()
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ default = ["default-hasher", "inline-more", "allocator-api2", "equivalent", "raw
 
 # Enables use of nightly features. This is only guaranteed to work on the latest
 # version of nightly Rust.
-nightly = ["bumpalo/allocator_api"]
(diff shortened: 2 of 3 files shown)


src/hasher.rs
```diff
@@ -0,0 +1,70 @@
...
+    fn build_hasher(&self) -> Self::Hasher {
+        DefaultHasher {
+            inner: self.inner.build_hasher(),
+        }
+    }
+}
+
+/// Default hasher for [`HashMap`](crate::HashMap) and [`HashSet`](crate::HashSet).
+#[cfg(feature = "default-hasher")]
+#[derive(Clone)]
```
Amanieu, line 33: Is it really necessary for this to implement `Clone`? I don't think it every makes sense to clone a `Hasher`.

src/hasher.rs
```diff
@@ -0,0 +1,70 @@
...
+        write_u128(u128),
+        write_usize(usize),
+        write_i8(i8),
+        write_i16(i16),
+        write_i32(i32),
+        write_i64(i64),
+        write_i128(i128),
+        write_isize(isize),
+    }
+
```
Amanieu, line 65: Could you also forward `write_length_prefix` and `write_str` under the `nightly` feature?

cuviper: Sure, but it won't be useful until we also upgrade to `foldhash` v0.2 and forward `foldhash/nightly`.

src/hasher.rs
cuviper: `std` added `Clone` in rust-lang/rust#42799, without much discussion.

Amanieu:
Hmm apparently updating the version number will cause release-plz to immediately publish a new version of the crate.

PR #649, opened 2025-09-14
Depend on `serde_core` instead of `serde`

DaniPopes:
This crate does not make use of serde derive macros, thus it can depend on `serde_core` instead of `serde` to speed up users' compile times.

See the documentation of [`serde_core`](https://docs.rs/serde_core) for more details.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ foldhash = { version = "0.2.0", default-features = false, optional = true }
 
 # For external trait impls
 rayon = { version = "1.2", optional = true }
-serde = { version = "1.0.25", default-features = false, optional = true }
+serde_core = { version = "1.0.221", default-features = false, optional = true }
 
 # When built as part of libstd
 core = { version = "1.0.0", optional = true, package = "rustc-std-workspace-core" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -33,6 +33,11 @@ allocator-api2 = { version = "0.2.9", optional = true, default-features = false,
 # NB: this is a public dependency because `Equivalent` is re-exported!
 equivalent = { version = "1.0", optional = true, default-features = false }
 
+# serde v1.0.220 is the first version that released with `serde_core`.
+# This is required to avoid conflict with other `serde` users which may require an older version.
+[target.'cfg(any())'.dependencies]
+serde = { version = "1.0.220", default-features = false }
```
cuviper, line 39: I think this hack should follow the `optional` state, so it has no effect otherwise.
```suggestion
serde = { version = "1.0.220", default-features = false, optional = true }
```

Cargo.toml
```diff
@@ -60,6 +65,9 @@ rustc-dep-of-std = [
     "rustc-internal-api",
 ]
 
+# Enables serde support.
+serde = ["dep:serde_core"]
```
cuviper, line 69:
```suggestion
serde = ["dep:serde_core", "dep:serde"]
```

cuviper (approved):

PR #650, opened 2025-09-16
Restore MRSV to 1.63

a1phyr:


cuviper:
Opinions may vary, but I don't see how this is less readable. Having `.cast_mut().cast()` reveals very little information.
(IMO `.cast_mut()` is only better when you *aren't* intending to change the type.)

a1phyr:
I think that rust 1.63 was chosen because it was the version of Debian Stable, but now that a new Debian version is out, changing it is probably OK. I usually don't push for freezing MRSV for years, I just needed that for `indexmap` and this was a simple change. As `indexmap` ended up allowing both versions, feel free to close this PR if you don't it desirable;

Amanieu:
The latest debian stable is on Rust 1.85. @cuviper do you think it would be reasonable to bump MSRV all the way to that?

cuviper:
No, e.g. RHEL only has 1.84.1 right now, but it will get an upgrade around November.

PR #654, opened 2025-09-27
style: Add rustfmt configuration

Dietr1ch:
I'm using a different style targeting the 2024 edition and it results in many unintended changes.

TESTED=I ran `cargo fmt --all` and it resulted in no changes.

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,3 @@
+# https://github.com/rust-lang/rustfmt/blob/master/Configurations.md
+
+edition = "2021"


Amanieu:
Shouldn't rustfmt automatically pick up the edition from Cargo.toml?

Dietr1ch:
That makes sense, but it seems that the edition specified on `~/.config/rustfmt/rustfmt.toml` currently trumps over the project's edition, which should be a bug in `rustfmt`. I'll just close this then.

PR #656, opened 2025-10-11
Add integration with Paralight

gendx:
[Paralight](https://github.com/gendx/paralight) is a lightweight parallelism library tuned for indexed structures such as slices. Given that the internal representation of hashbrown's hash tables is a slice of buckets (that each optionally contain a value), it's a good fit to integrate with (https://github.com/gendx/paralight/issues/5).

This pull request is here to iterate on the design. As the integration needs access to the raw hash table representation, it's done here in the hashbrown crate (similarly to Rayon's integration).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ str_to_string = "warn"
 foldhash = { version = "0.2.0", default-features = false, optional = true }
 
 # For external trait impls
-paralight = { version = "0.0.10", optional = true }
+paralight = { version = "0.0.11", optional = true }
 rayon = { version = "1.9.0", optional = true }
 serde_core = { version = "1.0.221", default-features = false, optional = true }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ str_to_string = "warn"
 foldhash = { version = "0.2.0", default-features = false, optional = true }
 
 # For external trait impls
-paralight = { version = "0.0.6", optional = true }
+paralight = { version = "0.0.10", optional = true }
 rayon = { version = "1.9.0", optional = true }
 serde_core = { version = "1.0.221", default-features = false, optional = true }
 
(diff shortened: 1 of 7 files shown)


gendx:
With Paralight v0.0.10, it's now possible to define parallel iterators where each bucket produces an `Option<Self::Item>` without needing `Item` to contain an option. This addresses @Amanieu's comment and makes the API much nicer to use. This change is therefore now ready for review.

Some general remarks:
- Regarding `Send`/`Sync` bounds, I've created dedicated wrapper types around `RawTable`, because the default bounds were not suitable:
  - `into_par_iter()` requires `Send` items but the wrapper needs to be `Sync` to be shared with worker threads.
  - `HashMap::par_iter_mut()` requires a `Sync` key (to produce `&K`) and a `Send` value (to produce `&mut V`) and again the wrapper needs to be `Sync`.
  - In all cases, the `Allocator` shouldn't have any particular `Send` nor `Sync` bounds, as Paralight iterators don't deallocate (nor allocate) the table on other threads. Only the `Drop` implementation will deallocate in the `into_par_iter()` case, and none of the wrappers have new `Send` implementations anyway.
- I've added a new support function `RawTable::deallocate_cleared_table()` to directly deallocate the table without clearing the control bytes when the iterator is dropped (after having fetched all the items*). This should be more efficient than using the pre-existing `clear_no_drop()` (although using that would work too).
- Paralight is still in the 0.0.x version. It's already usable but the API changes often between versions as more iterator adaptors become supported, hence I haven't committed to publishing a 0.x version yet. That said I don't envision further changes to the `IntoParallelSource` traits for the time being. Happy to leave this pull request open and keep updating it until Paralight v0.1 is ready (but feedback on the design is valuable to progress towards version 0.1).

*Internally, the Paralight execution engine will ensure that each bucket index has been passed once and only once to a call to `fetch_item()` + `cleanup_item_range()` calls. While it's possible to bypass the execution engine and directly manipulate and drop a `SourceDescriptor` with safe code (definitely not the intended use of the API), this is at worst a memory leak (not dropping all the items in the map) and therefore not unsound.

PR #657, opened 2025-10-23
Add `HashTable` methods related to the raw bucket index

### LINKED ISSUE #613: Bucket-based API for HashTable
I want to propose the following API for `HashTable`. On `OccupiedEntry` I would like to add

```rust
/// Return in which bucket this entry is located.
///
/// Note: this index is not stable if the hash table is rehashed.
fn get_bucket_idx(&self) -> usize;
```

On `HashTable` itself I would like to add the following three functions:

```rust
/// Returns the bucket index an entry in the hash table is located at, or `None` otherwise.
pub fn find_bucket_idx(&self, hash: u64, eq: impl FnMut(&T) -> bool) -> Option<usize>;

/// Return a reference to the value located at `bucket_idx`, or `None` otherwise.
fn at_bucket(&self, bucket_idx: usize) -> Option<&T>;

/// Return an `OccupiedEntry` for the entry located at `bucket_idx`, or `None` otherwise.
fn at_bucket_mut(&mut self, bucket_idx: usize) -> Option<OccupiedEntry<T>>;

/// Returns the total number of buckets in this `HashTable`.
fn num_buckets(&self) -> usize;
```

There are multiple reasons why I'd like this.

1. It allows incredibly flexible iteration over the hash table. I am aware that this iteration would be suboptimal compared to a SIMD-based iteration which checks the tags of 16 elements in one go, but the flexibility it provides is well worth it. For example: it makes parallel iteration over the hash table trivial.

2. It allows for intrusive data structures to interweave with the hash table, so long as you are careful to avoid rehashing. An example I ran into recently was the writing of a fixed-capacity LRU cache. When the LRU cache wants to evict a key it has to do a lookup for that key. It could instead simply store the bucket index and use that to directly remove the key, as the hash table never rehashes in this example.

cuviper:
On `HashTable<T, A>`:
```rust
pub fn num_buckets(&self) -> usize;
pub fn find_bucket_index(&self, hash: u64, eq: impl FnMut(&T) -> bool) -> Option<usize>;
pub fn get_bucket_entry(&mut self, index: usize) -> Result<OccupiedEntry<'_, T, A>, AbsentEntry<'_, T, A>>;
pub unsafe fn get_bucket_entry_unchecked(&mut self, index: usize) -> OccupiedEntry<'_, T, A>;
pub fn get_bucket(&self, index: usize) -> Option<&T>;
pub unsafe fn get_bucket_unchecked(&self, index: usize) -> &T;
pub fn get_bucket_mut(&mut self, index: usize) -> Option<&mut T>;
pub unsafe fn get_bucket_unchecked_mut(&mut self, index: usize) -> &mut T;
pub fn iter_buckets(&self) -> IterBuckets<'_, T>;
pub fn iter_hash_buckets(&self, hash: u64) -> IterHashBuckets<'_, T>;
```

On `OccupiedEntry<'_, T, A>`:
```rust
pub fn bucket_index(&self) -> usize;
```

`IterHashBuckets<'_, T>` implements `Clone`, `Debug`, `Default`, `Iterator<Item = usize>`, and `FusedIterator`.
`IterBuckets<'_, T>` implements all those and also `ExactSizeIterator`.

Closes #613, although I chose different names.

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -1100,7 +1100,7 @@ where
     /// #     test()
     /// # }
     /// ```
-    pub fn iter_buckets(&self) -> IterBuckets<'_> {
+    pub fn iter_buckets(&self) -> IterBuckets<'_, T> {
         IterBuckets {
             inner: unsafe { self.raw.full_buckets_indices() },
             marker: PhantomData,
@@ -1233,7 +1233,7 @@ where
     /// #     test()
     /// # }
     /// ```
-    pub fn iter_hash_buckets(&self, hash: u64) -> IterHashBuckets<'_> {
+    pub fn iter_hash_buckets(&self, hash: u64) -> IterHashBuckets<'_, T> {
         IterHashBuckets {
             inner: unsafe { self.raw.iter_hash_buckets(hash) },
             marker: PhantomData,
@@ -2569,33 +2569,55 @@ where
 ///
 /// This `struct` is created by the [`HashTable::iter_buckets`] method. See its
 /// documentation for more.
-#[derive(Clone, Default)]
-pub struct IterBuckets<'a> {
+pub struct IterBuckets<'a, T> {
     inner: FullBucketsIndices,
-    marker: PhantomData<&'a ()>,
+    marker: PhantomData<&'a T>,
 }
 
-impl Iterator for IterBuckets<'_> {
+impl<T> Clone for IterBuckets<'_, T> {
+    #[inline]
+    fn clone(&self) -> Self {
+        Self {
+            inner: self.inner.clone(),
+            marker: PhantomData,
+        }
+    }
+}
+
+impl<T> Default for IterBuckets<'_, T> {
+    #[inline]
+    fn default() -> Self {
+        Self {
+            inner: Default::default(),
+            marker: PhantomData,
+        }
+    }
+}
+
+impl<T> Iterator for IterBuckets<'_, T> {
     type Item = usize;
 
+    #[inline]
     fn next(&mut self) -> Option<usize> {
         self.inner.next()
     }
 
+    #[inline]
     fn size_hint(&self) -> (usize, Option<usize>) {
         self.inner.size_hint()
     }
 }
 
-impl ExactSizeIterator for IterBuckets<'_> {
+impl<T> ExactSizeIterator for IterBuckets<'_, T> {
+    #[inline]
     fn len(&self) -> usize {
         self.inner.len()
     }
 }
 
-impl FusedIterator for IterBuckets<'_> {}
+impl<T> FusedIterator for IterBuckets<'_, T> {}
 
-impl fmt::Debug for IterBuckets<'_> {
+impl<T> fmt::Debug for IterBuckets<'_, T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_list().entries(self.clone()).finish()
     }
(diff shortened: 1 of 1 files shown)


src/table.rs
```diff
@@ -591,6 +755,44 @@ where
...
+    /// assert_eq!(count, 3);
+    ///
+    /// assert_eq!(table.get_bucket(table.buckets()), None);
+    /// # }
+    /// # fn main() {
+    /// #     #[cfg(feature = "nightly")]
+    /// #     test()
+    /// # }
+    /// ```
+    pub fn buckets(&self) -> usize {
```
clarfonthey: I personally would name this `num_buckets` rather than `buckets` since the name as-is makes me think that it's returning an iterator instead.

clarfonthey (changes requested):
I think that `get_bucket` and `get_bucket_mut` are good, but `buckets` should be renamed to `num_buckets` to make it clear it's returning a count and not an iterator.

In terms of `find_bucket_index` and `get_bucket_entry`, I left some comments. But if you wanted to merge the smaller set first, I'm fine with that.

clarfonthey:
re: unsafe methods, I guess the main concern is whether the bounds checks can easily be optimised out. While there isn't the same robust codegen testing as the compiler here, maybe it would be worth poking around on godbolt and seeing how `find_bucket_index` followed by one of the relevant methods is handled.

cuviper:
I assume you mean adding that in `find_bucket_index`?

```rust
    pub fn find_bucket_index(&self, hash: u64, eq: impl FnMut(&T) -> bool) -> Option<usize> {
        match self.raw.find(hash, eq) {
            Some(bucket) => Some(unsafe {
                let index = self.raw.bucket_index(&bucket);
                core::hint::assert_unchecked(index < self.raw.buckets());
                core::hint::assert_unchecked(self.raw.is_bucket_full(index));
                index
            }),
            None => None,
        }
    }
```

It does remove the unwrap in my simple back-to-back test, but it also pessimizes the unchecked case a little, recomputing the final address. Different variations of that assertion had the same effect.

```asm
+        neg r9
+        lea rax, [rax + 8*r9]
         add rax, -8
         pop rcx
         .cfi_def_cfa_offset 8
         ret
```

In a real scenario I expect you'd also have other interim code, and who knows if the assertion could still be propagated. I don't think we should bother until there's a real measured benefit, and anyway even the checked case is still much less than a full hash probe that you would have needed before!

cuviper:
RE MSRV, I recently bumped `indexmap` to 1.82, which is just over a year old.

src/raw/mod.rs
```diff
@@ -11,6 +11,8 @@ use core::ptr::NonNull;
 use core::slice;
 use core::{hint, ptr};
 
+pub(crate) use crate::control::Tag;
```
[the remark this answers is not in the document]
cuviper: Because `table::VacantEntry` now has a field of this type.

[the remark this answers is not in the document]
clarfonthey: Yeah, but why re-export instead of just importing in that file too? I prefer to avoid having too many aliases because it makes the code difficult to follow.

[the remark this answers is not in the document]
cuviper: Oh, right. I was thinking it came from a submodule here, and I clearly didn't pay attention to the path. I've now made `table` use that directly, thanks!

clarfonthey (approved):

cuviper:
> Is there any downside to having `get_bucket_mut` always return an `OccupiedEntry`? It's strictly more flexible than returning a `&mut T` and this API is only intended for advanced use cases anyways. Same with the unchecked version.

i.e. instead of the separate `get_bucket_entry`, I suppose? At first I needed the hash value for that, which was a downside if you only wanted `&mut T`, but now I guess they are functionally equivalent.

I suppose the downside now is just that the caller would have to use an extra `.get_mut()` or `.into_mut()` on the entry if that's all they want, and hope that the compiler does optimize the unused stuff away. It probably will though.

Hmm, if `OccupiedEntry` implemented `Deref<Target = T>` and `DerefMut`, then my experimental branch in `indexmap` wouldn't even need to change the way it's currently using `get_bucket_mut()`. That could be nice, but it's not required.

cuviper:
Note that there is a similar trio in `find`, `find_mut`, and `find_entry` as well.

cuviper:
I went ahead and made that change to try it -- easy enough to drop that commit if we change our minds. :)

cuviper:
* Reverted `get_bucket_mut` (and unchecked) back to `&mut T`
* Changed `get_bucket_entry` to return `Result<OccupiedEntry, AbsentEntry>`
* Added `get_bucket_entry_unchecked` returning `OccupiedEntry`
* Added `iter_buckets` and `iter_hash_buckets`

> The only missing functionality here is the ability to insert a value just by bucket index, but it's not clear how useful this would be since the only way to get a valid index for this is by re-using the index of a previous insertion of the same key.

I don't think we should support this, as it would let you insert values in places that wouldn't be probed by the same hash, especially if we're not making any guarantees about index stability after removals. You *can* do something like:

```rust
    let (old, vacant) = table.get_bucket_entry(index).unwrap().remove();
    let new_index = vacant.insert(new).bucket_index();
```

... but I think even there we're not _promising_ that the new index will be the same. Supposing we were to shift items to avoid tombstones, then the vacant entry might already represent a different bucket.

src/table.rs
```diff
@@ -741,6 +1071,42 @@ where
         }
     }
 
+    /// An iterator producing the `usize` indices of all occupied buckets.
+    ///
+    /// The order in which the iterator yields indices is unspecified
+    /// and may change in the future.
```
cuviper: I think it would actually be fine to guarantee increasing order here, but the underlying `raw::FullBucketsIndices` is explicitly disclaiming that.

cuviper:
Rebased to resolve conflicts with the `InsertSlot` removal.

src/table.rs
```diff
@@ -2230,6 +2727,32 @@ where
...
+#[derive(Clone, Default)]
+pub struct IterHashBuckets<'a> {
+    inner: RawIterHashIndices,
+    marker: PhantomData<&'a ()>,
+}
+
+impl Iterator for IterHashBuckets<'_> {
+    type Item = usize;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
Amanieu: This method is not generic and need an `#[inline]`.

src/table.rs
Amanieu: I think it's fine to leave it unspecified for now unless there's a use case that clearly asks for it.

PR #667, opened 2025-11-29
Add `hash_table::UnsafeIter`, `iter()` method to various iterators

clarfonthey:
Both of these would be useful for #666, and presumably other downstream hash table implementations.

First, the `iter()` methods make it possible to implement downstream `Debug` implementations more easily, by letting iterators like `Drain` be paused and viewed immutably.

Second, the new `unsafe_iter` method and corresponding `UnsafeIter` exists to make implementing `hash_map::IterMut` a bit more reasonable without being *as* unsafe. The struct docs should hopefully make its existence clear, but I would definitely love some extra scrutiny on that.

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -1,4 +1,4 @@
-use core::{fmt, iter::FusedIterator, marker::PhantomData};
+use core::{fmt, iter::FusedIterator, marker::PhantomData, ptr::NonNull};
 
 use crate::{
     control::Tag,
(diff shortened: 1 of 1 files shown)


src/table.rs
```diff
@@ -2552,15 +2580,112 @@ where
...
+/// impl<'a, K, V> Iterator for IterMut<'a, K, V> {
+///     // Immutable keys, mutable values
+///     type Item = (&'a K, &'a mut V);
+///
+///     fn next(&mut self) -> Option<Self::Item> {
+///         // Even though the key is mutably borrowed here, this is sound
+///         // because we never actually mutate the key before yielding the
+///         // immutable reference
+///         let (ref key, ref mut val) = self.inner.next()?;
+///         Some((key, val))
```
clarfonthey: Since this is just an example, and not part of the actual implementation, I chose to use a clearer implementation instead of the one we actually use.

Amanieu:
Rather than adding a whole new iterator kind, would it make more sense to direct users that need custom variance to the bucket iterator instead?

PR #670, opened 2025-12-03
Add `hash_map::{OccupiedEntry::into_entry, VacantEntryRef::insert_entry_with_key}`, make `EntryRef` use `ToOwned` again

### LINKED ISSUE #660: QQ: Why use `&'b Q: Into<K>` over `Q: ToOwned<Owned = K>` as `EntryRef::insert` trait bound?
Hi. Thank you for paying attention to my (possibly silly) questions.

I'm using `HashMap::entry_ref()` and `EntryRef::insert()` APIs to try to reduce some clone overhead in my project. I found that the trait bound for `EntryRef::insert()` is `&'b Q: Into<K>`, rather than mostly used `Q: ToOwned<Owned = K>`. The `&'b Q: Into<K>` is a bit inconvenient to expose to the user since the primitives such as `u64` hasn't implemented `&u64: Into<u64>`.

May I ask why this trait bound is used here? Thank you very much. 🙏

clarfonthey:
This allows implementing `get_or_insert` and `get_or_insert_with` without accessing the internals of `HashMap`, which is useful for #666.

----

Since the merge queue is busted, I decided to incorporate the [suggestion](https://github.com/rust-lang/hashbrown/issues/660#issuecomment-3648505994) to make `EntryRef` use `ToOwned` again.

Resolves #660.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -2,6 +2,7 @@ use crate::raw::{
     Allocator, Bucket, Global, RawDrain, RawExtractIf, RawIntoIter, RawIter, RawTable,
 };
 use crate::{DefaultHashBuilder, Equivalent, TryReserveError};
+use ::alloc::borrow::ToOwned;
 use core::borrow::Borrow;
 use core::fmt::{self, Debug};
 use core::hash::{BuildHasher, Hash};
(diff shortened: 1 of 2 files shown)


clarfonthey:
Things should now compile, and remaining references in docs to `From` have been replaced with `ToOwned` instead.

PR #672, opened 2025-12-11
Update allocator-api2

zakarumych:
New version fixes some inconsistencies with std

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ core = { version = "1.0.0", optional = true, package = "rustc-std-workspace-core
 alloc = { version = "1.0.0", optional = true, package = "rustc-std-workspace-alloc" }
 
 # Support for allocators that use allocator-api2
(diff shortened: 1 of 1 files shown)


cuviper:
> But it is a breaking change for users who use allocator-api2 for hashbrown types, as they would need to upgrade their version.

That's really what I meant, as it's (implicitly) a [public dependency](https://rust-lang.github.io/rfcs/1977-public-private-dependencies.html). And this is fine for `hashbrown` to change IMO, as long as we do bump semver for it.

I admit I didn't look at what actually went wrong with bumpalo in particular.

clarfonthey:
The PR for that is fitzgen/bumpalo#299, FWIW. Just so others don't have to search for it.

scutuatua-crypto (approved):

clarfonthey:
Also something to consider here: rust-lang/rust#156882

It seems highly possible that updating to stable allocator on a future MSRV will happen before this gets merged.

PR #676, opened 2025-12-23
MSRV 1.85, Edition 2024

clarfonthey:
Includes clippy fixes for methods that were added in the meantime that could be used instead.

I also took the liberty of running `taplo fmt` in the repo to clean up the TOML files. This is now done in CI.

I've done my best to try and minimise the unsafe block bounds auto-added by fixes as well as possible, but there's a lot of code that needs to be annotated with safety comments anyway, so, some are left with very large unsafe blocks.

Additionally, I have made `unreachable_pub` now warn by default as agreed upon in #675. With a higher MSRV, we can avoid false positives.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -56,7 +56,7 @@ jobs:
           # Disabled due to issues with cross-rs
           #x86_64-pc-windows-gnu,
         ]
-        channel: [1.84.0, nightly]
+        channel: [1.85.1, nightly]
         include:
         - os: macos-latest
           target: aarch64-apple-darwin
@@ -66,10 +66,10 @@ jobs:
           channel: nightly
         - os: macos-latest
           target: aarch64-apple-darwin
-          channel: 1.84.0
+          channel: 1.85.1
         - os: windows-latest
           target: x86_64-pc-windows-msvc
-          channel: 1.84.0
+          channel: 1.85.1
         - os: ubuntu-latest
           target: x86_64-unknown-linux-gnu
           channel: beta
--- a/src/map.rs
+++ b/src/map.rs
@@ -4934,7 +4934,7 @@ mod test_map {
     use core::alloc::Layout;
     use core::ptr::NonNull;
     use core::sync::atomic::{AtomicI8, Ordering};
-    use rand::{rngs::SmallRng, Rng, SeedableRng};
+    use rand::{Rng, SeedableRng, rngs::SmallRng};
     use std::borrow::ToOwned;
     use std::cell::RefCell;
     use std::vec::Vec;
@@ -5800,7 +5800,7 @@ mod test_map {
     #[test]
     fn test_entry_take_doesnt_corrupt() {
         #![expect(deprecated)] //rand
-                               // Test for #19292
+        // Test for #19292
         fn check(m: &HashMap<i32, ()>) {
             for k in m.keys() {
                 assert!(m.contains_key(k), "{k} is in keys() but not in the map?");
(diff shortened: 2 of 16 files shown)


src/lib.rs
```diff
@@ -25,24 +25,9 @@
...
-    clippy::missing_safety_doc,
-    clippy::missing_errors_doc
-)]
-#![warn(missing_docs)]
-#![warn(rust_2018_idioms)]
 #![cfg_attr(feature = "nightly", warn(fuzzy_provenance_casts))]
-#![cfg_attr(
-    feature = "nightly",
-    allow(clippy::incompatible_msrv, internal_features)
-)]
```
clarfonthey: We probably could have removed `incompatible_msrv` here for a while; this now understands that a `nightly` feature means that you're not going to be following MSRV.

.github/workflows/rust.yml
```diff
@@ -54,7 +54,7 @@ jobs:
           # Disabled due to issues with cross-rs
           #x86_64-pc-windows-gnu,
         ]
-        channel: [1.80.0, nightly]
+        channel: [1.84.0, nightly]
```
clarfonthey: Is there a reason why this was hard-coded to 1.80.0? Since we're going to have a higher MSRV, is there a reason why we shouldn't make this just use the MSRV determined from `Cargo.toml` instead?

Amanieu: I believe some of our dev-dependencies had a higher MSRV than us, so we could only test on 1.80. MSRV is checked separately with cargo-msrv.

clarfonthey:
I think at this point, I've implemented all major changes up to the MSRV that I think are important before merging. If I've missed any, please feel free to mention it, although as usual, no rush. I'm not going to have a proper computer to dev this for the next couple days, so, I'll get to any requests when I'm back home unless I happen to notice them before I go to bed tonight.

src/map.rs
```diff
@@ -4732,7 +4724,7 @@ where
         let reserve = if self.is_empty() {
             iter.size_hint().0
         } else {
-            (iter.size_hint().0 + 1) / 2
+            iter.size_hint().0.div_ceil(2)
```
Amanieu: I'm generally not a fan of `div_ceil` because it generates [worse code](https://github.com/rust-lang/rust-clippy/issues/14944) in cases where you know the dividend is not `uX::MAX`.

Amanieu:
@cuviper Is RHEL still stuck on 1.84? Ideally we would like to directly upgrade to the 2024 edition with a MSRV of 1.85.

src/map.rs
clarfonthey: I'm fine manually allowing the lint in the repo. Although, is there a separate issue somewhere filed for the codegen issues, not just the clippy lint? Since I feel like, at least for powers of two, this should be trivial to make as good as the optimal addition + right shift.

clarfonthey:
As mentioned, with the unsafe fn lint fixed here, 1.85 is just a difference of `cargo fmt`, so, it's a quick fix. For the sake of avoiding more rebasing, though, I'll at least do it as a separate commit instead of modifying all of them.

ci/tools.sh
```diff
@@ -24,13 +24,16 @@ retry() {
...
 if retry rustup component add rustfmt ; then
     cargo fmt --all -- --check
 fi
 
 if retry rustup component add clippy ; then
-    cargo clippy --all --tests --features serde,rayon -- -D clippy::all
+    cargo clippy --all --tests --features serde,rayon -- -D warnings
+fi
+
+if command -v taplo ; then
```
clarfonthey: Decided to make this an "optional" check in this script so that people don't need to necessarily have `taplo` installed to have the script pass, although it will still be enforced on CI, where it will be installed.

clarfonthey:
That should be everything!

PR #701, opened 2026-03-15
Improve generated code for NEON

Dandandan:
Disclaimer: found by Claude

This forces LLVM to change:

```
  cmeq.8b  v2, v2, v0       ; NEON compare with EMPTY
  umaxv.8b b2, v2            ; horizontal vector max (~3-4 cycle latency)
  fmov     w17, s2           ; extract to 32-bit GPR
  tbnz     w17, #0, ...      ; test bit
```

into 

```
  fmov  x17, d1                    ; extract group to 64-bit GPR
  and   x17, x17, x17, lsl #1     ; SWAR: both top bits set = EMPTY
  tst   x17, #0x8080808080808080   ; any empty byte?
```

Effect on various benchmarks 5-15%

### DIFF
--- a/src/control/group/neon.rs
+++ b/src/control/group/neon.rs
@@ -76,7 +76,10 @@ impl Group {
     /// `EMPTY`.
     #[inline]
     pub(crate) fn match_empty(self) -> BitMask {
-        self.match_tag(Tag::EMPTY)
+        unsafe {
+            let ctrl = neon::vget_lane_u64(neon::vreinterpret_u64_u8(self.0), 0);
+            BitMask(ctrl & (ctrl << 1) & BITMASK_ITER_MASK)
+        }
     }
 
     /// Returns a `BitMask` indicating all tags in the group which are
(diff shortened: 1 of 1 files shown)


Amanieu:
The description here doesn't make sense. There shouldn't be a `umaxv` instruction in the generated code.

With that said the code is actually an improvement, but I would prefer if this was done directly by modifying `match_empty` instead of adding a new function. If you convert the `uint8x8_t` directly back to an integer then LLVM can eliminate the vector load entirely. This will also properly handle the case of big-endian AArch64, which your current PR doesn't handle correctly (the bitmask will be reversed).

Amanieu:
The UMAXV is actually an LLVM bug, I've opened https://github.com/llvm/llvm-project/issues/186710.

If that is fixed, then it's less clear whether it is always beneficial to use integer operations:

```
  cmeq.8b  v2, v2, v0       ; NEON compare with EMPTY
  fmov     x17, d2          ; extract to 64-bit GPR
  cbnz     x17, ...         ; test result
```

vs

```
  fmov  x17, d1                    ; extract group to 64-bit GPR
  and   x17, x17, x17, lsl #1      ; SWAR: both top bits set = EMPTY
  tst   x17, #0x8080808080808080   ; any empty byte?
  b.ne ...
```

I think that it will essentially boil down to whether the control bytes are already in a vector register because they are needed for a `match_tag`: if yes then it's faster to use `cmeq.8b`, otherwise it's faster to do everything using only integer operations.

Dandandan:
Closing as it was solved in LLVM

PR #712, opened 2026-04-11
lsx: use immediate form of vslti for zero comparison

heiher:
Replace the sequence that materializes a zero vector and performs `lsx_vslt_b` with the immediate variant `lsx_vslti_b::<0>`.


clarfonthey:
Hmm, interesting this is already being used elsewhere in the file but this got missed somehow.

PR #719, opened 2026-04-17
original key in RustcOccupiedEntry

malezjaa:
This is required for https://github.com/rust-lang/rust/pull/155360
Adds an `original_key` field to `RustcOccupiedEntry`

### DIFF
--- a/src/rustc_entry.rs
+++ b/src/rustc_entry.rs
@@ -35,10 +35,13 @@ where
     pub fn rustc_entry(&mut self, key: K) -> RustcEntry<'_, K, V, A> {
         let hash = make_hash(&self.hash_builder, &key);
         if let Some(elem) = self.table.find(hash, |q| q.0.eq(&key)) {
-            RustcEntry::Occupied(RustcOccupiedEntry {
-                elem,
-                table: &mut self.table,
-            })
+            RustcEntry::Occupied(
+                RustcOccupiedEntry {
+                    elem,
+                    table: &mut self.table,
+                },
+                key,
+            )
         } else {
             // Ideally we would put this in VacantEntry::insert, but Entry is not
             // generic over the BuildHasher and adding a generic parameter would be
@@ -64,7 +67,7 @@ where
     A: Allocator,
 {
     /// An occupied entry.
-    Occupied(RustcOccupiedEntry<'a, K, V, A>),
+    Occupied(RustcOccupiedEntry<'a, K, V, A>, K),
 
     /// A vacant entry.
     Vacant(RustcVacantEntry<'a, K, V, A>),
@@ -74,7 +77,7 @@ impl<K: Debug, V: Debug, A: Allocator> Debug for RustcEntry<'_, K, V, A> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match *self {
             Vacant(ref v) => f.debug_tuple("Entry").field(v).finish(),
-            Occupied(ref o) => f.debug_tuple("Entry").field(o).finish(),
+            Occupied(ref o, _) => f.debug_tuple("Entry").field(o).finish(),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/rustc_entry.rs
```diff
@@ -304,6 +307,52 @@ impl<'a, K, V, A: Allocator> RustcOccupiedEntry<'a, K, V, A> {
...
+    /// if let RustcEntry::Occupied(mut o) = map.rustc_entry("poneyland".to_string()) {
+    ///     assert_eq!(o.take_original_key(), "poneyland".to_string());
+    /// }
+    /// ```
+    #[cfg_attr(feature = "inline-more", inline)]
+    pub fn take_original_key(&mut self) -> K {
+        self.original_key
+            .take()
+            .expect("the original key should exist")
+    }
```
tgross35: I think you can drop this, any sort of "nice" abstraction can be on the std.

IMO if anything, this should return `Option<K>` anyway rather than having the panic here.

cuviper:
Note that this *used* to hold the lookup key, but that was removed in #535.

See also https://github.com/rust-lang/rust/issues/44286

cuviper:
I gave some alternate suggestions on https://github.com/rust-lang/rust/pull/155360

Amanieu:
Could you instead change `RustcEntry::Occupied` to contain 2 values (the key and the `OccupiedEntry`)? This is much cleaner and avoids issues with other users of `OccupiedEntry` in the API.

src/rustc_entry.rs
cuviper, line 65: Tangentially -- I don't think we need any of these `RustcEntry` methods, because all `std` does is immediately deconstruct the variants into its own enum. Maybe we could even remove `enum RustcEntry` altogether and use a `Result`, like `fn rustc_entry(..) -> Result<(Occupied, key), Vacant>`.

clarfonthey:
Just so we're on the same page, I was under the impression that `RustcEntry` and co. were different from the existing entries because the behaviour on `std` is different from the behaviour here?

One thing that had come to mind was just having multiple entry types based upon what functionality is desired, and having both be public API for this crate so we don't need to make this an internal module.

Not that this move should happen here, just, a potential justification for not just making methods return a key + entry and instead making it part of the entry, if it's going to be needed longer term for this.

Amanieu:
The main reason `RustcEntry` exists is that the `Entry` type in the standard library (which dates back to Rust 1.0) is not generic over the hasher, which limits what can be done with it. Notably it means that that in std we have to always call `reserve(1)` before creating the entry since it is unable to resize the hash table by itself.

Amanieu:
> Could you instead change `RustcEntry::Occupied` to contain 2 values (the key and the `OccupiedEntry`)? This is much cleaner and avoids issues with other users of `OccupiedEntry` in the API.

I'm actually having second thoughts about this since it would mean yet another major version bump for `hashbrown`. @clarfonthey @cuviper what do you think?

tgross35:
Isn't `RustcEntry` and similar API semver-exempt by being gated behind `rustc-internal-api`?

cuviper:
Does `cargo -Zbuild-std` always respect the lock file? (edit: [seems not](https://github.com/rust-lang/hashbrown/pull/719#issuecomment-4277646192)) If there's any doubt about that, then I don't think we can just wave away semver issues without having std pinned to an exact version.

An alternative is to add a *new* method, so we avoid perturbing the performance of the normal `entry` path at all.

```
pub fn rustc_try_insert(&mut self, key: K, value: V) -> Result<&mut V, (RustcOccupiedEntry<'_, K, V, A>, K, V)>
```

Then the parts in the `Err` can be mapped to std's `OccupiedError`. We could wrap that in a `RustcOccupiedError` here if you really want.

cuviper:
See #722 for `rustc_try_insert` -- I think this is a cleaner approach.

PR #722, opened 2026-04-20
Add `HashMap::rustc_try_insert`

cuviper:
This works like `try_insert` but with `RustcOccupiedEntry` in the error
case. The error also contains the attempted key as requested in the
upstream [`try_insert` tracking issue](https://github.com/rust-lang/rust/issues/82766#issuecomment-4253028464).

### DIFF
--- a/src/rustc_entry.rs
+++ b/src/rustc_entry.rs
@@ -1,6 +1,6 @@
 use self::RustcEntry::*;
 use crate::alloc::{Allocator, Global};
-use crate::map::{Drain, HashMap, IntoIter, Iter, IterMut, make_hash};
+use crate::map::{Drain, HashMap, IntoIter, Iter, IterMut, make_hash, make_hasher};
 use crate::raw::{Bucket, RawTable};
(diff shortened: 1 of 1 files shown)


src/rustc_entry.rs
```diff
@@ -52,7 +52,72 @@ where
...
+{
+    /// The entry in the map that was already occupied.
+    pub entry: RustcOccupiedEntry<'a, K, V, A>,
+    /// The key which was not inserted, because the entry was already occupied.
+    pub key: K,
+    /// The value which was not inserted, because the entry was already occupied.
+    pub value: V,
 }
+// NB: there are no impls for this type, because we expect std will
+// immediately reconstruct these errors into its own `OccupiedError`.
```
clarfonthey: I would just make this part of the documentation for the type, honestly. Putting it at the end makes it a bit easier to gloss over.

clarfonthey:
Looks okay to me with one minor comment. Would be cool with you merging after that's fixed.

PR #725, opened 2026-04-29
Support finding insertion positions from &HashTable

### LINKED ISSUE #723: Extend HashTable's bucket API to support insertion ?
When one wishes to "get or insert" from/into a `HashTable` that is behind a `RwLock`, one typically first attempts to "get" with only a read lock (via the `.find` API), and only if that fails fallback to a "get or insert" with a write lock (via the `.entry` API).

However, the `.entry` call repeats the probing that was undertaken during the `.find` call: this is necessary because the table could have been modified between release of the read lock and acquisition of the write lock.

But in some cases, one might be able to determine whether the table was modified in the interim and thus avoid such duplicate probing.  For example, if the table is append-only and its length has not changed.

The new bucket API almost exposes such an ability, but not quite.  Is there any interest it being extended to support this use case, or is this perhaps too niche for hashbrown?

eggyal:
Adds to `HashTable` the following APIs:

```rust
fn find_or_find_vacant_position(&self, hash, eq)
  -> Result<&T, Option<VacantPosition>>

unsafe fn insert_at_position(&mut self, position, value) -> &mut T
```

This enables an insertion position to be located through an immutable `&HashTable` reference (such as through a read-lock of a `RwLock`), in order that it might subsequently be used to perform an insertion when an exclusive `&mut HashTable` reference becomes available (such as through a write-lock of the same `RwLock`).  In so doing, users can avoid repeated hash probing when it is known that the table has not been mutated in the interim.

Closes #723 cc @cuviper

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -94,6 +94,17 @@ impl<T> HashTable<T, Global> {
     }
 }
 
+/// Represents the position in a particular [`HashTable`] at which an entry with
+/// some particular hash may be inserted.
+///
+/// Created by [`HashTable::find_or_find_vacant_position`] if the specified element
+/// does not currently exist in the table and there is capacity for its insertion
+/// without reallocating.
+pub struct VacantPosition {
+    hash: u64,
+    index: usize,
+}
+
 impl<T, A> HashTable<T, A>
 where
     A: Allocator,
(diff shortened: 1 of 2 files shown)


src/table.rs
```diff
@@ -229,6 +240,51 @@ where
...
+
+    /// Inserts `value` into `self` at `position`.
+    ///
+    /// # Safety
+    /// Behavior is undefined unless `self`:
+    ///
+    /// 1. is the same [`HashTable`] instance as was used to create `position`;
+    /// 2. has not been mutated (in any way) in the interim.
+    pub unsafe fn insert_at_position(&mut self, position: VacantPosition, value: T) -> &mut T {
+        debug_assert!(self.len() < self.capacity());
```
cuviper: I don't think this is the sanity check we need, rather that `index < num_buckets && !is_bucket_full(index)`.

We could also split function this into a safe version, perhaps `-> Result<&mut T, T>` giving the value back on error, and an unsafe unchecked version without any assertion. That safe version could have logic errors if the table was mutated in some way while keeping that index apparently valid -- but it would still be memory safe.

cuviper: Maybe the safe version also needs some check to make sure we don't exceed the load factor, but IIRC it's possible to do that even with no apparent capacity left if the index happens to be a `DELETED` tombstone.

eggyal: > That safe version could have logic errors if the table was mutated in some way while keeping that index apparently valid -- but it would still be memory safe.

Is this correct?  What would such a safe function call to perform the insertion?  Currently both `RawTable::insert_at_index` and `RawTable::insert_tagged_at_index` stipulate that it's unsound to call them with an `index` that wasn't previously returned by `find_or_find_insert_index` or if any mutation of the table has occurred since.

Looking at their implementations, they ultimately make four `unsafe` calls: I *think* that three of those (`RawTableInner::ctrl`, `RawTable::bucket` and `Bucket::write`) might actually only be unsound if the index is out of range; the fourth `unsafe` call, to `RawTableInner::record_item_insert_at`, doesn't have any safety documentation—and I'm not sufficiently familiar with how the metadata is maintained to be certain there's no safety concern here.  For example, it increments `self.items`; if there's an existing value in the bucket that's being overwritten, then surely that will result in an incorrect count?

Perhaps there's no soundness issues, and it's okay to relax the safety requirements of `RawTable::insert_at_index` and `RawTable::insert_tagged_at_index` to merely require that the index be in range (stipulating instead that passing incorrect indices may result in logic errors and/or panics)?  They are only crate-visible, after all.  Then it would indeed be pretty trivial to publicly expose a safe/checked version of this new `HashTable::insert_at_position` method.

cuviper: > For example, it increments `self.items`; if there's an existing value in the bucket that's being overwritten, then surely that will result in an incorrect count?

Checking `is_bucket_full` prevents that case.

cuviper: Anyway, I'd be interested if others have opinions before you change anything toward my suggestion.

clarfonthey:
I'm not necessarily against this API, but I did recently factor out the `InsertIndex` API and do wonder if we really want to be recreating a similar one. Since not mutating the table between reading is a pretty big invariant to uphold, I do wonder if it's worth wrapping the hash + index in an opaque struct instead of just passing the index directly and letting the user pass in the hash again.

PR #727, opened 2026-05-13
Add a way to prefetch a hash table bucket (#677)

### LINKED ISSUE #677: Allow to prefetch buckets
In [Apache DataFusion](https://github.com/apache/datafusion) we are using hashbrown extensively and I noticed in hash aggregation when the hash table (we use the raw API there) become very large we are memory bound so a natural solution would be to have some kind of prefetching (FYI we are using stable Rust), but there is no way to do that currently

We have something like this (simplified):

```rust
for (row, &target_hash) in batch_hashes.iter().enumerate() {
    let entry = self.map.entry(
        target_hash,
        // eq
        |(exist_hash, group_idx)| target_hash == *exist_hash && group_rows.row(row) == group_values.row(*group_idx),
        // hasher
        |(hash, _)|  *hash);

    let group_idx = match entry {
        // Existing group_index for this group value
        Entry::Occupied(o) => {
            let (_hash, group_idx) = o.get();  
            
            *group_idx
        }
        // Need to create new entry for the group
        Entry::Vacant(v) => {
            // Add new entry to aggr_state and save newly created index
            let group_idx = group_values.num_rows();
            group_values.push(group_rows.row(row));
            v.insert((target_hash, group_idx));

            group_idx
        }
    };
    groups.push(group_idx);
}
```

Allowing to prefetch from the map can help performance for our case.

Because we insert to the table on missing we would have to reserve the x amount of items beforehand for the prefetching to be valuable

RyanJamesStewart:
## Summary

Closes #677 (or starts the concrete discussion of it — see the API question below).

Adds `prefetch(hash)` to `RawTable`, exposed as `HashTable::prefetch(hash)`, `HashMap::prefetch(&Q)`, and `HashSet::prefetch(&Q)` (the map/set versions hash the key first). It issues a software prefetch hint for the two cache lines a lookup of that hash would touch first — the control-byte group at the start of the probe sequence, and the corresponding data bucket — like abseil's `prefetch_hash`. It's a hint only: no memory access, never faults (an invalid/dangling address is harmless), and a no-op in the abstract machine.

Stable path, per architecture: `_mm_prefetch::<_MM_HINT_T0>` on x86 / x86-64 (`prefetcht0`), a no-op everywhere else — aarch64 has no stable prefetch intrinsic yet, and `core::intrinsics::prefetch_read_data` is unstable (rust-lang/rust#146941). The new `src/prefetch.rs` shim is `#[cfg(not(miri))]`-gated for the intrinsic, like the SIMD `Group` impls. The pointer arithmetic uses `wrapping_*` deliberately (a prefetch of an out-of-bounds address is a safe no-op, so it must be the arithmetic that doesn't UB) — works on the empty singleton too.

## API question

For now this is **L1 read prefetch only** — the minimal thing, so there's something concrete to react to. If you'd rather offer the full `Locality` / read-vs-write surface, or wait for the std prefetch hints (rust-lang/rust#146941) to stabilize so this can re-use `Locality`, I'm happy to redo it — the `Locality` form is an easy extension on top of this. (Also, per the thread: this prefetches *both* the ctrl and data lines, and it's a `prefetch(hash)` method rather than a "raw bucket guess" getter.)

## When it helps

Only when looking up many hashes in a sequence *and* the table is large enough that its control bytes have spilled out of cache — then the caller prefetches a key several lookups ahead. On a single lookup, or a cache-resident table, it does nothing useful (it's a slight loss — extra instructions, no missed-load to hide). The new `benches/prefetch.rs` `batch_lookup` bench (16-byte keys, randomized lookup order, prefetch `i+8`) shows the crossover:

```
   4K slots:  ~0.7×   (cache-resident → pure overhead, a loss — the docs say so)
  64K slots:  ~0.93×
 256K slots:  ~1.15×
   1M slots:  ~1.15×
   4M slots:  ~1.12×   (ctrl array spilled L2/L3)
```

Matches the cross-architecture numbers @joshuaisaact posted in #677 (~8% @1M, ~15% @4M); batched multi-key lookups, per the thread, do better. The doc comments on the new methods spell this out so nobody reaches for it on a hot small table.

## Tests

- `cargo test` green, including a new `test_prefetch` over the empty singleton / tiny / 1000-entry / ZST-value tables, present and absent keys, and the look-ahead pattern (asserts the table is intact and lookups still work after prefetching).
- `cargo +nightly miri test --lib` green — no UB (the pointer math runs under Miri; the intrinsic is `not(miri)`-gated, so the no-op branch runs there).
- `cargo clippy --all-targets` 0 warnings; `cargo fmt -- --check` clean.

+374 lines, 8 files (the `src/prefetch.rs` shim, the `RawTable`/`HashTable`/`HashMap`/`HashSet` methods, the bench, the test).

### DIFF
--- a/benches/prefetch.rs
+++ b/benches/prefetch.rs
@@ -1,23 +1,41 @@
-//! Batch-lookup benchmark: look up a list of keys in a large `HashMap`, with
-//! and without software-prefetching a key a few iterations ahead.
+//! Benches for `HashMap::prefetch_get` and `HashMap::prefetch_insert`.
 //!
-//! Prefetching only pays off when the table is large enough that its control
-//! bytes spill out of the L2/L3 cache *and* the caller can issue the prefetch
-//! far enough ahead of the use. So this benchmark sweeps the table size and
-//! uses a randomized lookup order (so the access pattern is cache-hostile).
-//! On a small, cache-resident table the prefetch is noise (or a slight loss);
-//! the win shows up on the large sizes.
+//! Two flavors of workload: batch lookups (`prefetch_get`) and batch inserts
+//! (`prefetch_insert`). Each flavor runs against integer keys (`(u64, u64)`)
+//! and heap-allocated `String` keys; the string variant exists because
+//! heap-allocated keys force a pointer dereference to hash, which changes the
+//! cache-miss profile of the prefetch call itself.
 
 use criterion::{BenchmarkId, Criterion, Throughput};
 use hashbrown::{DefaultHashBuilder, HashMap};
 use std::hint::black_box;
 
-// 16-byte keys, like a common join-key shape (two u64s).
+// ---------- Shared knobs ----------
+//
+// Table-size sweep covers the in-cache → cache-spilled crossover. The prefetch
+// is a hint that pays off only when the control bytes have spilled out of L2/L3
+// and the caller has independent work to overlap with the fetch, so the small
+// sizes are a sanity check (prefetch should be noise or a slight loss) and the
+// large sizes are where the win materializes.
+const SIZES: &[usize] = &[1 << 12, 1 << 16, 1 << 18, 1 << 20, 1 << 22];
+
+// The number of iterations ahead we issue the prefetch. Eight is a common
+// rule-of-thumb (covers ~one cache-miss-worth of work on modern cores) and
+// matches the abseil prefetch_hash idiom.
+const LOOKAHEAD: usize = 8;
+
+// Query batch size. Large enough that fixed per-iteration overhead is
+// amortized; small enough that the bench finishes in seconds.
+const N_QUERIES: usize = 1 << 16;
+
+// 16-byte key, like a common join-key shape (two u64s).
 type Key = (u64, u64);
 
-const SIZES: &[usize] = &[1 << 12, 1 << 16, 1 << 18, 1 << 20, 1 << 22];
-const LOOKAHEAD: usize = 8;
-const N_QUERIES: usize = 1 << 16;
+// ---------- Integer-key workload ----------
+//
+// Keys are packed inline (16 bytes), so hashing the key never dereferences
+// outside the slice. This isolates the prefetch effect to the table's control
+// + data lines: there's no extra cache miss "behind" the key itself.
 
 fn build_map(n: usize) -> HashMap<Key, u64, DefaultHashBuilder> {
     let mut m = HashMap::with_capacity_and_hasher(n, DefaultHashBuilder::default());
@@ -27,7 +45,9 @@ fn build_map(n: usize) -> HashMap<Key, u64, DefaultHashBuilder> {
     m
 }
 
-// A cheap PRNG so the lookup order is unpredictable to the prefetcher.
+// A cheap PRNG so the lookup order is unpredictable to the hardware prefetcher.
+// `xorshift` is deterministic given the seed; the same query set is generated
+// each invocation so the comparison is apples-to-apples.
 fn xorshift(state: &mut u64) -> u64 {
     let mut x = *state;
     x ^= x << 13;
@@ -47,6 +67,22 @@ fn query_keys(n: usize) -> Vec<Key> {
         .collect()
 }
 
+// Miss-heavy query set: every key is drawn from outside the inserted range so
+// every lookup misses. The probe finds an empty control group and never reads
+// the data line. This is the regime where `prefetch_get`'s ctrl-only hint is
+// supposed to win over `prefetch`-both, because the data prefetch is wasted
+// bandwidth when the lookup terminates on the control bytes.
+fn query_keys_miss(n: usize) -> Vec<Key> {
+    let mut state = 0xCAFE_BABE_DEAD_BEEFu64;
+    let offset = (n as u64).saturating_mul(2);
+    (0..N_QUERIES)
+        .map(|_| {
+            let i = offset + (xorshift(&mut state) % n as u64);
+            (i, i.wrapping_mul(0x9E37_79B9_7F4A_7C15))
+        })
+        .collect()
+}
+
 fn lookup_naive(map: &HashMap<Key, u64, DefaultHashBuilder>, keys: &[Key]) -> u64 {
     let mut sum = 0u64;
     for k in keys {
@@ -57,11 +93,14 @@ fn lookup_naive(map: &HashMap<Key, u64, DefaultHashBuilder>, keys: &[Key]) -> u6
     sum
 }
 
+// The look-ahead pattern: prefetch the key `i + LOOKAHEAD` iterations ahead
+// while processing key `i`. By the time iteration `i + LOOKAHEAD` arrives, the
+// control line is already in cache.
 fn lookup_prefetched(map: &HashMap<Key, u64, DefaultHashBuilder>, keys: &[Key]) -> u64 {
     let mut sum = 0u64;
     for (i, k) in keys.iter().enumerate() {
         if let Some(next) = keys.get(i + LOOKAHEAD) {
-            map.prefetch(next);
+            map.prefetch_get(next);
         }
         if let Some(&v) = map.get(k) {
             sum = sum.wrapping_add(v);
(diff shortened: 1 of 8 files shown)


src/prefetch.rs
```diff
@@ -0,0 +1,54 @@
...
+//! (`_mm_prefetch` on x86/x86-64) and fall back to a no-op everywhere else.
+
+/// Issues an L1 read prefetch for the cache line containing `ptr`.
+///
+/// This is a hint only. `ptr` does not need to be valid, aligned, or even
+/// non-null; an out-of-bounds or dangling pointer is fine and will not fault.
+/// On targets without a stable prefetch intrinsic this is a no-op.
+#[inline]
+#[allow(clippy::let_unit_value)]
+pub(crate) fn prefetch_read_l1(ptr: *const u8) {
```
clarfonthey, line 33: I would love to see this feature-gated on using the intrinsics on nightly to see how the performance compares.

benches/prefetch.rs
```diff
@@ -0,0 +1,87 @@
+//! Batch-lookup benchmark: look up a list of keys in a large `HashMap`, with
+//! and without software-prefetching a key a few iterations ahead.
+//!
+//! Prefetching only pays off when the table is large enough that its control
```
clarfonthey: Since the usage behaviour is kind of complicated, I would prefer that this explanation of how the benchmark works be commented throughout the module just so it's a little clearer how the explanation matches the code provided. Not strictly required, just a suggestion.

clarfonthey:
So, first, thank you for working on this! Figuring out the benchmarking and examples is most of the work for a feature like this, and I appreciate you doing it.

In terms of main remarks on the implementation, I think that probably the first major one is I think these methods should all be labelled `prefetch_get` (or `prefetch_read`) instead of just `prefetch`. Even if the actual instructions don't care whether you're reading or writing, I'd say it matters for the kind of prefetching in this crate whether you plan to read or write, since if you end up finding an empty bucket, you have no reason to actually prefetch the place the value would be stored *unless* you were planning to insert there later. (This is also why I would use `prefetch_get` instead of `prefetch_read`, since it's specifically indicating the operation you'd be performing. Similarly, maybe `prefetch_insert` could be used instead of `prefetch_write`.)

Additionally, while I don't think the API needs to be resolved right now, but *especially* if the implementation plans to only prefetch a set number of elements in advance, we should have some way of returning the prefetched hash/bucket index such that it can be reused later. With a set number of elements, this could be stored in a fixed-size array and used later. (Note: you'd need to special case the first n elements in this case.)

Not actually sure whether that'd be a *helpful* API but it feels important to point out since the hashed data *is* stored potentially somewhere else in memory and you don't want to re-pull that page back into cache to hash it again if you don't need to. (Speaking of which, I think this is a rare case where benchmarking with both integers and strings might be useful. Particularly heap-allocated strings, which will be scattered way more in memory.)

src/raw.rs
```diff
@@ -2454,6 +2472,49 @@ impl RawTableInner {
...
+        // valid control index (`pos <= bucket_mask < num_ctrl_bytes`), so the
+        // pointer is in-bounds even before accounting for the hint-only nature
+        // of prefetch.
+        let ctrl_ptr = self.ctrl.as_ptr().wrapping_add(pos);
+
+        // Data bucket at index `pos`: `data_end - (pos + 1) * size`. `data_end`
+        // is `self.ctrl`, so this is `self.ctrl - (pos + 1) * size`. Use
+        // `wrapping_*` so this can never be UB even for the empty singleton
+        // (where it points just before the shared empty control array).
+        let data_ptr = self
```
clarfonthey, line 2530: As mentioned in my comment, but it feels relevant that you might not want to do this if you're optimising for get instead of insert, since the data would be irrelevant if the bucket is empty otherwise. Could perceive a world where this is important when prefetching sparse tables.

That said, you would almost certainly need to mark this as a cold path since otherwise I assume the branch predictor would just pull in the cache anyway.

RyanJamesStewart:
Want to make sure I'm reading the cold-branch suggestion right before coding it. Two implementations of `prefetch_get`:

(a) Issue only the ctrl prefetch, skip the data prefetch entirely. The wasted-data-prefetch-on-empty-buckets concern is solved structurally: the data line is never hinted from `prefetch_get`. Simple, no ctrl read in the prefetch call.

(b) Read ctrl during the prefetch call to probe for a matching tag, prefetch data only on a hit, `#[cold]` on the empty-group branch. Matches your literal description but reads ctrl synchronously, which collapses the prefetch-vs-lookup overlap window the caller is trying to exploit.

My read is that (a) gets the semantics you want without the synchronous ctrl read, but the `#[cold]` annotation in your comment reads like you had (b) in mind. Which were you thinking?

clarfonthey:
I did have b in mind, but I also am way less certain about what would work best, so, I'll trust in whatever actually does work best.

RyanJamesStewart:
v2 pushed (f05a9f9). Addresses all three asks:

1. API split: `prefetch` → `prefetch_get` + new `prefetch_insert`. Three layers (RawTable, HashTable, HashMap, HashSet), `test_prefetch` updated to exercise both. The split is named-only at this point: both methods share the same `RawTableInner::prefetch_both` implementation. Reason below.

2. Nightly intrinsics feature gate in `src/prefetch.rs`. Under `feature = "nightly"`, routes through `core::intrinsics::prefetch_read_data` with locality 3 (matches the stable shim's `_MM_HINT_T0`). Source comment documents the locality invariant so future readers can verify the comparison stays apples-to-apples.

3. Bench doc comments distributed through `benches/prefetch.rs`. New groups: `batch_lookup` (integer, hit-heavy), `batch_lookup_string` (heap-string, hit-heavy, per the non-blocking suggestion), `batch_lookup_miss` (integer, 0% hit rate), `batch_insert`.

On the (a)-vs-(b) decision for `prefetch_get`'s behavior. Benched (a) (ctrl-only) on Crucible (Ryzen 9 9950X):

```
batch_lookup (100% hit rate, integer keys):
  4K:   naive 88µs   prefetch_get 120µs  1.36x slowdown
  64K:  naive 135µs  prefetch_get 190µs  1.40x slowdown
  256K: naive 180µs  prefetch_get 215µs  1.20x slowdown
  1M:   naive 231µs  prefetch_get 282µs  1.22x slowdown
  4M:   naive 278µs  prefetch_get 327µs  1.18x slowdown

batch_lookup_miss (0% hit rate, integer keys):
  4K:   naive 159µs  prefetch_get 126µs  1.26x speedup (noisy, small table)
  64K:  naive 102µs  prefetch_get 207µs  2.03x slowdown
  256K: naive 108µs  prefetch_get 136µs  1.26x slowdown
  1M:   naive 156µs  prefetch_get 186µs  1.19x slowdown
  4M:   naive 170µs  prefetch_get 172µs  neutral
```

(a) regresses on hit-heavy and doesn't recover on miss-heavy. The data prefetch was load-bearing for v1's original `prefetch`-both; removing it loses the win without the trade-off paying off on either workload. (b)'s synchronous-ctrl-read structure would block during `prefetch_get`, which contradicts the overlap-window the API is designed to exploit. Didn't implement (b) because the bench evidence on (a) already showed the data prefetch isn't optional for the headline workload.

v2 ships both methods hinting both lines (behavioral parity with v1's original `prefetch`). Re-bench at 1M and 4M with `prefetch_get`-now-both-lines shows the win restored (1.36-1.46x speedup, in line with v1's reported numbers; absolute numbers shifted from the first run due to thermal variance, relative ratios are stable). The named split keeps room for a behavioral specialization in a follow-up if a workload surfaces that supports the trade-off.

Can implement (b) and re-bench if that's useful for the design conversation. My read of the evidence is that the API surface is the substantive part of your ask, the behavioral split is workload-dependent and currently doesn't pay, and the cleanest landing is named-split-only.

clarfonthey:
Thank you again for doing all the work on this! I am sceptical that the `prefetch_read` / `prefetch_write` distinction will be useful in the long run because I genuinely can't think of an instance where the intent would matter here, but it's worth pointing out that the nightly API *technically* would prefer that the `prefetch_both` be separated for that reason. I personally think it's fine to stay on read prefetch for now on nightly.

I'm particularly interested in what the nightly intrinsics actually do to the benchmark, since I would assume that they don't affect the performance significantly, but intrinsics tend to be weird in a number of ways. It would be particularly helpful for considering how those intrinsics should be long-term since this is a pretty compelling use case.

RyanJamesStewart:
Bench results, stable shim vs nightly intrinsic on Crucible (Ryzen 9 9950X, taskset to core 0). Both runs use `cargo +nightly bench`; only the `--features nightly` flag differs, so the compiler is identical and the only swap is `_mm_prefetch::<_MM_HINT_T0>` for `core::intrinsics::prefetch_read_data::<_, 3>`. Locality-3-equals-T0 invariant documented at the source.

Cleanest signal is the insert group (stable to nightly delta per size, both for naive and prefetch_insert):

```
              naive delta   prefetch_insert delta
  4K:          +0.3%         +0.8%
  64K:          0.0%         +0.2%
  256K:        -1.8%         -1.2%
  1M:          -0.1%         +0.3%
  4M:          -0.6%         -0.4%
```

All within ±2%, both for naive (which doesn't call the prefetch path at all, so it acts as a control) and prefetch_insert (which does). On x86 the intrinsic lowers to the same `prefetcht0` as the stable shim, so equivalence is what the codegen predicts.

Lookup groups had wider cross-run delta (~5-10%) but the delta hits naive too, indicating thermal drift across runs rather than a shim-vs-intrinsic signal. The insert group ran first, lookups second; the nightly run was warmer in the lookup phase. The within-run prefetch_get/naive ratio cancels the baseline drift:

```
batch_lookup (prefetch_get / naive, within run):
                stable shim    nightly intrinsic
  4K:           1.361 slow     1.353 slow
  64K:          1.157 slow     1.103 slow
  256K:         1.016 ~        0.950 fast
  1M:           0.902 fast     0.851 fast
  4M:           0.844 fast     0.792 fast

batch_lookup_miss (prefetch_get / naive, within run):
                stable shim    nightly intrinsic
  4K:           1.403 slow     1.255 slow
  64K:          1.490 slow     1.341 slow
  256K:         1.537 slow     1.444 slow
  1M:           1.214 slow     1.260 slow
  4M:           1.369 slow     1.391 slow
```

The prefetch perf signature is preserved across both impls. Hit-heavy: prefetch starts winning around 256K, peaks at 1.18-1.26x at 1-4M. Miss-heavy: prefetch is a loss across the sweep (the data-line hint is wasted on probes that terminate on control bytes). Both behaviors hold under both shim and intrinsic.

Read: no meaningful codegen difference on x86. If the intrinsic does anything weird on aarch64 or another arch, this bench wouldn't catch it because Crucible is x86-only. That's the open question for the long-term API decision, but not one this PR can answer.

Methodology: `cargo +nightly bench --bench bench -- --warm-up-time 1 --measurement-time 3 "batch_(lookup|insert)"`, with `--features nightly` added for the intrinsic run. Both pinned to core 0 via taskset. Full bench logs available on request.

PR #729, opened 2026-06-06
clippy: fix assert lints

xtqqczze:
https://rust-lang.github.io/rust-clippy/master/index.html#manual_assert
https://rust-lang.github.io/rust-clippy/master/index.html#manual_assert_eq

### DIFF
--- a/tests/hasher_unwind.rs
+++ b/tests/hasher_unwind.rs
@@ -49,9 +49,10 @@ impl BuildHasher for PanicBuildHasher {
     type Hasher = ZeroHasher;
 
     fn build_hasher(&self) -> Self::Hasher {
-        if PANIC_COUNTER.fetch_sub(1, Ordering::SeqCst) == 0 {
-            panic!("panic in BuildHasher::build_hasher");
-        }
+        assert!(
+            PANIC_COUNTER.fetch_sub(1, Ordering::SeqCst) != 0,
+            "panic in BuildHasher::build_hasher"
+        );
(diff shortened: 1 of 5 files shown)


tests/hasher_unwind.rs
```diff
@@ -49,9 +49,10 @@ impl BuildHasher for PanicBuildHasher {
     type Hasher = ZeroHasher;
 
     fn build_hasher(&self) -> Self::Hasher {
-        if PANIC_COUNTER.fetch_sub(1, Ordering::SeqCst) == 0 {
-            panic!("panic in BuildHasher::build_hasher");
-        }
+        assert!(
```
clarfonthey, line 52: …why isn't the lint requiring `assert_ne!` here?

xtqqczze: It doesn't lint in the `assert!(n != 0)` case, so perhaps it is by design.

xtqqczze: `manual_assert_eq` specifically doesn't lint for `!=`, and at least one of the sides is a constant value:

https://github.com/rust-lang/rust-clippy/blob/05a1ffe03584fdf526d90df5c5df3f1e0ef23a63/tests/ui/manual_assert_eq.fixed#L34

clarfonthey: I see. I guess that I had noticed `assert_ne` used elsewhere in this PR so had gotten a bit confused.

PR #733, opened 2026-06-20
Fake `ToOwned` trait

clarfonthey:
This follows one of the suggestions for removing the `alloc` dependency, specifically to allow moving `HashMap` into `alloc`: essentially, if we don't have `ToOwned`, store-bought is fine.

The expectation here is that specifically for libstd, they can implement a blanket forwarding of the actual `ToOwned` to this trait and it should work as expected. This is fine by coherence rules even though the coherence rules kind of don't make any more sense once you *are* libstd.

Not 100% sure if this should be merged immediately, since I want to make sure that we actually can get a full alloc-less version of hashbrown working first. But the implementation seems relatively straightforward.

### DIFF
--- a/src/alloc.rs
+++ b/src/alloc.rs
@@ -2,6 +2,22 @@
 pub(crate) use self::inner::AllocError;
 pub(crate) use self::inner::{Allocator, Global, do_alloc};
 
+// Since `ToOwned` is defined in `alloc`, this would prevent us from using
+// this crate in `alloc`; that would be a circular dependency. To get around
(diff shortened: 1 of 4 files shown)


RustyYato:
For posterity: see this [Zulip](https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/Implementation.20of.20.60HashTable.60/with/605284503) thread

Amanieu:
This seems mostly fine to me, but I would like to see the full story for not depending on alloc to be resolved before merging it.

PR #736, opened 2026-07-03
Use wrapping arithmetic in ProbeSeq::move_next to avoid overflow

### LINKED ISSUE #735: ProbeSeq::move_next() potential arithmetic overflow
Hi,

I found a potential arithmetic overflow issue through Kani formal verification.

Location: src/raw.rs:89-90

Current Code:

impl ProbeSeq {
fn move_next(&mut self, bucket_mask: usize) {
self.stride += Group::WIDTH; // Can overflow
self.pos += self.stride; // Can overflow
self.pos &= bucket_mask;
}
}
Problem:

In extreme cases with very long probe sequences, stride and pos can overflow before being masked.

Suggested Mitigation:

impl ProbeSeq {
fn move_next(&mut self, bucket_mask: usize) {
self.stride = self.stride.wrapping_add(Group::WIDTH);
self.pos = self.pos.wrapping_add(self.stride) & bucket_mask;
}
}
Could you please confirm if this is a valid concern?

amitmishra11:
## Bug

In `ProbeSeq::move_next` (src/raw.rs), `stride` and `pos` are updated with plain `+=`:

```rust
self.stride += Group::WIDTH;
self.pos += self.stride;
self.pos &= bucket_mask;
```

In a debug build this panics; in a release build it produces undefined behavior (integer overflow). Although the existing `debug_assert!` verifies that `stride <= bucket_mask` before the addition, it fires *before* the increment, so the assert itself does not rule out overflow once `Group::WIDTH` is added.

This was identified via Kani formal verification and confirmed as a bug by @Amanieu in #735.

## Fix

Replace the plain additions with `wrapping_add`, which is what the subsequent `& bucket_mask` mask assumes:

```rust
self.stride = self.stride.wrapping_add(Group::WIDTH);
self.pos = self.pos.wrapping_add(self.stride) & bucket_mask;
```

The logic is otherwise identical; `wrapping_add` only differs from `+` when overflow would occur, and the final mask ensures `pos` stays in range regardless.

## Verification

`cargo check` passes cleanly. `cargo test` cannot run on this machine due to an absent MSVC linker, but the type-level correctness of the change is confirmed by `cargo check`.

Closes #735

### DIFF
--- a/src/raw.rs
+++ b/src/raw.rs
@@ -87,9 +87,8 @@ impl ProbeSeq {
             "Went past end of probe sequence"
         );
 
-        self.stride += Group::WIDTH;
-        self.pos += self.stride;
-        self.pos &= bucket_mask;
+        self.stride = self.stride.wrapping_add(Group::WIDTH);
+        self.pos = self.pos.wrapping_add(self.stride) & bucket_mask;
     }
 }
 


clarfonthey:
It appears that the test failures are due to some unrelated lint changes; I can work on those in a separate PR, or someone else can get to that first. Otherwise, this PR looks fine and can be merged after that is fixed.

amitmishra11:
Thanks for the review and for clarifying the CI failures. Happy to wait for those lint changes to land first, or rebase once they're merged if that helps.

clarfonthey:
#738 is the fix; shouldn't need any changes from you once that merges.

clarfonthey:
Trying to remember how to push changes to someone else's branch.

PR #740, opened 2026-07-17
Replace `cfg_if!` macro with `cfg_select!` fallback

clarfonthey:
The main benefit this gives is rustfmt support; `cfg_select!` is properly formatted, unlike `cfg_if!`. Additionally, it makes migrating when we eventually bump MSRV easier.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1,70 +1,42 @@
-// See the cfg-if crate.
-#[expect(unused_macro_rules)]
-macro_rules! cfg_if {
-    // match if/else chains with a final `else`
-    ($(
-        if #[cfg($($meta:meta),*)] { $($it:item)* }
-    ) else * else {
-        $($it2:item)*
-    }) => {
-        cfg_if! {
-            @__items
-            () ;
-            $( ( ($($meta),*) ($($it)*) ), )*
-            ( () ($($it2)*) ),
-        }
+// FIXME: remove once we bump MSRV >= 1.95.0
+// Originally by @folkertdev on Zulip:
+// https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/compiler-builtins.20msrv.20policy.3F/near/606320965
+/// Fallback `cfg_select!` on older Rust.
+///
+/// Requires braces around output, but works for items and expressions.
+#[cfg(not(feature = "nightly"))]
+#[macro_export]
+macro_rules! cfg_select {
+    ({ $($tt:tt)* }) => {{
+        $crate::cfg_select! { $($tt)* }
+    }};
+    (_ => { $($output:tt)* }) => {
+        $($output)*
     };
-
-    // match if/else chains lacking a final `else`
     (
-        if #[cfg($($i_met:meta),*)] { $($i_it:item)* }
-        $(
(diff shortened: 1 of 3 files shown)


src/macros.rs
```diff
@@ -1,32 +1,15 @@
...
-    }) => {
-        cfg_if! {
-            @__items
-            () ;
-            $( ( ($($meta),*) ($($it)*) ), )*
-            ( () ($($it2)*) ),
-        }
-    };
-
-    // match if/else chains lacking a final `else`
```
clarfonthey, line 18: We actually never used this branch, but I guess we wanted it for completeness.

xtqqczze:
Perhaps this implementation might work better with `_` fallback syntax:

https://github.com/rust-lang/compiler-builtins/blob/9c7d3e8d064e331144dbe3426a0bf5311399a857/libm/src/math/support/macros.rs#L5

clarfonthey:
That sure is a simpler implementation. I'll try that one out.

src/macros.rs
xtqqczze, line 1: If you're happy with this implementation, could you please add a link to https://github.com/rust-lang/compiler-builtins/blob/9c7d3e8d064e331144dbe3426a0bf5311399a857/libm/src/math/support/macros.rs in the comments. And then update the PR description.

xtqqczze:
Perhaps someone could publish a `cfg_select` crate, since this has come up elsewhere: https://github.com/rusqlite/rusqlite/pull/1850#issuecomment-4592539293

clarfonthey:
We explicitly avoid depending on `cfg-if` to avoid increasing dependencies, so, unlikely we'd add a new dependency here. But if someone wants to maintain that, it might be helpful.

src/macros.rs
clarfonthey: I'll link the original: [#t-libs > compiler-builtins msrv policy? @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/compiler-builtins.20msrv.20policy.3F/near/606320965)

xtqqczze:
One limitation of the current implementation is that branch bodies must be enclosed in braces. For example, the following is not supported:

```rust
cfg_select! {
    unix => foo()
    windows => bar()
}
```

xtqqczze:
I’d still include a link to the `compiler-builtins` implementation in a comment, as it may help keep the two implementations in sync over time:

https://github.com/rust-lang/compiler-builtins/blob/9c7d3e8d064e331144dbe3426a0bf5311399a857/libm/src/math/support/macros.rs

src/macros.rs
xtqqczze: > I'll link the original: [#t-libs > compiler-builtins msrv policy? @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/219381-t-libs/topic/compiler-builtins.20msrv.20policy.3F/near/606320965)

Since @folkertdev is already credited for this code in the `compiler-builtins` repository, I don’t think we need to link to the original Zulip discussion as well.

clarfonthey:
That seems pretty unnecessary IMHO. This satisfies our needs until we can update the MSRV to remove it.

clarfonthey:
poke @cuviper @Amanieu for review (no rush)

cuviper (approved):
LGTM!

PR #741, opened 2026-07-27
fix: use core instead of std

xtqqczze:


benches/general_ops.rs
```diff
@@ -2,14 +2,14 @@
...
-use std::{
-    collections::hash_map::RandomState,
+use core::{
     hint::black_box,
     sync::atomic::{self, AtomicUsize},
 };
+use criterion::Criterion;
+use hashbrown::DefaultHashBuilder;
+use hashbrown::{HashMap, HashSet};
+use std::collections::hash_map::RandomState;
```
clarfonthey: Actually, with our MSRV bump, this can be `std::hash::RandomState`

src/map.rs
```diff
@@ -2669,13 +2669,13 @@ pub struct ValuesMut<'a, K, V> {
 ///
 /// // Existing key (or_insert)
 /// let v = map.entry("b").or_insert(2);
-/// assert_eq!(std::mem::replace(v, 2), 20);
```
clarfonthey, line 2672: Mostly in favour of this change, but I would rather not update the doctests specifically since in general we prefer making those as simple as possible, and using only `std` allows better import grouping.

src/raw.rs
```diff
@@ -4507,7 +4507,7 @@ mod test_map {
         }
 
         unsafe impl Allocator for MyAlloc {
-            fn allocate(&self, layout: Layout) -> std::result::Result<NonNull<[u8]>, AllocError> {
+            fn allocate(&self, layout: Layout) -> core::result::Result<NonNull<[u8]>, AllocError> {
```
clarfonthey: This can just be `Result` from the prelude.

tests/hasher_unwind.rs
```diff
@@ -9,14 +9,15 @@
...
     hash::{BuildHasher, Hash, Hasher},
-    panic::{AssertUnwindSafe, catch_unwind},
-    sync::Mutex,
+    panic::AssertUnwindSafe,
     sync::atomic::{AtomicUsize, Ordering},
 };
+use hashbrown::HashMap;
+use std::collections::BTreeSet;
+use std::panic::catch_unwind;
+use std::sync::Mutex;
```
clarfonthey, line 20: Would brace-group these:

```suggestion
use std::{
    collections::BTreeSet,
    panic::catch_unwind,
    sync::Mutex,
};
```

xtqqczze: I had considered that, but rustfmt wanted to remove whitespace which I didn't like:

```rust
use std::{collections::BTreeSet, panic::catch_unwind, sync::Mutex};
```

src/raw.rs
xtqqczze: Hmm, would be nice if there was a lint for shadowing the prelude.

tests/hasher_unwind.rs
clarfonthey: I see; personally, I still would go with it, but it's up to you.

src/raw.rs
xtqqczze: I'd rather this PR didn't get too large, so let's make the prelude changes separately: https://github.com/rust-lang/hashbrown/pull/743

tests/hasher_unwind.rs
xtqqczze: In an ideal world, we'd use [`imports_layout = "Vertical"`](https://rust-lang.github.io/rustfmt/?version=v1.10.0&search=import#imports_layout) like [`libc`](https://github.com/rust-lang/libc/blob/63b34a986c570232352e2aee1ea35afe31229998/.rustfmt.toml#L6) but it's an unstable feature.

clarfonthey:
FWIW: there are clippy lints for `alloc_instead_of_core`, `std_instead_of_core`, and `std_instead_of_alloc`. Don't think they can be set up to exclude doctests though.

src/map.rs
xtqqczze: Import grouping shouldn’t be an issue here. In the doctests, we only use the following `std` imports, and they are never used in combination with `core`:

```rust
/// use std::hash::RandomState;
/// use std::rc::Rc;
```

benches/general_ops.rs
xtqqczze: Split to https://github.com/rust-lang/hashbrown/pull/747.

tests/hasher_unwind.rs
xtqqczze: @clarfonthey By the way, would you prefer a fixup commit or rebase?

clarfonthey: For the main repo we don't want nested merge commits so a rebase would be ideal. Don't really mind if the result is multiple commits, although figured you'd also squash them as well anyway.

clarfonthey: If you want to propose new `rustfmt.toml` changes, wouldn't actually mind that, since formatting in CI is always via nightly. But probably better as a separate PR which can be ignored on blame.

clarfonthey:
At the moment, LGTM minus the clippy lints to help avoid this in the future.

tests/hasher_unwind.rs
xtqqczze: Right, yeah. Since the repo doesn’t use squash merges, I think a rebase is needed despite any reviewer preference.

PR #745, opened 2026-08-26
Add debug assertion documenting the bucket_mask overflow invariant

### LINKED ISSUE #734: bucket_mask_to_capacity() integer overflow when bucket_mask == usize::MAX
Hi,

I found a potential integer overflow issue through Kani formal verification.

Location: src/raw.rs:189

Current Code:

fn bucket_mask_to_capacity(bucket_mask: usize) -> usize {
if bucket_mask < 8 {
bucket_mask
} else {
((bucket_mask + 1) / 8) * 7 // Overflow when bucket_mask == usize::MAX!
}
}
Problem:

When bucket_mask = usize::MAX:

bucket_mask + 1 wraps to 0

Returns 0 instead of correct capacity

Suggested Fix:

fn bucket_mask_to_capacity(bucket_mask: usize) -> usize {
if bucket_mask < 8 {
bucket_mask
} else {
bucket_mask.checked_add(1)
.map(|bm| (bm / 8) * 7)
.unwrap_or(usize::MAX)
}
}
Could you please confirm if this is a valid concern?

ChrisJr404:
Follows up #734. You said a debug assertion would be a good way to document that bucket_mask can never be usize::MAX, so this adds one where the +1 happens.

Fixes #734.

### DIFF
--- a/src/raw.rs
+++ b/src/raw.rs
@@ -185,6 +185,9 @@ fn bucket_mask_to_capacity(bucket_mask: usize) -> usize {
         // Keep in mind that the bucket mask is one less than the bucket count.
         bucket_mask
     } else {
+        // `bucket_mask` is bounded by the maximum allocation size, so it can
+        // never be `usize::MAX` and the `+ 1` below cannot overflow.
+        debug_assert!(bucket_mask != usize::MAX);
         // For larger tables we reserve 12.5% of the slots as empty.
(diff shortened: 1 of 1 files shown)


clarfonthey:
Longer-term I think it would be good to use a niche type for the mask to more explicitly encode the valid values for this type, but for now, this should be fine.

PR #747, opened 2026-09-06
refactor: use `std::hash::RandomState`

xtqqczze:
Since Rust 1.76 we can use `std::hash::RandomState` instead of `std::collections::hash_map::RandomState`.


clarfonthey:
Right, that actually is well within our MSRV. Thank you!

EDIT: Right, had actually recommended this to you explicitly. Thank you regardless.

PR #748, opened 2026-09-06
style: refactor import layout

xtqqczze:
Configure `rustfmt` to use vertical import formatting and apply the new layout throughout the crate.

- Update` rustfmt` configuration and reformat imports throughout.
- Group related imports into multi-line use statements.

Use the following import order:

1. `core::`
1. `std::`
1. _external crates_
1. `crate::`
1. `super::`
1. `self::`

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,1 @@
+imports_layout = "Vertical"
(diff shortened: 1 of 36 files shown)


xtqqczze:
@clarfonthey Doesn't look like `cargo fmt --all -- --check` is using nightly in `ci/tools.sh`

clarfonthey:
Ah, I thought it was. Either way, that should be an easy change.

xtqqczze:
@clarfonthey Right, I'm not completely sure how `ci/tools.sh` works, would you mind taking a look? And what do you think about the changes in principle?

